PR #960, opened 2024-05-01
feat(testing): OP stack rollup

calbera:
Provides setup and guide to run a OP stack L2 on top of a beacon-kit L1


Ferret-san:
wen L3s?

PR #1278, opened 2024-06-02
feat: init nix

benluelo:
init nix with a simple flake and build `beacond`, as well as a very basic dev shell. checks for beacond will be a bit more work, since the tests are a bit impure (see the comment in `beacond/beacond.nix`).

to build:

```sh
nix build .#beacond
# or, since beacond is reexported as the default flake output:
nix build .#
```

`beacond` is now available in `result/bin/beacond`.

alternatively, you can run it directly:

```sh
nix run .#beacond
# or
nix run .#
```

to enter the devshell:

```sh
nix develop
```

all of the above commands can be run from anywhere in the repo (nix will walk up the filesystem until it finds a `flake.nix`)

you may need to escape the `#` (`.\#`) depending on your shell.

### DIFF
--- a/build/scripts/golines.sh
+++ b/build/scripts/golines.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # SPDX-License-Identifier: MIT
 #
 # Copyright (c) 2024 Berachain Foundation
--- a/flake.nix
+++ b/flake.nix
@@ -30,6 +30,7 @@
             go_1_22
             nil
             nixfmt-classic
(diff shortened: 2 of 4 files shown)


itsdevbear:
@corduroybera u wanna figure out what we should do here?

corduroybera:
> @corduroybera u wanna figure out what we should do here?

@itsdevbear yeah, I want to look into methods for more automated vendorHash updates so it stays up-to-date before we merge. We update dependencies so frequently in BeaconKit I think it'd be good.

PR #1336, opened 2024-06-05
fix(runtime): cleanup and fix lagging dep injects

ocnc2:
**Summary:
2 aims:**
1. make`ProvideRuntime` great again: No longer builds every db, service, registry, and almost every other miscellaneous nodebuilder component
2. show that it is possible for the root of all our problems (_depinject_) to be resolved

POC.
Currently, provide runtime is very bloated with building all of its constituents. This PR attempts to resolve this by allowing depinject to handle the construction of all the dependencies. This will remove most of the unnecessary hard dependencies and responsibilities of the runtime. Additionally as a side effect, previous nil pointer issues with dependencies not properly populated with values should also be resolved if done in this way.

still need to:
- let depinject handle final building of BKRuntime
- move the initial source of all injection of hard-types (for generics) into the actual `depinject.inject` call
- a lot of cleanup+refactoring


ocnc2:
addressed in other PRs

PR #1376, opened 2024-06-07
chore(depinject): provide chainspec

archbear:


ocnc2:
4th times the charm https://github.com/berachain/beacon-kit/pull/1452

PR #1408, opened 2024-06-10
refactor: add miss defer and refactor name function

hoank101:


itsdevbear (approved):
thanks for the contribution

PR #1412, opened 2024-06-10
refactor: fix lint issues by reducing nestif

trungnt1811:


itsdevbear:
Personally I think it's more complicated to understand this way, I think it's possible to reuse handleOptimisticPayloadBuild as a helper

trungnt1811:
Hi @itsdevbear, could you take a look at my new commit? Thanks

Issue #1407, opened 2024-06-10
beacond suggested-fee-recipient address flag not working


NodeKing:
When using the `beacond` binary with the  `--beacon-kit.builder.suggested-fee-recipient` flag it doesn't work. Not a huge issue as setting it in the app.toml file instead works fine, just a FYI.

ocnc2:
addressed by https://github.com/berachain/beacon-kit/pull/1415

Suggested fee recipient now set using the following flag:
`--beacon-kit.payload-builder.suggested-fee-recipient`

PR #1415, opened 2024-06-10
fix(cli): payload builder attributes

ocnc2:
https://github.com/berachain/beacon-kit/issues/1407

### DIFF
--- a/mod/node-core/pkg/config/flags/flags.go
+++ b/mod/node-core/pkg/config/flags/flags.go
@@ -28,9 +28,12 @@ const (
 	// Builder Config.
 	builderRoot              = beaconKitRoot + "payload-builder."
 	SuggestedFeeRecipient    = builderRoot + "suggested-fee-recipient"
-	Graffiti                 = builderRoot + "graffiti"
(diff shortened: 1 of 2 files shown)


itsdevbear:
Oh these flags are cooked, some should be under validator

mod/node-core/pkg/config/flags/flags.go
```diff
@@ -26,12 +26,15 @@ const (
...
 
 	// Builder Config.
-	builderRoot              = beaconKitRoot + "builder."
+	builderRoot              = beaconKitRoot + "payload-builder."
 	SuggestedFeeRecipient    = builderRoot + "suggested-fee-recipient"
-	Graffiti                 = builderRoot + "graffiti"
 	LocalBuilderEnabled      = builderRoot + "local-builder-enabled"
 	LocalBuildPayloadTimeout = builderRoot + "local-build-payload-timeout"
 
+	// Validator Config.
```
itsdevbear, line 34: @ocnc2 is the template.go okay?

ocnc2: yep template.go already has this config reflected

PR #1436, opened 2024-06-12
refactor: disable gochecknoglobals in test files

TropicalDog17:
## Description

Disable the `gochecknoglobals` linter on tests files, since it is not really useful for tests.


itsdevbear:
Agreed thanks!

itsdevbear (approved):

PR #1444, opened 2024-06-12
chore(deposit): cleanup + remove deadcode

archbear:


mod/primitives/pkg/feed/event.go
```diff
@@ -28,26 +28,26 @@ import (
...
-		ctx:  ctx,
-		name: name,
-		data: data,
+		ctx:       ctx,
+		eventType: eventType,
+		data:      data,
 	}
 }
 
 // Name returns the name of the event.
```
nidhi-singh02, line 48: nitpick : `Type` in comment instead of Name

nidhi-singh02 (approved):

PR #1438, opened 2024-06-12
chore(node-core): decouple runtime and app module

archbear:
## Summary by CodeRabbit

- **New Features**
  - Introduced an `ABCIMiddleware` struct that enhances interaction between ABCI and Beacon logic.
  
- **Enhancements**
  - Updated middleware functions and types to improve flexibility and functionality in managing Beacon blocks and states.
  - Added new methods to the `BeaconState` interface for better access to beacon state information.
  - Updated various middleware components to use more specific type parameters and structures.

- **Refactor**
  - Restructured import paths and type declarations to streamline middleware and runtime components.
  
- **Chores**
  - Removed outdated imports and adjusted dependencies in several components for cleaner integration and better maintainability.

PR #1453, opened 2024-06-12
feat(node-core): decouple middleware from runtime

ocnc2:
necessary for https://github.com/berachain/beacon-kit/pull/1438, so we can directly provide the middleware handlers to the ABCI functions in AppCreator


mod/node-core/pkg/components/middleware.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/berachain/beacon-kit/mod/primitives"
+	"github.com/berachain/beacon-kit/mod/runtime/pkg/runtime/middleware"
+	"github.com/berachain/beacon-kit/mod/runtime/pkg/service"
+	depositdb "github.com/berachain/beacon-kit/mod/storage/pkg/deposit"
+)
+
+// ValidatorMiddlewareInput is the input for the validator middleware provider.
+type ValidatorMiddlewareInput struct {
+	depinject.In
+	// could instead just take in the services
```
archbear, line 40: take in the services themselves, no reason to give the provider knowledge of what exists in the registry

itsdevbear:
lgtm, I kind of agree with @archbear, the other option would be the only dep everyone takes is the registry itself, but that'll get quite complex most likely

PR #1471, opened 2024-06-13
chore(node-core): types.go

archbear:
ease consistency of typing by defining type aliases for generics

## Summary by CodeRabbit

- **Refactor**
  - Updated various types to more generic forms across multiple components to improve maintainability and streamline code.
  - Simplified function signatures and struct declarations, removing unnecessary imports and specific type details.
  - Replaced specific nested types with direct type assignments in service components.
  - Removed redundant type alias declarations and import statements for a cleaner codebase.


mod/node-core/pkg/components/types.go
```diff
@@ -0,0 +1,193 @@
...
+		AvailabilityStore,
+		BeaconBlockBody,
+		BeaconState,
+		BlobSidecars,
+		Deposit,
+		DepositStore,
+	]
+
+	// ValidatorMiddleware is a type alias for the validator middleware.
+	ValidatorMiddleware = *middleware.ValidatorMiddleware[
```
itsdevbear, line 172: this is bad practice, don't alias the pointer, just the type directly

itsdevbear (approved):
fix the pointer thing, you shouldn't hide the fact it's a pointer behind the alias

PR #1472, opened 2024-06-14
fix(deposit): synchronized failedBlocks between multiple goroutines

htiennv:
## Summary by CodeRabbit

- **Refactor**
  - Improved the retry mechanism for failed block processing in the deposit service by switching from a map to a channel, enhancing efficiency and reliability.

### DIFF
--- a/mod/execution/pkg/deposit/service.go
+++ b/mod/execution/pkg/deposit/service.go
@@ -27,7 +27,7 @@ import (
 	"github.com/berachain/beacon-kit/mod/primitives/pkg/math"
 )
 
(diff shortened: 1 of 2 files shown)


itsdevbear:
hi @htiennv I agree with this change, but this piece of the code base needs a broader re-architecture overall

your logic here (much like our not so great existing solution) will result in infinite number of go-routines being spun up

htiennv:
> hi @htiennv I agree with this change, but this piece of the code base needs a broader re-architecture overall
> 
> your logic here (much like our not so great existing solution) will result in infinite number of go-routines being spun up

thanks for your response, i'll close this PR

PR #1475, opened 2024-06-14
Use 'require' instead of 'if' and 'revert'

haitranwang:
**Use require instead of if and revert:** require is a built-in function in Solidity that throws an exception if the condition is not met and reverts the changes. It's more readable and concise.

## Summary by CodeRabbit

- **Refactor**
  - Improved input validation in the staking contract by replacing `revert` statements with `require` statements, enhancing control flow and error messaging for users.


rustybera:
No improvement on the bytecode level.

PR #1492, opened 2024-06-15
fix(ssz): adjust the initial offset of variables-length fields

### LINKED ISSUE #1131: fix(ssz/serialization): Offsets dont match
When serializing full objects, e.g. Deneb or Bellatrix sample beacon state, our serialized byte buffer does not match fastssz output completely. 
It appears to be different at the same positions that offsets for variable elements are written, e.g. the offset written for historicalRoots with ssz-size:"?,32" differs in our output vs fastssz. 
Currently, all fixed length till the offset are encoded correctly. 

1. Assess, via our own deserializer if our implementation is correct and can be deserialized.
2. If it can't be deserialized, debug and fix the issue. 
3. If it can be deserialized, check against the spec and leave a note for the discrepancy between fastssz and our impl

Repro steps:
1. Look at `spec_test.go` in `ssz/v2/...` 
2. Bellatrix and Deneb parity tests, compare the length of the results match, change to comparing the 2 objects
3. Fix test 

parent: #1129 #808

TropicalDog17:
### Description

Fix offset calculation and update the parity test.

Closes #1131
Related to #1064  

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected the order of operations in the offset calculation to ensure accurate results.
  
- **Tests**
  - Improved test assertions to ensure equality and length checks for values, enhancing test reliability.


TropicalDog17:
Hi @itsdevbear could you take a look at this PR, would be nice to get this bug fix merged so I could continue working on improving perf.

itsdevbear:
lgtm. @TropicalDog17 we are likely going a different direction with our v2 ssz library, so i wouldn't spend too much time here rn. It needs some more broader design, so it may end up being thrown away, will merge this though bc I agree it's correct.

TropicalDog17:
> lgtm. @TropicalDog17 we are likely going a different direction with our v2 ssz library, so i wouldn't spend too much time here rn. It needs some more broader design, so it may end up being thrown away, will merge this though bc I agree it's correct.

Thanks for your information!

PR #1511, opened 2024-06-17
feat(async): Async PreBlock part 1 

itsdevbear:
## Summary by CodeRabbit

- **New Features**
  - Introduced the `ABCIMiddleware` for enhanced block processing, proposal handling, and validator updates management.
  - Added `ValidatorUpdates` type with methods to remove duplicates and sort updates.

- **Bug Fixes**
  - Improved management of validator data by eliminating duplicates and sorting entries.

- **Refactor**
  - Renamed `ValidatorMiddleware` to `ABCIMiddleware` across several components for consistency.
  - Removed outdated finalize block middleware functions.

- **Tests**
  - Added tests for `ValidatorUpdates` to ensure correct removal of duplicates and sorting.


mod/primitives/pkg/transition/validator_update.go
```diff
@@ -33,3 +38,26 @@ type ValidatorUpdate struct {
 	// EffectiveBalance is the effective balance of the validator.
 	EffectiveBalance math.Gwei
 }
+
+// RemoveDuplicates removes duplicate validator updates.
+func (vu ValidatorUpdates) RemoveDuplicates() ValidatorUpdates {
```
itsdevbear, line 43: @ocnc can you review, you did original work here

ocnc: I see what you're doing but I think in the original logic, we were taking the latest occurrence of a val update (whatever was appended last to the val updates array is the update we want to retain), and this logic takes the first occurrence instead

ocnc: Because of how the process multiple slots works => slot1, slot2, slot3, and the updates from slot3 should override slot1

itsdevbear: addressed

PR #1519, opened 2024-06-18
add otterscan to kurtosis

nidhi-singh02:
Attaching otterscan snippet

## Summary by CodeRabbit

- **New Features**
  - Introduced Otterscan module to enhance blockchain scanning capabilities.
  - Added functionality to launch an Otterscan service with new port and environment configurations.

- **Enhancements**
  - Updated service configuration to support public port specifications and full node deployment.
  - Improved execution environment by adding the `ots` module to the configuration for Erigon nodes.

- **Configuration Changes**
  - Adjusted `beaconkit-all.yaml` to increase replicas for Erigon and added Otterscan service for better network support.

Screenshot 2024-06-28 at 11 37 54 AM


kurtosis/src/services/otterscan/launcher.star
```diff
@@ -0,0 +1,53 @@
...
+    ),
+}
+
+def launch_otterscan(
+        plan,
+        full_node_el_clients,
+        client_from_user):
+    el_client_info = {}
+
+    # TODO: If client_from_user is other than erigon node, give error or something to user.
```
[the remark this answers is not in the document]
nidhi-singh02: yes

itsdevbear:
@corduroybera can you review?

kurtosis/src/lib/service_config.star
```diff
@@ -131,12 +133,23 @@ def validate_service_config_types(service_config):
...
 
-def create_from_config(config):
+def create_public_port_specs_from_config(config, is_full_node):
+    ports = {}
+    if is_full_node:
+        ports = {}
+        for port_key, port_spec in config["public_ports"].items():
+            ports[port_key] = port_spec_lib.create_port_spec(port_spec)
+
+    return ports
```
corduroybera, line 143: Does otterscan let you customize the ETH JSON-RPC port you're hitting? It seems messy to put in all these exceptions just to expose erigon's RPC as 8545 when we can pass along the generated port like we do with other services

kurtosis/src/nodes/execution/execution.star
```diff
@@ -107,10 +121,11 @@ def add_bootnodes(node_module, config, bootnodes):
 
     return config
 
-def deploy_nodes(plan, configs):
+def deploy_nodes(plan, configs, is_full_node = False):
     service_configs = {}
     for config in configs:
         service_configs[config["name"]] = service_config_lib.create_from_config(config)
+        service_configs[config["name"]] = service_config_lib.create_from_config(config, is_full_node)
```
corduroybera, line 128: This is called twice, even when it's unnecessary

corduroybera (changes requested):

kurtosis/src/lib/service_config.star
corduroybera: I would prefer to leave 8545 available on local machine because you never know when someone is already running a local node on 8545. For example, I commonly port-forward a kubernetes node to check things on another network and will usually use 8545 when doing so

nidhi-singh02: Otterscan uses ERIGON_URL to customize the URL so yes we could change the port. As otterscan communicates with the erigon RPC URL, port 8545 does not work in this case unlike other services, we need the public port that is exposed by kurtosis and since kurtosis does it in random order there is no way to figure out the public port from ServiceConfig. So I am explicitly setting the public port for the erigon full node, I kept it the same for ease of identification. We could change the public port to any other port, may be 8546 or others ?
I agree with the point whether 8545 might be occupied on the machine.

kurtosis/src/nodes/execution/execution.star
nidhi-singh02: Thanks for pointing out, addressed it. I may have messed this while resolving merge conflicts.

kurtosis/src/lib/service_config.star
corduroybera: nidhi tried all the other variants, seems the hostname : port method isn't working with erigon for some reason so marking this as resolved

kurtosis/src/services/otterscan/launcher.star
```diff
@@ -0,0 +1,36 @@
...
+def launch_otterscan(
+        plan,
+        client_from_user):
+    # Currently fetching the public port from service config is not allowed,
+    # .ports["eth-json-rpc"].number returns the private port. Hence hardcoding the public port.
+    if client_from_user.split("-")[2] != "erigon":
+        fail("Currently only erigon client is supported for otterscan")
+    config = get_config()
+    plan.add_service(SERVICE_NAME, config)
+
```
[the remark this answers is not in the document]
nidhi-singh02: Handle the errror for me.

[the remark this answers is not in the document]
nidhi-singh02: what is updated in the above code, it is same as before.

corduroybera (approved):

PR #1520, opened 2024-06-18
fix(deposit): synchronized failedBlocks from channel

trungnt1811:
## Summary by CodeRabbit

- **Refactor**
  - Improved the efficiency of handling failed deposit blocks by using a buffered channel instead of a map.
  - Enhanced error handling and retry logic for better performance and reliability.

- **New Functionality**
  - Introduced a new mechanism to drain and retry failed blocks using the `drainFailedBlocks` function.


ocnc (commented):
I don't like this change as the channel has a cap. Theres no guarantee that a block which was failed to be retrieved will eventually be retrieved and not just pushed out -- Missing deposits is a non-negotiable

trungnt1811:
@ocnc thank you for your feedback. I completely agree and understand your concern. I will close this PR and will try to think of a better solution. Thanks.

PR #1528, opened 2024-06-19
test(net/jwt): refactor test to use `require` pkg

TropicalDog17:
### Description

Use `require` pkg for better test results' readability.

### DIFF
--- a/mod/primitives/pkg/net/jwt/jwt_test.go
+++ b/mod/primitives/pkg/net/jwt/jwt_test.go
@@ -22,13 +22,13 @@
 package jwt_test
 
 import (
-	"reflect"
 	"strings"
 	"testing"
 
 	"github.com/berachain/beacon-kit/mod/primitives/pkg/common"
 	"github.com/berachain/beacon-kit/mod/primitives/pkg/hex"
 	"github.com/berachain/beacon-kit/mod/primitives/pkg/net/jwt"
+	"github.com/stretchr/testify/require"
 )
 
 func TestNewFromHex(t *testing.T) {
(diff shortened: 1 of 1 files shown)


ocnc2 (approved):

mod/primitives/pkg/net/jwt/jwt_test.go
```diff
@@ -105,95 +102,54 @@ func TestSecretString(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if got := tt.secret.String(); got != tt.want {
-				t.Errorf("Secret.String() = %v, want %v", got, tt.want)
-			}
+			require.Equal(t, tt.want, tt.secret.String(), "Secret.String() mismatch")
```
nidhi-singh02, line 105: If both are equal, then why mismatch message ? Kindly check the condition once, it should be NotEqual.

mod/primitives/pkg/net/jwt/jwt_test.go
```diff
@@ -105,95 +102,54 @@ func TestSecretString(t *testing.T) {
...
-	}
-
-	if len(secret.Bytes()) != 32 {
-		t.Errorf(
-			"NewRandom() generated a secret of incorrect length: got %d, want %d",
-			len(secret.Bytes()),
-			32,
-		)
-	}
+	require.NoError(t, err, "NewRandom() error")
```
nidhi-singh02, line 112: If there's no error, why do we need message ?

nidhi-singh02 (changes requested):

nidhi-singh02 (approved):
lgtm!

itsdevbear (approved):

PR #1529, opened 2024-06-19
test(primitives): add unit tests for chain/helpers.go

trungnt1811:
## Summary by CodeRabbit

- **Tests**
  - Added comprehensive test coverage for epoch, slot, and fork version handling methods to ensure robust blockchain functionality.


mod/primitives/pkg/chain/helpers_test.go
archbear, line 1: you can pull out the types from the functions and the chainspec var declaration and then set the spec data at the top of each function to reduce redundancy

trungnt1811: @archbear, thanks so much! I just refactored it. Could you please help me review it again?

mod/primitives/pkg/chain/helpers_test.go
```diff
@@ -0,0 +1,128 @@
...
+	}{
+		{epoch: 9, expected: version.Deneb},
+		{epoch: 10, expected: version.Electra},
+		{epoch: 11, expected: version.Electra},
+	}
+
+	// Run test cases
+	for _, tc := range testCases {
+		result := spec.ActiveForkVersionForEpoch(tc.epoch)
+		assert.Equal(t, tc.expected, result, "unexpected active fork version for epoch %d", tc.epoch)
```
nidhi-singh02, line 63: Q: Do we want the tests to be executed after assertion failure ? Was wondering should we use assert or require.

trungnt1811: hi @nidhi-singh02, imo using assert is suitable for our test cases because each test case is independent, and we want to see the results for all the defined epochs

itsdevbear: we should use require not assert ya?

trungnt1811: @nidhi-singh02 @itsdevbear thanks guys, I just refactored it so please help me review it again

itsdevbear (approved):

PR #1553, opened 2024-06-20
feat(server-v2): yippee

archbear:


ocnc2 (commented):
@archbear tmr we should:
1. refactor abci handlers to adhere to v2 function signature
2. figure out a way to expose our middleware handlers to the comet server, either by building passing these to the comet server on build, pulling these out of our BKNode, or through the AppManager

itsdevbear:
@ocnc2 we should have the comet consensus call the "middleware", then middleware should be stripped of all the comet / cosmos specific types, and become the "event orchestrator" of some kinda. 


Comet "Consensus" -> Calls Middleware -> Emits Events 
Rollkit "Consensus" -> Calls Middleware -> Emits Events 


something like this.

beacond/cmd/main.go
```diff
@@ -44,44 +47,59 @@ func run() error {
...
 			append(
 				clicomponents.DefaultClientComponents(),
 				// TODO: remove these, and eventually pull cfg and chainspec
 				// from built node
 				nodecomponents.ProvideNoopTxConfig,
 				nodecomponents.ProvideConfig,
 				nodecomponents.ProvideChainSpec,
-			),
+				nodecomponents.ProvideTxCodec[T],
+				servercomponents.ProvideCometServer[NodeT, T, ValidatorUpdateT],
```
ocnc2, line 89: @archbear your depinject unexported error is because were sending a provider that accepts generics btw. I'm assuming this is because the depinject graph isn't built at runtime, so depinject won't be able to resolve the function type. 

so unfortunately we have to inject the hard types whenever we set the providers 🙁

ocnc2 (commented):
commenting out pruning flags until this is merged:

https://github.com/cosmos/cosmos-sdk/pull/20736

ocnc2 (commented):
waiting on https://github.com/cosmos/cosmos-sdk/pull/20788 as well

mod/node-core/pkg/app/export.go
```diff
@@ -21,54 +21,51 @@
...
+func (app *BeaconApp[T]) ExportAppStateAndValidators(
 	forZeroHeight bool,
 	_, modulesToExport []string,
 ) (servertypes.ExportedApp, error) {
-	// as if they could withdraw from the start of the next block
-	ctx := app.NewContextLegacy(
-		true,
-		cmtproto.Header{Height: app.LastBlockHeight()},
-	)
+	panic("not supported, just look at the genesis file u goofy")
```
itsdevbear, line 34: based

archbear:
https://github.com/berachain/beacon-kit/pull/1653

:pepe-not-yippee:

PR #1558, opened 2024-06-21
chore: remove true comparision

TropicalDog17:
## Description

Remove instances of redundant logical conditions that alway result in true.

## Summary by CodeRabbit

- **Refactor**
	- Simplified file path handling in the CLI application for better efficiency and readability.
	- Improved the logic for decoding hexadecimal nibbles by removing redundant conditions.


mod/cli/pkg/commands/jwt/jwt.go
```diff
@@ -117,16 +117,14 @@ func getFilePath(cmd *cobra.Command, path string) (string, error) {
...
-		specifiedFilePath = filepath.Join(
-			clientCtx.HomeDir+"/config/", DefaultSecretFileName,
-		)
+	clientCtx, ok := cmd.Context().
+		Value(client.ClientContextKey).(*client.Context)
+	if !ok {
+		return "", ErrNoClientCtx
 	}
+	specifiedFilePath = filepath.Join(
+		clientCtx.HomeDir+"/config/", DefaultSecretFileName,
```
archbear, line 126: nit: filepath.Join appends the backslashes correctly according to os spec so hardcoding /config/ isn't necessary if i remember correctly

archbear (approved):
nit otherwise lgtm

mod/primitives/pkg/hex/nibble.go
```diff
@@ -28,11 +28,11 @@ import (
 func decodeNibble(in byte) uint64 {
 	// uint64 conversion here is safe
 	switch {
-	case in >= '0' && in <= '9' && in >= hexBaseOffset:
```
itsdevbear, line 31: what's this change?

mod/cli/pkg/commands/jwt/jwt.go
```diff
@@ -117,16 +118,14 @@ func getFilePath(cmd *cobra.Command, path string) (string, error) {
 	// If no path is specified, try to get the cosmos client context and use
 	// the configured home directory to write the secret to the default file
 	// name.
-	if specifiedFilePath == "" {
```
itsdevbear, line 120: i don't think this is logically equivalent

ocnc2: is it not?

mod/primitives/pkg/hex/nibble.go
TropicalDog17: hexBaseOffset is already '0' so no need for comparison here. Same with the other `if` statements

mod/cli/pkg/commands/jwt/jwt.go
TropicalDog17: we already early return if `specifiedFilePath` is not empty [here](https://github.com/berachain/beacon-kit/blob/056a0a5f53a6898672ad44048dd997693a8a66e6/mod/cli/pkg/commands/jwt/jwt.go#L114) so no need to check again.

ocnc2 (approved):

mod/primitives/pkg/hex/nibble.go
ocnc2: you're right. wanna add a brief comment clarifying that this would never underflow?

ocnc2: or maybe above each return, add a comment containing the expected range of the result. 

ex. `in - hexBaseOffset` would result in a value in 0-9, `in - hexAlphaOffsetUpper` would result in a value in 10-15, etc.

itsdevbear (approved):
yup I agree with all points, 

@TropicalDog17 just add a comment for what ocnc2 and i'm good to merge

ocnc2 (approved):
lgtm thanks @TropicalDog17!

PR #1561, opened 2024-06-21
chore(node-core): cleanup + prep for server v2

archbear:


mod/cli/pkg/config/client.go
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: BUSL-1.1
+//
```
itsdevbear, line 2: what was your plan? would a client folder make more sense?

archbear: eh no pref i just moved this cause i hate utils.go as a concept and its like just a convenience func for cosmos client config

itsdevbear:
lgtm

PR #1568, opened 2024-06-22
test(consensus-types): add more unit tests for ExecutableDataDeneb

minhngoc274:
Partially address #1190


mod/consensus-types/pkg/types/payload_test.go
```diff
@@ -89,7 +103,15 @@ func TestExecutableDataDeneb_GetTree(t *testing.T) {
 
 func TestExecutableDataDeneb_Getters(t *testing.T) {
 	payload := generateExecutableDataDeneb()
-
+	transactions := make([][]byte, 1)
+	transactions[0] = make([]byte, 0)
```
nidhi-singh02, line 107: Is this needed here ? I don't see it is used.

mod/consensus-types/pkg/types/payload_test.go
```diff
@@ -89,7 +103,15 @@ func TestExecutableDataDeneb_GetTree(t *testing.T) {
...
+	withdrawals := make([]*engineprimitives.Withdrawal, 1)
+	withdrawals[0] = &engineprimitives.Withdrawal{
```
nidhi-singh02, line 109: Same for withdrawals[0] , do we need it at this place ?

mod/consensus-types/pkg/types/payload_test.go
minhngoc274: I use this to make sure the logic when this array has elements works

minhngoc274: Here the logic that it will cover

minhngoc274: Here is the logic that this will cover err = ssz.UnmarshalDynamic(buf, num, func(indx int, buf []byte) (err error) {
			if len(buf) > 1073741824 {
				return ssz.ErrBytesLength
			}
			if cap(e.Transactions[indx]) == 0 {
				e.Transactions[indx] = make([]byte, 0, len(buf))
			}
			e.Transactions[indx] = append(e.Transactions[indx], buf...)
			return nil
		})

minhngoc274: Here is the line that it covers https://github.com/berachain/beacon-kit/blob/main/mod/consensus-types/pkg/types/payload_header.ssz.go#L217-L227

mod/consensus-types/pkg/types/payload_test.go
minhngoc274: Same with the logic above, here the line that it covers: https://github.com/berachain/beacon-kit/blob/main/mod/consensus-types/pkg/types/payload.ssz.go#L241-L247

itsdevbear (approved):

nidhi-singh02 (approved):

itsdevbear (approved):

itsdevbear (approved):

PR #1626, opened 2024-06-26
feat(ssz): SSZTypes

itsdevbear:
## Summary by CodeRabbit

- **New Features**
  - Introduced new SSZ (Simple Serialize) types for boolean, uint8, uint16, uint32, uint64, and byte.
  - Added methods for marshalling and unmarshalling SSZ types.
  - Implemented SSZ vector and list handling for basic and composite types.
  - Introduced `Merkleizer` interface for merkleizing SSZ types.

- **Bug Fixes**
  - Updated `MarshalSSZ` method in `BasicItem` for byte slice creation to use `binary.LittleEndian.PutUint64`.

- **Tests**
  - Added test cases for SSZ vector functionalities.
  - Replaced individual marshaling tests with fuzzing functions for better coverage.

- **Refactor**
  - Streamlined serialization logic by removing redundant marshal and unmarshal functions.

- **Chores**
  - Updated dependencies and import paths for consistency.


nidhi-singh02 (approved):
lgtm, nitpick: let's remove commented code in test

kocubinski:
@itsdevbear with regard to schema definition and parsing they’re separate concerns, in [schema.go](https://github.com/kocubinski/beacon-kit/blob/f499ff860a40c15fce5a9b73b4bc93ace766c616/mod/storage/pkg/sszdb/schema/schema.go#L173) one parser could be substituted for another.  In other words if the beacon monolith defines files with these types a slightly different parser is needed to get the reflection right.

It looks like that PR could end up being part of a substitute for fastssz though which is probably good.

On the other hand, if you commit to wrapping all types in an SSZ monolithic struct maybe reflection wouldn't even be needed?  The generic impl could return the metadata required to build the schema? Not sure yet.

PR #1642, opened 2024-06-28
fix - breaking change in ethereum-package

nidhi-singh02:
starting a network was failing due to changes in ethereum-package


nidhi-singh02:
I was getting below error. Not sure how it is working without these changes hence closing it.
Screenshot 2024-06-28 at 11 28 30 AM

PR #1660, opened 2024-06-30
feat(storage,ssz): introduce SSZ schema and backend

kocubinski:
**Goal**: Replace IAVL usage with a performant SSZ db backend.  Instead of storing the leaves of an SSZ merkle tree in an AVL tree, store the SSZ tree directly with nodes keyed on their generalized index.  This has powerful implications for disk, memory and compute usage thanks to the caching and partial hashing properties of the resultant structure, i.e. less usage for far more throughput. 

**Status**: Hacked but working, PoC quality.

One of the more complex (and indirected) pieces of state, `ExecutionPayloadHeader` is being stored in both `sszdb` and IAVL with a double write.  A double read verifies the hash of the objects on each `Get` through the block lifecycle.  Verified with `make start`.


itsdevbear:
🧑‍🍳 🔥

itsdevbear:
closing in favour of: https://github.com/berachain/beacon-kit/pull/1720

PR #1672, opened 2024-07-01
deposit transaction with create-validator

nidhi-singh02:
WIP

### DIFF
--- a/mod/cli/pkg/utils/deposit/BeaconDepositContract.go
+++ b/mod/cli/pkg/utils/deposit/BeaconDepositContract.go
@@ -1,0 +1,1021 @@
+// Code generated - DO NOT EDIT.
+// This file is a generated binding and any manual changes will be lost.
+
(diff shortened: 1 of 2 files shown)


nidhi-singh02:
I realised don't have to generate the go bindings for the Deposit contract, it is already there. I could re-use it.

mod/cli/pkg/utils/deposit/BeaconDepositContract.go
```diff
@@ -0,0 +1,1021 @@
+// Code generated - DO NOT EDIT.
```
nidhi-singh02, line 1: Noticed don't have to generate the go bindings for the Deposit contract, it is already there. I could re-use it.

nidhi-singh02:
Noticed don't have to generate the go bindings for the Deposit contract, it is already there. Will re-use it by creating a new PR.

PR #1688, opened 2024-07-04
submit deposit transaction on create-validator cmd

nidhi-singh02:
Deposit transaction is successful with all EL clients such as geth, besu, nethermind, erigon, reth and ethereumjs.

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced deposit transaction management with improved efficiency.
  
- **Bug Fixes**
  - Improved error handling for deposit transactions and private key requirements.
  
- **Refactor**
  - Updated flag names and default values for better clarity and usability.
  
- **Chores**
  - Updated and removed dependencies for streamlined project management.


nidhi-singh02:
As reth was not working due to `Method Not Found` error, we have changed the approach - we are creating the ethClient using `NewFromRPCClient`, it works with all the EL clients except ethereumjs.

testing/files/eth-genesis.json
archbear, line 1: revert these changes to the genesis files

contracts/lib/forge-std
archbear, line 1: fix sub modules diff

archbear (changes requested):

mod/cli/pkg/commands/deposit/create.go
```diff
@@ -21,24 +21,35 @@
...
+	"github.com/berachain/beacon-kit/mod/execution/pkg/client/ethclient"
+	"github.com/berachain/beacon-kit/mod/geth-primitives/pkg/deposit"
+	"github.com/berachain/beacon-kit/mod/geth-primitives/pkg/rpc"
 	"github.com/berachain/beacon-kit/mod/node-core/pkg/components"
 	"github.com/berachain/beacon-kit/mod/node-core/pkg/components/signer"
 	"github.com/berachain/beacon-kit/mod/primitives/pkg/common"
+	"github.com/berachain/beacon-kit/mod/primitives/pkg/constraints"
 	"github.com/berachain/beacon-kit/mod/primitives/pkg/crypto"
 	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
```
archbear, line 41: also get rid of the go-ethereum imports from here and use geth-primitives instead

archbear (dismissed):

testing/files/eth-genesis.json
nidhi-singh02: Below is the latest bytecode, and without these changes if I submit a transaction, getting error `deposit transaction failed`. I suspect reason could be deposit binding is generated from the ABI of the current contract.

contracts/lib/forge-std
nidhi-singh02: I ran `forge update` earlier and tried again, still see diff. My local sub-module is pointing to the correct version it seems.

mod/cli/pkg/commands/deposit/create.go
nidhi-singh02: addressed.

build/scripts/testing.mk
```diff
@@ -173,12 +173,15 @@ start-ethereumjs:
 	-p 30303:30303 \
 	-p 8545:8545 \
 	-p 8551:8551 \
-	ethpandaops/ethereumjs:stable \
+	ethpandaops/ethereumjs:master \
```
ocnc, line 176: Is there a reason we're using master instead of stable?

mod/cli/pkg/commands/deposit/create.go
```diff
@@ -139,55 +151,167 @@ func createValidatorCmd(
...
+	// Parse the private key.
+	var (
+		fundingPrivKey string
+		privKey        *ecdsa.PrivateKey
+	)
+	fundingPrivKey, err := cmd.Flags().GetString(privateKey)
+	if err != nil {
+		return common.ExecutionHash{}, err
+	}
+	if fundingPrivKey == "" {
```
archbear, line 190: probably one of these checks is not required since "" => nil pk i believe

mod/cli/pkg/commands/deposit/create.go
```diff
@@ -139,55 +151,167 @@ func createValidatorCmd(
...
+		return common.ExecutionHash{}, ErrPrivateKeyEmpty
+	}
+
+	// Parse the execution client RPC URL.
+	rpcURL, err := cmd.Flags().GetString(rpcURL)
+	if err != nil {
+		return common.ExecutionHash{}, err
+	}
+	eth1ChainID := new(big.Int).SetUint64(chainSpec.DepositEth1ChainID())
+	depositContractAddress := chainSpec.DepositContractAddress()
```
archbear, line 208: nit: i'm not a fan of having this variable, i don't think it adds any clarity

mod/cli/pkg/commands/deposit/create.go
```diff
@@ -139,55 +151,167 @@ func createValidatorCmd(
...
+				_ common.ExecutionAddress, tx *gethprimitives.Transaction,
+			) (*gethprimitives.Transaction, error) {
+				return gethprimitives.SignTx(
+					tx,
+					gethprimitives.LatestSignerForChainID(eth1ChainID),
+					privKey,
+				)
+			},
+			//nolint:mnd // The gas tip cap.
+			// It is necessary for besu to work, not sure why though.
```
archbear, line 245: remove these comments

also the reason why they're required is the client is just not inferring a valid set of gas parameters so we need to specify it manually

archbear (commented):
reconcile the contracts submodules and nits otherwise lgtm

build/scripts/testing.mk
archbear: dialing stable doesn't connect correctly but does with master

nidhi-singh02: Yeah arch is right, seems some issue with stable, it is like 6 months old.

mod/cli/pkg/commands/deposit/create.go
nidhi-singh02: hehe, removed :D Thanks!

mod/cli/pkg/commands/deposit/create.go
nidhi-singh02: I removed above check, and on testing by not passing privateKey and setting broadcast flag as true, getting error `2024/07/15 22:18:43 ERROR startup failure error="invalid length, need 256 bits"` which is not giving clarity I believe. This check specifically tells users that the private key is required.

mod/cli/pkg/commands/deposit/create.go
```diff
@@ -139,55 +151,163 @@ func createValidatorCmd(
 			"signature", signature.String(),
 		)
 
-		// TODO: once broadcast is fixed, remove this.
-		logger.Info("Send the above calldata to the deposit contract 🫡")
+		if broadcast {
+			var txHash common.ExecutionHash
```
archbear, line 155: add a log to indicate that the transaction is being broadcasted

mod/cli/pkg/commands/deposit/create.go
```diff
@@ -139,55 +151,163 @@ func createValidatorCmd(
...
+		return common.ExecutionHash{}, err
+	}
+	if fundingPrivKey == "" {
+		return common.ExecutionHash{}, ErrPrivateKeyRequired
+	}
+	privKey, err = gethprimitivescrypto.HexToECDSA(fundingPrivKey)
+	if err != nil {
+		return common.ExecutionHash{}, err
+	}
+
```
archbear, line 197: nit: remove the newline here

archbear (approved):
minor nits but otherwise lgtm

ocnc (approved):
Looks good now

PR #1709, opened 2024-07-06
feat(ssz): basic schema construction and tests

kocubinski:
Given an object schema representing the shape of an SSZ monolith object this small library answers the following question:

- "What is the generalized index of the tree node identified by this [object path](https://github.com/ethereum/consensus-specs/blob/v1.4.0/ssz/merkle-proofs.md#ssz-object-to-index)?"

The byte offset within the referenced node is also returned.

## Summary by CodeRabbit

- **New Features**
  - Introduced functionality for defining and working with SSZ (Simple Serialize) types, supporting basic, container, and enumerable types such as lists and vectors.
  - Added methods for size calculation, chunking, and traversal within SSZ structures.

- **Tests**
  - Added test cases for schema handling focusing on paths, indices, traversal, and indexing within structured schemas.


mod/primitives/pkg/encoding/ssz/schema/schema.go
```diff
@@ -0,0 +1,205 @@
...
+
+type Node struct {
+	SSZType
+
+	GIndex uint64
+	Offset uint8
+}
+
+func Path(path string) []string {
+	return strings.Split(path, "/")
```
itsdevbear, line 158: What was your rationale again for not doing []int everywhere?

itsdevbear: we could create a hard type 

ObjectPath []int

itsdevbear (approved):

mod/primitives/pkg/encoding/ssz/schema/schema.go
kocubinski: I like how the spec described object paths this way, they're human readable and writable.  I think its easier for a dev to write and maintain e.g. `"validator/n"`vs `10/n` when e.g. [retrieving a validator by index](https://github.com/berachain/beacon-kit/blob/d69a1941513c852e7b10ff4d67db6cfd561ad772/mod/storage/pkg/beacondb/registry.go#L128). 

Also the length node for lists should be directly addressable somehow, right now the token `"__len__"` is used, for e.g. [GetTotalValidators](https://github.com/berachain/beacon-kit/blob/d69a1941513c852e7b10ff4d67db6cfd561ad772/mod/storage/pkg/beacondb/registry.go#L169).

Either way can work though.

itsdevbear: I'm good with following spec!

mod/primitives/pkg/encoding/ssz/schema/schema.go
```diff
@@ -0,0 +1,205 @@
...
+func (e enumerable) IsByteVector() bool {
+	return e.Element.Size() == 1 && e.length > 0
+}
+
+func (e enumerable) IsList() bool {
+	return e.maxLength > 0
+}
+
+func (e enumerable) IsFixed() bool {
+	_, ok := e.Element.(basic)
```
itsdevbear, line 146: this is wrong, what is the element is a Vector of ints

itsdevbear (changes requested):
acc I think there are a few bugs

mod/primitives/pkg/encoding/ssz/schema/schema.go
kocubinski: what are you seeing? I added a test for a vector of 128 bit ints, I think it's correct.

kocubinski: oh, well from the spec:
>We recursively define "variable-size" types to be lists, unions, Bitlist and all types that contain a variable-size type.
so to determine if a vector is fixed length the entire subtree should be recursed for any var sized type? I've removed it for now since I don't think there's a use case from the schema side.

itsdevbear:
closing in favour of: https://github.com/berachain/beacon-kit/pull/1720

PR #1710, opened 2024-07-06
chore(merkle):  Re-use bytes buffer for merkle hasher

calbera:
@itsdevbear Key here is that `leaves` or the `chunks` are currently NOT re-used. but within the `RootHasher` we can reuse the bytes buffer for the hash's **digests** (or `outputList`). This output can be re-used indefinitely across hashing different leaves.

Next thing to check would be whether we can set the chunks from a re-usable bytes buffer (may be worth trying go 1.23 built-in chunkify as well). 

## Summary by CodeRabbit

- **New Features**
  - Enhanced memory efficiency in the hashing process by utilizing a reusable buffer for hash output storage in the `RootHasher` struct.
  - Improved performance by eliminating redundant copying of results in the `BuildParentTreeRootsWithNRoutines` function.

- **Tests**
  - Renamed `BenchmarkHasherWithReusableBuffer` function to `BenchmarkHasher` and updated performance metrics.


itsdevbear:
@calbera can you add a benchmark as well as triple check to make sure that it syncs with bArtio?

calbera:
@itsdevbear yeah, still need to fix it because we're still doing the `make` [here](https://github.com/berachain/beacon-kit/pull/1710/files#diff-d14d60c37a4268e020bf8b177658c797f418f3a0715db9011f0efb5eef1d54f1R118-R120).

itsdevbear:
yeah need to add Put() as well then so think this should be good

calbera:
@itsdevbear updated the benchmark and synced from genesis, all good

itsdevbear (approved):

PR #1713, opened 2024-07-07
DNM WIP bug example

itsdevbear:


itsdevbear:
should be 6 actually 



image

PR #1715, opened 2024-07-07
devs ssz ideas

itsdevbear:


mod/primitives/pkg/encoding/ssz/proof/object_path.go
```diff
@@ -0,0 +1,117 @@
...
+		if !ok {
+			return nil, fmt.Errorf("expected string variable name for Container type, got %T", indexOrVariableName)
+		}
+		return element.(Container[SSZType]).GetFieldByName(variableName), nil
+	case types.Elements:
+		index, ok := indexOrVariableName.(uint64)
+		if !ok {
+			return nil, fmt.Errorf("expected string variable name for Container type, got %T", indexOrVariableName)
+		}
+		return element.(Elements).ElementAtIndex(index), nil
```
kocubinski, line 34: this is a bit problematic, when traversing (for example) a very large list, gindex calculation would need the _nth_ element to be set when in reality just the element type is needed.

PR #1742, opened 2024-07-15
test(consensus-types): add more test for block_denebplus

hoank101:
## Summary by CodeRabbit

- **Tests**
	- Added comprehensive tests for the `BeaconBlockDenebPlus` type, including versioning, state root setting, body retrieval, marshaling/unmarshaling, and tree generation.


ocnc (commented):
Would be good if the tests also tested a case where attestation data is not empty, as thats what deneb+ adds

hoank101:
> Would be good if the tests also tested a case where attestation data is not empty, as thats what deneb+ adds

yeah, I just updated the test, thank you

ocnc (approved):
lgtm

PR #1759, opened 2024-07-18
refactor(beacon): decouple from cosmos

archbear:


contracts/lib/forge-std
itsdevbear, line 1: undo these changes

itsdevbear (approved):
lgtm, just fix the contracts dep issue

PR #1800, opened 2024-07-28
feat(eip4788): Beacon Block Prover Contract

calbera:
Used by RootHelper contract. Removing unused code.

Future TODOs:
- add proving eth block number `n` for a given timestamp (of eth block `n + 1`)
- add a function that combines both proofs into 1 external

## Summary by CodeRabbit

- **New Features**
	- Introduced `BeaconVerifier` smart contract for verifying validator public keys in the beacon chain.
	- Added new `IBeaconVerifier` interface to enhance beacon chain interactions.
	- Implemented functions for Merkle tree operations using SHA256 in the `SSZ` library.
	- Added constant for expected validator public key length and new error handling for validation.

- **Bug Fixes**
	- Improved error handling and cryptographic proof validation in the `BeaconVerifier` contract.

- **Documentation**
	- Enhanced documentation for new functions and updated comments for clarity and consistency.


contracts/src/eip4788/BeaconVerifier.sol
```diff
@@ -0,0 +1,123 @@
...
+            proposerIndex
+        );
+    }
+
+    /// @notice Verifies the validator pubkey is in the registry of beacon state.
+    /// @param beaconBlockRoot `bytes32` root of the beacon block.
+    /// @param validatorPubkeyProof `bytes32[]` proof of the validator.
+    /// @param validatorPubkey `ValidatorPubkey` to verify.
+    /// @param validatorIndex `uint64` index of the validator.
+    function proveValidatorPubkeyInBeaconBlock(
```
sj448, line 96: instead of proving only pubkey, imo we should keep the logic of proving whole validator struct, helps us to prove other properties of validator, will be useful when we implement withdrawal/slashing

sj448 (approved):
implementation looks good

contracts/src/eip4788/BeaconVerifier.sol
calbera: every proof will require its unique gindex, proof, leaf, etc. Will implement each one as needed

PR #1816, opened 2024-07-30
feat (full node) : Kurtosis package to sync bArtio node 

nidhi-singh02:


nidhi-singh02:
At first, running the nodes (geth and beacond) using docker. Once the parameters and how to restore from snapshot is figured out, will create kurtosis package around it.

nidhi-singh02 (commented):
At first, running the nodes (geth and beacond) using docker. Once the parameters and how to restore from snapshot is figured out, will create kurtosis package around it.

PR #1818, opened 2024-07-30
feat(node-api): Optimizations usage of math.U64

calbera:
- Use math.U64 types 
- TODO: LRU cache the query contexts (to optimize for 1 endpoint handler calling backend multiple times, should only incur 1 query context)

## Summary by CodeRabbit

- **New Features**
	- Enhanced type safety for slot handling across various backend methods, transitioning from `uint64` to `math.Slot`.
	- Updated multiple backend interfaces to use `math.Slot` and `math.Epoch` for consistent slot and epoch parameter management.
	- Improved mock implementation of `BlockStore` to reflect new return types for slot retrieval.

- **Bug Fixes**
	- Improved error handling by refining internal logic for parsing slots and epochs.

- **Chores**
	- Updated utility functions to support new data types, ensuring consistency across the application.


mod/node-api/backend/block.go
```diff
@@ -46,26 +45,21 @@ func (b Backend[
...
 	st, slot, err := b.stateFromSlot(slot)
 	if err != nil {
 		return common.Root{}, err
 	}
 
 	// As calculated by the beacon chain. Ideally, this logic
 	// should be abstracted by the beacon chain.
-	index := slot % b.cs.SlotsPerHistoricalRoot()
-	return st.GetBlockRootAtIndex(index)
+	return st.GetBlockRootAtIndex(slot.Unwrap() % b.cs.SlotsPerHistoricalRoot())
```
calbera, line 56: this Unwrap is needed since chain spec returns uint64. this unwrap also is done in every StateProcesso for every slot.

mod/node-api/handlers/utils/id.go
```diff
@@ -71,5 +73,9 @@ func SlotFromBlockID[StorageBackendT interface {
 	if err != nil {
 		return 0, err
 	}
-	return storage.GetSlotByRoot(bytes.ToBytes32(root))
+	slot, err := storage.GetSlotByRoot(bytes.ToBytes32(root))
```
itsdevbear, line 76: just make this return math.Slot

PR #1819, opened 2024-07-30
perf(node-api): Add caching of query contexts to backend

calbera:
LRU cache the query contexts (to optimize for 1 endpoint handler calling backend multiple times, should only incur 1 query context)

## Summary by CodeRabbit

- **New Features**
	- Introduced a caching mechanism for query contexts, improving performance and efficiency.
	- Added configuration options for cache size and expiration settings.
	- Enhanced configuration management for the Node API, allowing for modular customization.

- **Bug Fixes**
	- Enhanced context management to prevent unintended state mutations.

- **Tests**
	- Implemented comprehensive unit tests for the new caching functionality, ensuring reliability across various scenarios.
	- Updated tests to validate new initialization values for the ExecutionPayloadHeader structure.

- **Documentation**
	- Added documentation detailing the new cache configuration and usage.


mod/node-api/backend/backend.go
```diff
@@ -115,9 +116,10 @@ func New[
 		NodeT, StateStoreT, StorageBackendT, ValidatorT, ValidatorsT, WithdrawalT,
 		WithdrawalCredentialsT,
 	]{
-		sb: storageBackend,
-		cs: cs,
-		sp: sp,
+		sb:    storageBackend,
+		cs:    cs,
+		sp:    sp,
+		cache: cache.NewQueryCacheWithDefaultConfig(),
```
archbear, line 122: if we're doing a whole config might be worthwhile to upstream these changes to an api config and have the api config divide into server + backend instead of server only which is what we have now, could be useful to be able to configure this

archbear (approved):

mod/node-api/backend/cache/cache.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"context"
+
+	"github.com/berachain/beacon-kit/mod/primitives/pkg/math"
+	lru "github.com/hashicorp/golang-lru/v2/expirable"
+)
+
+// QueryCache is a cache for query contexts.
+type QueryCache struct {
```
itsdevbear, line 31: I get the motivation here, but I am ensure if this is actually doing anything, you'll have to do some benchmarks / traces

itsdevbear (approved):

mod/node-api/backend/cache/cache.go
calbera: yea we don't have traces set up anywhere. was thinking to use this lib https://github.com/open-telemetry/opentelemetry-go

calbera:
postponing this till we have tracers enabled and live trace data

PR #1824, opened 2024-07-30
feat(node-api): Allow querying blocks by timestamp

calbera:
Also prune from the roots and timestamps stores

### DIFF
--- a/contracts/src/eip4788/interfaces/IBeaconVerifier.sol
+++ b/contracts/src/eip4788/interfaces/IBeaconVerifier.sol
@@ -1,8 +1,6 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.21;
 
-import { SSZ } from "../SSZ.sol";
-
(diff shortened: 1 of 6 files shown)


calbera:
TODO: also introduce timestamp_id (which is queryable by slot + with timestamps optionally as well)

contracts/src/eip4788/SSZ.sol
```diff
@@ -32,14 +32,6 @@ library SSZ {
         }
     }
 
-    function uint64HashTreeRoot(uint64 v)
-        internal
-        pure
-        returns (bytes32 root)
-    {
-        return toLittleEndian(uint256(v));
```
itsdevbear, line 40: write a toLittleEndianU64 impl that will save a bunch of gas.

archbear (approved):
utack makes sense

itsdevbear (approved):
lgtm

PR #1834, opened 2024-08-01
Randomized serialization of beacon block body (karalabe/ssz)

sbudella-gco:
* `types.BeaconBlockBody` needs custom generator in order to avoid nil pointer dereference during serialization.
* The test currently fails due to discrepancy after a serialization-deserialization round trip, namely empty slices get deserialized to `nil`.

Tests fail with `Deserialize: deserialized body different than former body after serialization`
Further investigation reveals that serialized empty slices get deserialized as `nil`. For instance:
image
The same applies to `Deposits`, `Withdrawals` and similar objects.

The culprit seems to be this line: https://github.com/karalabe/ssz/blob/026e3979cc6d91bbd573a6b7d012738615326fca/decoder.go#L616

which instead should be something like:
```go
*blobs = make([]T, 0)
```


itsdevbear:
attaching issue in peter's repo https://github.com/karalabe/ssz/issues/18

abi87:
Fixed in the upstream, closing this down

PR #1854, opened 2024-08-07
feat(Deposit Contract) : Pubkey to operator mapping

sj448:
## Summary by CodeRabbit

- **New Features**
	- Introduced enhanced operator management capabilities, including functions for requesting, canceling, and accepting operator changes.
	- Streamlined deposit process by removing the amount parameter from the deposit function.
	- Enhanced visibility of deposit authorization counts for better transparency.

- **Bug Fixes**
	- Improved error handling with new specific error types related to operator actions.

- **Documentation**
	- Updated function signatures for clarity and added documentation where necessary.

- **Refactor**
	- Simplified function signatures and improved readability in the contract code.


itsdevbear:
I don't think this code should live in BeaconKit

contracts/src/staking/BeaconDepositContract.sol
```diff
@@ -29,36 +29,56 @@ contract BeaconDepositContract is IBeaconDepositContract, Ownable {
     /// @dev The length of the credentials, 1 byte prefix + 11 bytes padding + 20 bytes address = 32 bytes.
     uint8 private constant CREDENTIALS_LENGTH = 32;
 
+    uint256 private constant TWO_DAYS = 172_800; // 2 days
```
ocnc, line 32:
```suggestion
    uint256 private constant TWO_DAYS = 172_800; // 2 days in seconds.
```

contracts/src/staking/BeaconDepositContract.sol
```diff
@@ -80,7 +100,13 @@ contract BeaconDepositContract is IBeaconDepositContract, Ownable {
             revert InvalidSignatureLength();
         }
 
-        uint64 amountInGwei = _deposit(amount);
+        // Update the pubkey to operator mapping if first deposit.
+        if (_pubkeyToOperator[pubkey] == address(0)) {
+            _pubkeyToOperator[pubkey] = msg.sender;
```
ocnc, line 105: i feel like this would be more extensible if we just allowed for operator to be passed in as an arg.

calbera:
> I don't think this code should live in BeaconKit

image

cc @sj448 @neverDefined

testing/e2e/e2e_staking_test.go
```diff
@@ -85,25 +84,24 @@ func (s *BeaconKitE2ESuite) TestDepositRobustness() {
...
-		From:   genesisAccount.Address(),
-		Signer: genesisAccount.SignerFunc(chainID),
-	})
-	s.Require().NoError(err)
-
-	// Wait for the transaction to be mined.
-	receipt, err := bind.WaitMined(s.Ctx(), s.JSONRPCBalancer(), tx)
-	s.Require().NoError(err)
-	s.Require().Equal(coretypes.ReceiptStatusSuccessful, receipt.Status)
+	// tx, err := dc.InitializeOwner(&bind.TransactOpts{
```
sj448, line 87: @calbera can remove this commented code

contracts/src/staking/BeaconDepositContract.sol
sj448: that will just make operator change process one step, i feel it should be two step process for validators to change their operator address, its better security wise.

sj448:
> I don't think this code should live in BeaconKit

There is no other place where we can define pubkey-operator mapping without signature verification, keeping it here saves the sig verification part as whoever make the first deposit get assigned as the operator.

testing/e2e/e2e_staking_test.go
sj448: removed it.

itsdevbear:
yeah imo this should be its own thing, the deposit contract can stay very minimal

contracts/src/staking/BeaconDepositContract.sol
neverDefined: one step is fine imo they already comitting capital so this should be verified before hand

itsdevbear:
closing for now

contracts/src/staking/BeaconDepositContract.sol
```diff
@@ -104,11 +130,80 @@ contract BeaconDepositContract is IBeaconDepositContract, Ownable {
         external
         onlyOwner
     {
+        // If the number is non-zero, set it to zero to avoid front-running by depositors to make more deposits.
+        if (depositAuth[depositor] != 0) {
+            depositAuth[depositor] = 0;
+            return;
+        }
```
calbera, line 137: Can fully remove the `allowDeposit` as it'll be permissionless on mainnet

PR #1860, opened 2024-08-08
perf(storage): remove fmt.Sprintf in the loop

hoank101:
## Summary by CodeRabbit

- **Improvements**
	- Enhanced the clarity of the deletion process in the application by refining how paths are constructed for deleting a range of items.

### DIFF
--- a/mod/storage/pkg/filedb/range_db.go
+++ b/mod/storage/pkg/filedb/range_db.go
@@ -94,7 +94,8 @@ func (db *RangeDB) DeleteRange(from, to uint64) error {
 		return errors.New("rangedb: delete range not supported for this db")
 	}
 	for ; from < to; from++ {
-		if err := f.fs.RemoveAll(fmt.Sprintf("%d/", from)); err != nil {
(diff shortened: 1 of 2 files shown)


mod/storage/pkg/filedb/range_db.go
```diff
@@ -94,7 +94,8 @@ func (db *RangeDB) DeleteRange(from, to uint64) error {
 		return errors.New("rangedb: delete range not supported for this db")
 	}
 	for ; from < to; from++ {
-		if err := f.fs.RemoveAll(fmt.Sprintf("%d/", from)); err != nil {
+		path := strconv.FormatUint(from, 10) + "/"
```
nidhi-singh02, line 97: Curious : What performance benefit do we get.

ocnc (approved):

mod/storage/pkg/filedb/range_db.go
ocnc: its within the realm of ns but several multipliers faster

PR #1910, opened 2024-08-16
feat(async): dispatcher

ocnc2:
## Summary by CodeRabbit

- **New Features**
  - Improved error handling capabilities with new tailored error types for enhanced clarity.
  - Introduced multiple new error handling functions for more informative timeout messages.

- **Bug Fixes**
  - Corrected the description of an existing error type to provide accurate context.

- **Refactor**
  - Updated dependency management to facilitate local development and streamline module architecture.
  - Enhanced error handling structure within the middleware for better context and reporting.

- **Chores**
  - Updated several dependencies to the latest versions to improve performance and stability.


mod/async/pkg/types/event.go
```diff
@@ -25,56 +25,72 @@ import (
...
+// BaseEvent defines the minimal interface that the dispatcher expects from a
+// message.
+type BaseEvent interface {
+	ID() EventID
+	Context() context.Context
+}
+
+// Event defines the interface that the underlying route expects from a
+// message with data.
+type Event[DataT any] interface {
```
itsdevbear, line 40: Arch and I are thinking this maybe goes into primitives

mod/beacon/blockchain/service.go
```diff
@@ -164,134 +166,154 @@ func (s *Service[
...
 		)
+		panic(err)
 	}
 }
 
 func (s *Service[
 	_, BeaconBlockT, _, _, _, _, _, _, _, _,
-]) handleBeaconBlockReceived(
-	msg *asynctypes.Event[BeaconBlockT],
+]) handleVerifyBeaconBlockEvent(
```
itsdevbear, line 256: this should still be received imo, bc ignoring comet, this is the first time this node as seen the block. We are verifiying the block as a result of receving it, it shouldnt be labelled as processing a request.

ocnc2: yeah agree

mod/async/pkg/types/event.go
ocnc2: yep that makes sense

mod/async/pkg/types/event.go
```diff
@@ -25,56 +25,72 @@ import (
...
+	BaseEvent
+	Data() DataT
+	Error() error
+	Is(id EventID) bool
 }
 
 // NewEvent creates a new Event with the given context and beacon event.
 func NewEvent[
```
itsdevbear, line 48: imo constructor should be below the type def, also move type to events @archbear you still think thats the right call?

mod/beacon/block_store/pruner.go
```diff
@@ -17,17 +17,18 @@
...
+	"github.com/berachain/beacon-kit/mod/primitives/pkg/events"
+)
 
 // BuildPruneRangeFn builds a function that returns the range of blocks to
 // prune.
 func BuildPruneRangeFn[BeaconBlockT BeaconBlock](
 	cfg Config,
-) func(*asynctypes.Event[BeaconBlockT]) (uint64, uint64) {
-	return func(event *asynctypes.Event[BeaconBlockT]) (uint64, uint64) {
+) func(events.Event[BeaconBlockT]) (uint64, uint64) {
```
itsdevbear, line 30: events.Event is typically frowned upon since.
`<x>s.<X>` bad import pattern

mod/beacon/blockchain/execution_engine.go
```diff
@@ -53,6 +53,8 @@ func (s *Service[
 
 // sendNextFCUWithAttributes sends a forkchoice update to the execution
 // client with attributes.
+//
```
itsdevbear, line 56: revert

mod/beacon/blockchain/service.go
```diff
@@ -164,134 +166,154 @@ func (s *Service[
...
+			"error", err,
+		)
 	}
 }
 
 func (s *Service[
 	_, BeaconBlockT, _, _, _, _, _, _, _, _,
-]) handleBeaconBlockFinalization(
-	msg *asynctypes.Event[BeaconBlockT],
+]) handleFinalBeaconBlockReceived(
```
itsdevbear, line 284: I would revert this, since the fact that we are "receiving" the block again from CometBFT is a CometBFT specific thing, in a lot of other consensus engines, we might just load a stored beacon block, that was stored during verifiy.

mod/async/pkg/types/event.go
ocnc2: oops forgot to delete this file

mod/beacon/block_store/pruner.go
ocnc2: yeah true. Do you think we should rename the package from primitives/events -> primitives/async, or just create an "asynctypes/async" alias every time we import

itsdevbear (approved):

PR #1970, opened 2024-08-23
feat(kurtosis) : Deploy berps contract

nidhi-singh02:


itsdevbear:
@nidhi-singh02 this doesn't belong here, reach out to cal

PR #2016, opened 2024-09-20
nits

abi87:
A couple of small nits I found while studying the codebase:

- a duplicated config
- we never call `KVStore`'s `SchemaBuilder.Build()`. `SchemaBuilder`s are required to instantiate collections but not really used in their working (see https://docs.cosmos.network/v0.47/build/packages/collections#schemabuilder). Still why not getting all the checks that `Build` performs? Performances hit should be minimal and happens only at start up.

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling during the schema building process for the KVStore, improving reliability.
- **Bug Fixes**
	- Updated the KVStore initialization to ensure proper error reporting.
	- Removed unnecessary configuration settings in the DefaultCometConfig function for better clarity.


nidhi-singh02 (approved):

itsdevbear (commented):
should we make the fn return an error instead of panic?

cli/builder/config.go
```diff
@@ -58,7 +58,6 @@ func DefaultCometConfig() *cmtcfg.Config {
 	cfg.Mempool.Recheck = false
 	cfg.Mempool.Broadcast = false
 	cfg.Storage.DiscardABCIResponses = true
-	cfg.Storage.DiscardABCIResponses = true
```
abi87, line 61: dropped duplicated config

PR #2021, opened 2024-09-25
chore(cometBFTService): some more minor cleanups

abi87:
I have collected here some minor cleanups around CometBFT service:

1. dropped `loadLatest bool` since I believe we must call `LoadLatestVersion()` in order to initialize the stores. Failing of doing that causes a panic
2. dropped `SetName` since it is not called and name is set in service constructor
3. dropped `SetVersion` since it's simple the assignment of a service attribute that can be done inline
4. modified `StoreKeyServiceProvider` to return a pointer rather than a pointer to a pointer. It's important to avoid copying the key, but passing around the pointer should be enough. All consumers of the key deference the double pointer.

I collected the changes together in the interest of reducing the number of PRs. Happy to further split them if reviewers require it.

### DIFF
--- a/mod/consensus/pkg/cometbft/service/service.go
+++ b/mod/consensus/pkg/cometbft/service/service.go
@@ -87,7 +87,6 @@ func NewService[
 	logger LoggerT,
 	db dbm.DB,
 	middleware MiddlewareI,
-	loadLatest bool,
 	cmtCfg *cmtcfg.Config,
 	cs common.ChainSpec,
 	options ...func(*Service[LoggerT]),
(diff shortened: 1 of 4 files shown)


mod/consensus/pkg/cometbft/service/service.go
```diff
@@ -115,11 +114,9 @@ func NewService[
...
 	}
 
-	// Load the s.
-	if loadLatest {
-		if err := s.sm.LoadLatestVersion(); err != nil {
-			panic(err)
-		}
+	// Load latest height, once all stores have been set
+	if err := s.sm.LoadLatestVersion(); err != nil {
+		panic(err)
```
abi87, line 119: we must do this, otherwise beconKit will panic. The `if loadLatest` is confusing to me, as it suggest this initialization is not mandatory

itsdevbear: I think its safe to remove, I think loadLatest was for if u want to instantiate a the cosmos BaseApp without initializing the DB, I don't see why we need this tho.

itsdevbear (approved):

PR #2022, opened 2024-09-25
chore(DefaultServiceOptions): minor chainID loading refactoring

abi87:
chainID is either specified by flag or read by genesis file. In handling the genesis file, we currently have a couple of gosec warnings around fliepath sanitization and closing the file, which I believe are easily solvable. This PR does that.

Ignoring errors in closing a file may be [dangerous](https://www.joeshaw.org/dont-defer-close-on-writable-files/). This is not the case for genesis ( we only read the file, not write it).  Still I wanted to introduce the error handling to solicit conversation around how to generally handle it in the overall codebase

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method for loading the chain ID from the genesis file, enhancing error handling and code organization. 
- **Refactor**
	- Streamlined the implementation of the `DefaultServiceOptions` function by removing inline logic and improving readability.

### DIFF
--- a/mod/node-core/pkg/builder/baseapp_options.go
+++ b/mod/node-core/pkg/builder/baseapp_options.go
@@ -58,27 +58,12 @@ func DefaultServiceOptions[
 		panic(err)
 	}
 
-	homeDir := cast.ToString(appOpts.Get(flags.FlagHome))
+	// get chainID, possibly falling back to genesis if flag is not set
 	chainID := cast.ToString(appOpts.Get(flags.FlagChainID))
-	var reader *os.File
 	if chainID == "" {
-		// fallback to genesis chain-id
-		//#nosec:G304 // bet.
-		reader, err = os.Open(filepath.Join(homeDir, "config", "genesis.json"))
+		chainID, err = loadChainIDFromGenesis(appOpts)
 		if err != nil {
 			panic(err)
-		}
-		//#nosec:307 // bet.
-		defer reader.Close()
-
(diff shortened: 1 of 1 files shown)


mod/node-core/pkg/builder/baseapp_options.go
```diff
@@ -98,3 +84,28 @@ func DefaultServiceOptions[
...
 	}
 }
+
+func loadChainIDFromGenesis(appOpts config.AppOptions) (string, error) {
+	var (
+		homeDir = cast.ToString(appOpts.Get(flags.FlagHome))
+		fp      = filepath.Join(homeDir, "config", "genesis.json")
+	)
+
+	f, err := os.Open(filepath.Clean(fp))
```
abi87, line 94: Found [here](https://securego.io/docs/rules/g304.html) the suggestion to fix it up

mod/node-core/pkg/builder/baseapp_options.go
```diff
@@ -98,3 +84,28 @@ func DefaultServiceOptions[
...
+		return "",
+			errors.Join(
+				f.Close(),
+				fmt.Errorf(
+					"failed to parse chain-id from genesis file: %w",
+					err,
+				),
+			)
+	}
+	return chainID, f.Close()
```
abi87, line 110: here we avoid defer f.Close() (which ignores the error causing the gosec warning) and we explicitly return the (unlikely) f.Close error to the caller.

itsdevbear (approved):

PR #2024, opened 2024-09-26
chore(node-api): Refactor code to improve readability

nidhi-singh02:


mod/node-api/engines/echo/validator/constants/constants.go
```diff
@@ -0,0 +1,45 @@
...
+const (
+	StatusPendingInitialized = "pending_initialized"
+	StatusPendingQueued      = "pending_queued"
+	StatusActiveOngoing      = "active_ongoing"
+	StatusActiveExiting      = "active_exiting"
+	StatusActiveSlashed      = "active_slashed"
+	StatusExitedUnslashed    = "exited_unslashed"
+	StatusExitedSlashed      = "exited_slashed"
+	StatusWithdrawalPossible = "withdrawal_possible"
+	StatusWithdrawalDone     = "withdrawal_done"
```
tskoyo, line 44: I think it's more memory-efficient to use integers rather than strings for constants

nidhi-singh02: This is to validate the `statuses` in the API request against these strings specified on the right, hence cannot change it to integers. If I change, I have to maintain another mapping, which may be redundant.

mod/node-api/engines/echo/validator/validation.go
```diff
@@ -0,0 +1,106 @@
...
+		return false
+	}
+	if valid {
+		return true
+	}
+	if ValidateUint64(fl) {
+		return true
+	}
+	return false
+}
```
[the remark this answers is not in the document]
nidhi-singh02: If it is error, then I think it makes sense to return and not proceed with another validation. I will keep the check as is, I see similar pattern used in few other places, don't want to omit the error.

PR #2028, opened 2024-09-28
chore(CometBFTService): simplified cometBFTService Info

abi87:
Simplified `Service.Info` to avoid `CreateQueryContext` call.
The proposed implementation of `Service.Info` differs more from cosmos-sdk Info implementation,  but it is simpler and marginally faster

### DIFF
--- a/mod/consensus/pkg/cometbft/service/abci.go
+++ b/mod/consensus/pkg/cometbft/service/abci.go
@@ -39,6 +39,7 @@ import (
 	cmtproto "github.com/cometbft/cometbft/api/cometbft/types/v1"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	sdkversion "github.com/cosmos/cosmos-sdk/version"
 	"github.com/sourcegraph/conc/iter"
 )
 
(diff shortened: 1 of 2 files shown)


mod/consensus/pkg/cometbft/service/abci.go
```diff
@@ -187,21 +188,18 @@ func (s *Service[LoggerT]) Info(
 	*cmtabci.InfoRequest,
 ) (*cmtabci.InfoResponse, error) {
 	lastCommitID := s.sm.CommitMultiStore().LastCommitID()
-	appVersion := InitialAppVersion
+	appVersion := initialAppVersion
 	if lastCommitID.Version > 0 {
-		ctx, err := s.CreateQueryContext(lastCommitID.Version, false)
-		if err != nil {
-			return nil, fmt.Errorf("failed creating query context: %w", err)
-		}
```
abi87, line 195: dropping this is the main change of this PR

mod/consensus/pkg/cometbft/service/abci.go
```diff
@@ -39,6 +39,7 @@ import (
 	cmtproto "github.com/cometbft/cometbft/api/cometbft/types/v1"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	sdkversion "github.com/cosmos/cosmos-sdk/version"
```
abi87, line 42: aligned with the other usage of this package in our codebase

mod/consensus/pkg/cometbft/service/service.go
```diff
@@ -54,7 +53,10 @@ const (
 	execModeFinalize
 )
 
-const InitialAppVersion uint64 = 0
+const (
+	initialAppVersion uint64 = 0
+	appName           string = "beacond"
```
abi87, line 58: I prefer the name here since it's constant. Happy to revert the change if we don't agree with it

mod/consensus/pkg/cometbft/service/service.go
```diff
@@ -75,8 +76,6 @@ type Service[
 	initialHeight        int64
 	minRetainBlocks      uint64
 
-	// application's version string
-	version string
```
abi87, line 79: does not seem necessary to make it an attribute of node since it's value won't change upon execution

mod/consensus/pkg/cometbft/service/service.go
```diff
@@ -184,6 +181,10 @@ func (s *Service[_]) CommitMultiStore() storetypes.CommitMultiStore {
 
 // AppVersion returns the application's protocol version.
 func (s *Service[_]) AppVersion(_ context.Context) (uint64, error) {
+	return s.appVersion()
+}
+
+func (s *Service[_]) appVersion() (uint64, error) {
```
abi87, line 187: allows reusing this in `Info`

itsdevbear (approved):

PR #2033, opened 2024-10-02
chore(encoding): dropped go-json direct dependency

abi87:
Replaced `go-json` with `encoding/json` library.
Note that `go-json` is still an indirect dependency, but only an indirect dependency.

## Summary by CodeRabbit

- **New Features**
	- Transitioned to using the standard Go JSON encoding and decoding functionalities, enhancing compatibility and performance.

- **Bug Fixes**
	- Removed unnecessary dependency on the `gojson` library, streamlining the module's requirements.


nidhi-singh02 (approved):

ocnc (approved):
nice

PR #2038, opened 2024-10-03
chore(consensus-types): Make slashings use math.Gwei

abi87:
Make `Slashings` and `TotalSlashing` share the same type.

## Summary by CodeRabbit

- **New Features**
	- Updated the representation of slashing amounts in various components to a more specialized type, enhancing clarity and consistency.

- **Bug Fixes**
	- Adjusted the handling of slashing amounts across multiple methods to ensure compatibility with the new data type.

- **Refactor**
	- Standardized the data types used in interfaces and methods related to slashing, improving overall type safety and maintainability.


mod/consensus-types/pkg/types/state.go
```diff
@@ -69,7 +69,7 @@ type BeaconState[
 	NextWithdrawalValidatorIndex math.ValidatorIndex
 
 	// Slashing
-	Slashings     []uint64
+	Slashings     []math.Gwei
```
abi87, line 72: this is the main point of the PR. Make `Slashing` and `TotalSlashing` share the same type

mod/consensus-types/pkg/types/state.go
```diff
@@ -378,7 +378,7 @@ func (st *BeaconState[
 	}
 	subIndx = hh.Index()
 	for _, i := range st.Slashings {
-		hh.AppendUint64(i)
+		hh.AppendUint64(uint64(i))
```
abi87, line 381: q: should I use `i.Unwrap()` here instead? It seems the principled way, but it makes a copy, while casting should not.  We do use the uint64 cast for TotalSlashing as well (line 392)

itsdevbear (approved):

PR #2042, opened 2024-10-04
chore(consensus-type): Fixed ExecutionPayload json marshalling

abi87:
The `ExecutionPayload` struct implements `MarshalJSON` on a pointer receiver instead of on a value receiver.
Due to how the json library handles these definitions, using `json.Marshal` on a value receiver will result in attempting to use the default marshalling behavior instead of the custom defined `MarshalJSON` method.

This could result in unexpected bytes or invalid round trip conversions if the value receiver is ever used to marshal via json.

## Summary by CodeRabbit

- **New Features**
	- Enhanced JSON marshaling and unmarshaling processes for the `ExecutionPayload` struct.
	- Updated the type of the `BaseFeePerGas` field for improved data handling.

- **Bug Fixes**
	- Improved nil value checks for several fields during unmarshalling, ensuring data integrity.

- **Tests**
	- Added a new test to verify JSON marshaling behavior for both value and pointer types of `ExecutionPayload`.


mod/consensus-types/pkg/types/payload.go
```diff
@@ -273,7 +273,7 @@ func (p *ExecutionPayload) GetTree() (*fastssz.Node, error) {
 /* -------------------------------------------------------------------------- */
 
 // MarshalJSON marshals as JSON.
-func (p *ExecutionPayload) MarshalJSON() ([]byte, error) {
```
abi87, line 276: q: not a fan of mixing value and pointer receivers. Should we consider having the whole struct using value receivers (despite all the coping around)?

itsdevbear: I would bias towards always using pointers. Copying full blocks gets expensive.

nidhi-singh02 (approved):

PR #2045, opened 2024-10-05
exploratory: dropping `hex.String`

abi87:
WIP, working on defining the wanted end state, will break this into smaller PR as soon as I have done.


mod/primitives/pkg/encoding/hex/bytes.go
```diff
@@ -28,31 +28,36 @@ import (
...
-func EncodeBytes[B ~[]byte](b B) []byte {
-	result := make([]byte, len(b)*2+prefixLen)
-	copy(result, prefix)
-	hex.Encode(result[prefixLen:], b)
-	return result
+func FromBytes[B ~[]byte](b B) string {
+	return prefix + hex.EncodeToString(b)
 }
 
-func UnmarshalByteText(input []byte) ([]byte, error) {
```
abi87, line 38: replaced by `ToBytes` since:

1. `formatAndValidateText` is equivalent to `IsValidHex` since:
  - inputs are never zero-lenght
  - The check on lenght being even is implicitly done by `DecodeString`
2. `hex.Decode` behaviour with  `dec := make([]byte, len(raw)/encDecRatio)` matches `hex.DecodeString(`

mod/primitives/pkg/encoding/hex/validation.go
```diff
@@ -20,64 +20,65 @@
...
+		return nil, err
 	}
 
-	if len(input) == 0 {
-		return *new(T), ErrEmptyNumber
-	}
-	if len(input) > 1 && input[0] == '0' {
-		return *new(T), ErrLeadingZero
+	if len(input)%2 != 0 {
+		return nil, ErrOddLength
```
abi87, line 81: q: shouldn't this be part of `ValidateBasicHex` but some UTs currently fail

abi87:
Replaced by https://github.com/berachain/beacon-kit/pull/2048

PR #2062, opened 2024-10-10
chore(state-transition): improving `applyDeposit` error handling

abi87:


mod/state-transition/go.sum
```diff
@@ -1,3 +1,11 @@
+cosmossdk.io/api v0.7.0 h1:QsEMIWuv9xWDbF2HZnW4Lpu1/SejCztPu0LQx7t6MN4=
+cosmossdk.io/api v0.7.0/go.mod h1:kJFAEMLN57y0viszHDPLMmieF0471o5QAwwApa+270M=
```
itsdevbear, line 2: introducing cosmos deps to state-transition is bad

abi87: I agree, but I believe we have to properly handle the errors from state, however unlikely they are. Will work on finding a different way for this

PR #2066, opened 2024-10-11
feat(node-api): use timestamp ID instead of execution ID for proof namespace

calbera:
Proofs are indexed by timestamp in EIP 4788, and not execution block numbers. Consequently this allows offchain clients to easily query for beacon block proofs by timestamps, which is directly applicable for use in EL.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `GetTimestamp` method in the `BeaconBlock` structure to retrieve the timestamp from the execution payload.
	- Added `Timestamp` field in the `ExecutionPayload` structure.
	- Updated methods across various components to focus on timestamps instead of execution numbers.

- **Bug Fixes**
	- Corrected method references and assertions in tests to align with the new timestamp functionality.

- **Refactor**
	- Renamed multiple methods and parameters to reflect the shift from execution IDs to timestamp IDs.
	- Updated the `KVStore` structure to replace execution numbers with timestamps.
	- Removed deprecated methods related to execution numbers and introduced new methods for handling timestamps.


mod/consensus-types/pkg/types/block_test.go
```diff
@@ -43,6 +43,7 @@ func generateValidBeaconBlock() *types.BeaconBlock {
 		Body: &types.BeaconBlockBody{
 			ExecutionPayload: &types.ExecutionPayload{
 				Number:    10,
+				Timestamp: 10,
```
abi87, line 46: nit: consider dropping Number if it's not relevant anymore. It makes the UT marginally clearer.

mod/storage/pkg/block/store.go
```diff
@@ -86,19 +86,17 @@ func (kv *KVStore[BeaconBlockT]) GetSlotByBlockRoot(
...
+func (kv *KVStore[BeaconBlockT]) GetParentSlotByTimestamp(
+	timestamp math.U64,
 ) (math.Slot, error) {
-	slot, ok := kv.executionNumbers.Peek(executionNumber)
-	if !ok {
-		return 0, fmt.Errorf(
-			"slot not found at execution number: %d",
-			executionNumber,
-		)
+	slot, _ := kv.timestamps.Peek(timestamp)
```
abi87, line 94: nit: rename slot to timestamp or t

mod/storage/pkg/block/store.go
```diff
@@ -86,19 +86,17 @@ func (kv *KVStore[BeaconBlockT]) GetSlotByBlockRoot(
...
+	if slot == 0 {
+		return slot, fmt.Errorf("slot not found at timestamp: %d", timestamp)
 	}
-	return slot, nil
+
+	return slot - 1, nil
```
abi87, line 99: q: seems we are assuming a block is produced every second here. I don't think we can't simply subtract 1 to get parent's block timestamp

mod/storage/pkg/block/store.go
```diff
@@ -32,9 +32,9 @@ import (
 // KVStore is a simple memory store based implementation that stores metadata of
 // beacon blocks.
 type KVStore[BeaconBlockT BeaconBlock] struct {
-	blockRoots       *lru.Cache[common.Root, math.Slot]
-	executionNumbers *lru.Cache[math.U64, math.Slot]
-	stateRoots       *lru.Cache[common.Root, math.Slot]
+	blockRoots *lru.Cache[common.Root, math.Slot]
+	timestamps *lru.Cache[math.U64, math.Slot]
+	stateRoots *lru.Cache[common.Root, math.Slot]
```
calbera, line 37: @abi87 all of these mappings are actually in the form of some key -> slot value. In this case we store blockRoot -> slot, timestamp -> slot, stateRoot -> slot. Maybe its better to rename these more explicitly, along the lines of  `blockRootsToSlots` or similar

mod/storage/pkg/block/store.go
calbera: value returned is the slot corresponding to timestamp key

mod/storage/pkg/block/store.go
calbera: returns the parent slot for the given timestamp key. as long as slot != 0, which is checked, we should be good to return slot -  1.

abi87: @calbera , you're right, this is fine!
I have only one concern left: we are assuming that different blocks have different timestamps here.
This is guaranteed by CometBFT for finalized block (see [specs](https://docs.cometbft.com/main/spec/consensus/time)), so I need a pass just to make sure we never store here, non finalized blocks (forks may have the same timestamp and mess things up).

mod/storage/pkg/block/store.go
abi87: Naming as is works fine for me.

mod/storage/pkg/block/store.go
abi87: You're right

mod/storage/pkg/block/store.go
abi87: store.Set seems to be called upon onFinalizedBlock only so all good

mod/node-api/handlers/proof/block_proposer.go
```diff
@@ -26,8 +26,9 @@ import (
 	"github.com/berachain/beacon-kit/mod/node-api/handlers/utils"
 )
 
-// GetBlockProposer returns the block proposer pubkey for the given block id
-// along with a merkle proof that can be verified against the beacon block root.
+// GetBlockProposer returns the block proposer pubkey for the given timestamp
+// id along with a merkle proof that can be verified against the beacon block
```
abi87, line 30: nit: `for a give timestamp along`. Drop id

mod/node-api/handlers/proof/execution_fee_recipient.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 // GetExecutionFeeRecipient returns the fee recipient from the latest execution
-// payload header for the given block id, along with the proof that can be
```
abi87, line 30: nit: `for the given timestamp, along`. Drop id

mod/node-api/handlers/utils/id.go
```diff
@@ -70,36 +70,37 @@ func SlotFromBlockID[StorageBackendT interface {
...
 }
 
-// SlotFromExecutionID returns a slot from the execution number ID.
+// ParentSlotFromTimestampID returns the parent slot corresponding to the
+// timestamp ID.
 //
-// NOTE: `executionID` shares the same semantics as `stateID`, with the
-// modification of being able to query by beacon block <executionNumber>
-// instead of <stateRoot>.
+// NOTE: `timestampID` shares the same semantics as `stateID`, with the
```
abi87, line 76: nit: here and in the rest of the package I would use `timestamp` instead of `timestampID`

mod/node-api/handlers/utils/id.go
```diff
@@ -70,36 +70,37 @@ func SlotFromBlockID[StorageBackendT interface {
...
-	GetSlotByExecutionNumber(executionNumber math.U64) (math.Slot, error)
-}](executionID string, storage StorageBackendT) (math.Slot, error) {
-	if !IsExecutionNumberPrefix(executionID) {
-		return slotFromStateID(executionID)
+// The <timestamp> must be prefixed by the 't', followed by the timestamp
+// in decimal UNIX notation. For example 't1728681738' corresponds to the slot
+// which has the next block with a timestamp of 1728681738. Providing just the
+// string '1728681738' (without the prefix 't') will query for the beacon block
+// for slot 1728681738.
+func ParentSlotFromTimestampID[StorageBackendT interface {
```
abi87, line 85: nit: `ParentSlotFromTimestamp` instead of `ParentSlotFromTimestampID`

mod/node-core/pkg/components/interfaces.go
```diff
@@ -1090,7 +1091,7 @@ type (
 		ChainSpec() common.ChainSpec
 		GetSlotByBlockRoot(root common.Root) (math.Slot, error)
 		GetSlotByStateRoot(root common.Root) (math.Slot, error)
-		GetSlotByExecutionNumber(executionNumber math.U64) (math.Slot, error)
+		GetParentSlotByTimestamp(executionNumber math.U64) (math.Slot, error)
```
abi87, line 1094: nit: rename argument to `timestamp`

mod/node-core/pkg/components/interfaces.go
```diff
@@ -293,10 +295,9 @@ type (
 		GetSlotByBlockRoot(root common.Root) (math.Slot, error)
 		// GetSlotByStateRoot retrieves the slot by a given root from the store.
 		GetSlotByStateRoot(root common.Root) (math.Slot, error)
-		// GetSlotByExecutionNumber retrieves the slot by a given execution
-		// number
-		// from the store.
-		GetSlotByExecutionNumber(executionNumber math.U64) (math.Slot, error)
+		// GetParentSlotByTimestamp retrieves the parent slot by a given
+		// timestamp from the store.
+		GetParentSlotByTimestamp(executionNumber math.U64) (math.Slot, error)
```
abi87, line 300: nit: rename argumento to `timestamp`

abi87 (changes requested):
logic seems fine to me.
Found a few nits to cleanup, major of which is the `slot == 0` case handling.
Ready to approve as soon as nits are cleared

mod/node-api/handlers/proof/block_proposer.go
calbera: @abi87 I would prefer using timestampID here since the beacon spec uses the blockID, stateID terminology. Similarly here timestampID conveys to the caller that they have to provide not just the timestamp but also prefix it with `t`. This function also then parses for timestampID with the `t` prefix, so providing just the timestamp would mistakenly return.

mod/node-api/handlers/utils/id.go
calbera: The reason why timestampID is different from timestamp is explained in the comment above this function

calbera: follows a similar convention to what the beacon spec has set up for `state_id` and `block_id`, for ex https://ethereum.github.io/beacon-APIs/#/Beacon/getStateFork

mod/storage/pkg/block/store.go
```diff
@@ -86,19 +86,17 @@ func (kv *KVStore[BeaconBlockT]) GetSlotByBlockRoot(
...
 ) (math.Slot, error) {
-	slot, ok := kv.executionNumbers.Peek(executionNumber)
-	if !ok {
-		return 0, fmt.Errorf(
-			"slot not found at execution number: %d",
-			executionNumber,
-		)
+	slot, _ := kv.timestamps.Peek(timestamp)
+	if slot == 0 {
+		return slot, fmt.Errorf("slot not found at timestamp: %d", timestamp)
```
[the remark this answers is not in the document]
calbera: sure, don't mind being explicit here

mod/storage/pkg/block/store.go
calbera: sounds good, will include this comment.

mod/storage/pkg/block/store.go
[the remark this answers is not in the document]
calbera: FYI, one of the robustness todos is to make this store persistent to handle node crashes/restarts, on which the LRU cache will no longer be needed.

abi87 (approved):

PR #2070, opened 2024-10-14
chore(execution): another defer ticker fix

abi87:
Duly stopping the ticker within the right goroutine

## Summary by CodeRabbit

- **Bug Fixes**
	- Corrected a typographical error in the comment of the `Start` method for the `EngineClient` struct.

- **Improvements**
	- Enhanced the logging mechanism in the `ReportingService` to ensure continuous version reporting until completion, improving service reliability.


mod/node-core/pkg/services/version/version.go
```diff
@@ -66,10 +66,9 @@ func (*ReportingService) Name() string {
 // Start begins the periodic logging of the chain version.
 func (v *ReportingService) Start(ctx context.Context) error {
 	ticker := time.NewTicker(v.reportingInterval)
-	defer ticker.Stop()
-
 	v.metrics.reportVersion(v.version)
 	go func() {
+		defer ticker.Stop()
```
abi87, line 71: this is the fix. Ticker should be stopped in the goroutine. If not ticker will be stopped as soon as `Start` terminates and `reportVersion` will never be called.

PR #2077, opened 2024-10-16
chore(storage): close storage iterators

abi87:
We left unclosed a few iterators once iteration is done.
The PR fixes the issue and add the first UTs of the package.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new method to register validators with their effective balances.
  - Added unit tests for managing balances and validators in the key-value store.

- **Bug Fixes**
  - Improved error handling in the management of validators and balances.

- **Refactor**
  - Enhanced iterator management for better error handling in the slashing functionality.

### DIFF
--- a/mod/node-core/pkg/components/depinject.go
+++ b/mod/node-core/pkg/components/depinject.go
@@ -51,7 +51,7 @@ type kvStoreService struct {
 }
 
 func (k kvStoreService) OpenKVStore(ctx context.Context) store.KVStore {
-	return newKVStore(sdk.UnwrapSDKContext(ctx).KVStore(k.key))
+	return NewKVStore(sdk.UnwrapSDKContext(ctx).KVStore(k.key))
 }
 
 // CoreKVStore is a wrapper of Core/Store kvstore interface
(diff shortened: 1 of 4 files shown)


mod/storage/pkg/beacondb/registry.go
```diff
@@ -138,27 +140,27 @@ func (kv *KVStore[
...
 
 	iter, err := kv.validators.Iterate(kv.ctx, nil)
 	if err != nil {
 		return nil, err
 	}
+	defer func() {
+		err = errors.Join(err, iter.Close())
+	}()
 
-	i := 0
```
abi87, line 150: not necessary for the context of the PR but cleaner

mod/node-core/pkg/components/depinject.go
```diff
@@ -62,7 +62,7 @@ type coreKVStore struct {
 
 // NewKVStore returns a wrapper of Core/Store kvstore interface
 // Remove once store migrates to core/store kvstore interface.
-func newKVStore(store storetypes.KVStore) store.KVStore {
+func NewKVStore(store storetypes.KVStore) store.KVStore {
```
abi87, line 65: I need this exported to duly initialize a store in unit tests

nidhi-singh02: I am wondering if it is a good practice to export a function, just for UTs.

abi87: Not great, but I think the worst is not having proper test coverage.
My bet with exporting this function is that other packages' unit tests will need this function to instantiate their in-memory DB and KVStore.
If I am wrong, we can always open a new PR, un export this function and structure `storage` unit tests differently.
What do you think?

itsdevbear (approved):

PR #2082, opened 2024-10-17
chore(consensus-types): Coherent `BeaconBlock` construction

abi87:
Let `NewWithVersion` and `NewFromSSZ` both err instead of panicking

### DIFF
--- a/mod/consensus-types/pkg/types/block.go
+++ b/mod/consensus-types/pkg/types/block.go
@@ -21,6 +21,8 @@
 package types
 
 import (
+	"fmt"
+
 	"github.com/berachain/beacon-kit/mod/primitives/pkg/common"
 	"github.com/berachain/beacon-kit/mod/primitives/pkg/math"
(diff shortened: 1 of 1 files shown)


mod/consensus-types/pkg/types/block.go
```diff
@@ -56,41 +58,38 @@ func (b *BeaconBlock) NewWithVersion(
...
 			StateRoot:     common.Root{},
 			Body:          &BeaconBlockBody{},
-		}
-	default:
-		return &BeaconBlock{}, ErrForkVersionNotSupported
+		}, nil
 	}
 
-	return block, nil
+	return nil, fmt.Errorf(
```
itsdevbear, line 71: Can we use errors.Wrap everywhere?

abi87: Done

itsdevbear (approved):

PR #2094, opened 2024-10-23
fix(beacon): consolidated next payload timestamp calculations

abi87:
- fixed `markRebuildPayloadForRejectedBlockFailure` metric
-  consolidated next payload timestamp calculations
- nits and typos

Note: the PR simply enforces a single function to calculate next payload time. The logic by which this timestamp is chosen will be revised in an upcoming PR

### DIFF
--- a/mod/beacon/blockchain/execution_engine.go
+++ b/mod/beacon/blockchain/execution_engine.go
@@ -23,7 +23,7 @@ package blockchain
 import (
 	"context"
 
-	blocktime "github.com/berachain/beacon-kit/mod/beacon/block-time"
+	payloadtime "github.com/berachain/beacon-kit/mod/beacon/payload-time"
 	engineprimitives "github.com/berachain/beacon-kit/mod/engine-primitives/pkg/engine-primitives"
 )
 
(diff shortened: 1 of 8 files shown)


mod/beacon/blockchain/payload.go
```diff
@@ -137,10 +128,10 @@ func (s *Service[
 		// and possibly should be made more explicit later on.
 		lph.GetParentHash(),
 	); err != nil {
-		s.metrics.markRebuildPayloadForRejectedBlockFailure(slot, err)
+		s.metrics.markRebuildPayloadForRejectedBlockFailure(stateSlot, err)
```
abi87, line 131: slot is instantiated and never written. I believe we want to log the slot we are trying to build a block for.

mod/beacon/blockchain/execution_engine.go
```diff
@@ -121,18 +122,3 @@ func (s *Service[
 		)
 	}
 }
-
```
abi87, line 124: moved to `block-time` package

mod/beacon/blockchain/process.go
```diff
@@ -107,8 +102,15 @@ func (s *Service[
 	defer s.metrics.measureStateTransitionDuration(startTime)
 	valUpdates, err := s.stateProcessor.Transition(
 		&transition.Context{
-			Context:          ctx,
+			Context: ctx,
+
+			// We set `OptimisticEngine` to true since this is called during
```
abi87, line 107: just moved here from above

PR #2095, opened 2024-10-23
fix(consensus, blockchain): set execution payload timestamp based on CometBFT timestamp

abi87:
We currently:

- setup execution payload timestamp based on node local clock;
- we don't really enforce any bound on execution payload timestamp (EL enforces just strict monotonicity) 
This is brittle and dangerous. 
A better alternative is to let consensus dictate time. In particular:

- For any block, let consensus service provide an estimation or proposal for next execution payload timestamp. The estimation MUST BE strictly monotonic, to make sure any payload building run (even optimistic ones, carried out before current block is finalized) would result in an execution payload with a valid timestamp.
- Use the proposed timestamp for next execution payload to bound current payload timestamp

In more details

- a `ConsensusBlock` type is introduced, which is meant to be a decorator around `BeconBlock` adding `CometBFT` based data (nextPayloadTimestamp in this PR: possibly ProposerAddress in the next future).
- Consensus time from a `ConsensusBlock` is used to calculate next execution payload
- Next Execution Payload timestamp is used to bound current payload timestamp and build next payload.

For reference:

- [here](https://github.com/cometbft/cometbft/blob/9b8eafa8b0279f5bfcb7876ccc463a3ad0f8a67c/state/state.go#L252) how CometBFT picks the next block timestamp
- [here](https://github.com/cometbft/cometbft/blob/9b8eafa8b0279f5bfcb7876ccc463a3ad0f8a67c/state/validation.go#L121) how CometBFT validates block timestamp

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `ConsensusBlock` type for enhanced handling of consensus-related operations.
	- Added `nextPayloadTimestamp` to various methods for improved payload management.
	- Updated middleware to manage execution payload delays effectively.

- **Bug Fixes**
	- Enhanced error handling for payload timestamps exceeding acceptable limits.

- **Refactor**
	- Modified several method signatures to incorporate new consensus block types and parameters.
	- Streamlined event handling logic to process consensus blocks.
	- Updated context handling to align with the Cosmos SDK.

- **Chores**
	- Removed outdated interfaces and functions to clean up the codebase.


mod/consensus/pkg/cometbft/service/middleware/abci.go
```diff
@@ -219,32 +215,35 @@ func (h *ABCIMiddleware[
...
 	}
 
 	// Request the blob sidecars.
-	if sidecars, err = encoding.
+	sidecars, err := encoding.
 		UnmarshalBlobSidecarsFromABCIRequest[BlobSidecarsT](
-		req, 1,
-	); err != nil {
+		req,
+		BlobSidecarsTxIndex,
```
abi87, line 238: nit: reuse constant

mod/consensus/pkg/cometbft/service/middleware/abci.go
```diff
@@ -219,32 +215,35 @@ func (h *ABCIMiddleware[
 	defer h.metrics.measureProcessProposalDuration(startTime)
 
 	// Request the beacon block.
-	if blk, err = encoding.
+	blk, err := encoding.
 		UnmarshalBeaconBlockFromABCIRequest[BeaconBlockT](
-		req, 0, h.chainSpec.ActiveForkVersionForSlot(math.U64(req.Height)),
-	); err != nil {
+		req,
+		BeaconBlockTxIndex,
```
abi87, line 221: nit: reuse constant

mod/consensus/pkg/cometbft/service/types.go
```diff
@@ -76,10 +70,3 @@ type SlotData[AttestationDataT, SlashingInfoT, SlotDataT any] interface {
 	// New creates a new slot data instance.
 	New(math.Slot, []AttestationDataT, []SlashingInfoT) SlotDataT
 }
-
-// StorageBackend defines an interface for accessing various storage components
-// required by the beacon node.
-type StorageBackend[BeaconStateT BeaconState] interface {
-	// StateFromContext retrieves the beacon state from the given context.
-	StateFromContext(context.Context) BeaconStateT
-}
```
abi87, line 85: seems only used in `helpers.go` file which has been commented out and dropped

beacond/cmd/types.go
```diff
@@ -63,14 +62,6 @@ import (
...
 
 type (
-	// ABCIMiddleware is a type alias for the ABCIMiddleware.
-	ABCIMiddleware = middleware.ABCIMiddleware[
-		*BeaconBlock,
-		*BlobSidecars,
-		*Genesis,
-		*SlotData,
-	]
-
```
abi87, line 73: I believe unnecessary once Middleware has been encapsulated in consensus service

beacond/cmd/types.go
```diff
@@ -114,9 +105,6 @@ type (
 	// CometBFTService is a type alias for the CometBFT service.
 	CometBFTService = cometbft.Service[*Logger]
 
-	// ConsensusMiddleware is a type alias for the consensus middleware.
-	ConsensusMiddleware = cometbft.MiddlewareI
-
```
abi87, line 119: I believe unnecessary once Middleware has been encapsulated in consensus service

mod/consensus/pkg/types/consensus_block.go
```diff
@@ -0,0 +1,54 @@
...
+
+package types
+
+import (
+	"time"
+
+	"github.com/berachain/beacon-kit/mod/primitives/pkg/math"
+)
+
+type ConsensusBlock[BeaconBlockT any] struct {
```
abi87, line 29: q: wonder if this should be `ConsensusContainer` rather than `ConsensusBlock`. Technically this is just a decorator that could be used for Sidecars as well. I call it block rn because we are not currently used this for sidecars, but we may.

abi87: `ConsensusBlock` is not a decorator strictly speaking, because it does not implement the `BeaconBlockT` interface.
This was done on purpose to try and break type asserts where they happen (especially around bus), but can be lifted if we want.

itsdevbear:
I presume this is only for non-optimistic builds?

mod/consensus/pkg/cometbft/service/abci.go
```diff
@@ -201,13 +201,19 @@ func (s *Service[LoggerT]) PrepareProposal(
...
+	var slotData *types.SlotData[
+		*ctypes.AttestationData,
+		*ctypes.SlashingInfo,
+	]
+	slotData = slotData.New(
+		math.Slot(req.Height),
+		nil,
+		nil,
+		math.U64(req.GetTime().Unix()),
+	)
```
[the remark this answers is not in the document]
abi87: This is fine, `New` set the nil pointer. `types.NewSlotData` does not exist

mod/consensus/pkg/cometbft/service/middleware/abci.go
```diff
@@ -219,32 +210,40 @@
...
 
 	// notify that the beacon block has been received.
-	if err = h.dispatcher.Publish(
-		async.NewEvent(ctx, async.BeaconBlockReceived, blk),
-	); err != nil {
+	var enrichedBlk *types.ConsensusBlock[BeaconBlockT]
+	enrichedBlk = enrichedBlk.New(
+		blk,
+		req.GetTime(),
+	)
```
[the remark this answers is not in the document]
abi87: It's fine. `New` set the empty pointer

mod/consensus/pkg/cometbft/service/middleware/abci.go
```diff
@@ -342,9 +337,14 @@ func (h *ABCIMiddleware[
...
 	// notify that the final beacon block has been received.
-	if err = h.dispatcher.Publish(
-		async.NewEvent(ctx, async.FinalBeaconBlockReceived, blk),
-	); err != nil {
+	// notify that the beacon block has been received.
+	var enrichedBlk *types.ConsensusBlock[BeaconBlockT]
+	enrichedBlk = enrichedBlk.New(
+		blk,
+		req.GetTime(),
+	)
```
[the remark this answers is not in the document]
abi87: It's fine

mod/consensus/pkg/types/slot_data.go
```diff
@@ -30,6 +30,8 @@ type SlotData[AttestationDataT, SlashingInfoT any] struct {
 	AttestationData []AttestationDataT
 	// SlashingInfo is the slashing info of the incoming slot.
 	SlashingInfo []SlashingInfoT
+	// Time tracked by consensus engine at the time SlotData is emitted.
+	ConsensusTime math.U64
```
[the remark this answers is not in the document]
abi87: Done

mod/consensus/pkg/types/consensus_block.go
```diff
@@ -0,0 +1,54 @@
...
+func (b *ConsensusBlock[BeaconBlockT]) New(
+	beaconBlock BeaconBlockT,
+	blkTime time.Time,
+) *ConsensusBlock[BeaconBlockT] {
+	b = &ConsensusBlock[BeaconBlockT]{
+		blk:           beaconBlock,
+		consensusTime: math.U64(blkTime.Unix()),
+	}
+	return b
+}
```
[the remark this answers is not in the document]
abi87: It's fine

mod/beacon/blockchain/execution_engine.go
```diff
@@ -44,7 +44,7 @@ func (s *Service[
 		return
 	}
 
-	if !s.shouldBuildOptimisticPayloads() && s.localBuilder.Enabled() {
```
abi87, line 47: `shouldBuildOptimisticPayloads` already checks `s.localBuilder.Enabled()`

abi87:
> non-optimistic

It's for optimisti builds as well. The whole idea is to avoid calls to `time.Now()` to setup execution payload time and use the CometBFT time instead.

abi87:
> I presume this is only for non-optimistic builds?

Actually is for any build. What I have done technically is replacing the call to `time.Now()` with `ConsensusTime`

mod/beacon/payload-time/time.go
```diff
@@ -21,23 +21,24 @@
...
+//     extra stability properties.
 func Next(
 	chainSpec common.ChainSpec,
 	parentPayloadTime math.U64,
+	consensusTime math.U64,
 ) uint64 {
-	//#nosec:G701 // not an issue in practice.
 	return max(
-		uint64(time.Now().Unix())+chainSpec.TargetSecondsPerEth1Block(),
+		uint64(consensusTime)+chainSpec.TargetSecondsPerEth1Block(),
```
abi87, line 41: if we assume consensusTime is monotonic, we could drop the `parentPayloadTime` as in my understanding, it is there only to ensure monotonicity.

mod/consensus/pkg/cometbft/service/middleware/middleware.go
```diff
@@ -68,16 +73,37 @@
...
+	// - Prepare/FinalizeBlock: to req.Time + minPayloadDelay
+	// Monotonicity across request sequences is ensured by construction of
+	// minPayloadDelay: no block can be finalized before minPayloadDelay.
+	minPayloadDelay := min(
+		cmtCfg.Consensus.TimeoutPropose,
+		cmtCfg.Consensus.TimeoutPrevote,
+		cmtCfg.Consensus.TimeoutPrecommit,
+		// TimeoutCommit can be zero
+		max(cmtCfg.Consensus.TimeoutCommit, time.Second),
+	)
```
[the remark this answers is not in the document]
abi87: It's fine. min and max are defined for OrderedTypes and time.Duration is currently a int64 alias

mod/beacon/blockchain/execution_engine.go
abi87: Stil the refactoring is not equivalent

mod/state-transition/pkg/core/state_processor_payload.go
```diff
@@ -96,14 +100,24 @@ func (sp *StateProcessor[
...
 	payload := body.GetExecutionPayload()
 
+	if pt := payload.GetTimestamp(); pt > nextPayloadTimestamp {
+		return errors.Wrapf(
+			ErrTooFarInTheFuture,
+			"payload timestamp, max: %d, got: %d",
+			nextPayloadTimestamp, pt,
+		)
+	}
+
```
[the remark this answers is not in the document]
abi87: This is not a problem. EL will ensure monotonicity

mod/consensus/pkg/types/consensus_block.go
calbera: Im okay with `ConsensusBlock` as its uniquely 1:1 to a certain slot num

mod/consensus/pkg/cometbft/service/middleware/abci.go
calbera, line 1: right now it looks like SlotData is used exclusively for PrepareProposal and ConsensusBlock is used exclusively for ProcessProposal. Is this always the case? If so, we should probably rename / explain their purposes more specifically

mod/consensus/pkg/cometbft/service/abci.go
```diff
@@ -204,13 +204,23 @@ func (s *Service[LoggerT]) PrepareProposal(
...
+		*ctypes.SlashingInfo,
+	]
+	slotData = slotData.New(
+		math.Slot(req.GetHeight()),
+		nil,
+		nil,
+
+		// we do not add h.minPayloadDelay here since req.GetTime()
+		// is guaranteed to be strictly larger than
+		// prevBlock.GetTime() + h.minPayloadDelay
```
calbera, line 218: can you quickly remind me why this is the case? Is it because we factor it in to the nextPayloadTime [here](https://github.com/berachain/beacon-kit/pull/2095/files#diff-877718cea28b2a3ddf39d2cf7b9d5ad968d1a209fc23017de98e6cb0870b6b2fR227)?

mod/consensus/pkg/types/slot_data.go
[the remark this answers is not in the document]
abi87: Slot is used when it's the node turn to propose a block.
Block is used when a block alreay built (locally or received from network) os to be verified or commetted.
We already used two different struts, I just extended them

mod/consensus/pkg/cometbft/service/middleware/abci.go
abi87: Correct. Names are not super telling but i would defer the renaming to a future PR to avoid inflating scope

mod/consensus/pkg/types/slot_data.go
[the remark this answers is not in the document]
calbera: the distinction between slot/block here kinda makes sense. but it should definitely be explained in the code for it to be clear and a standard to follow everywhere.

mod/consensus/pkg/cometbft/service/abci.go
abi87: Yes. The req.Time here is the time of the block to be built.
CometBFT consensus sets this time to be larger than the nextPayloadTime, as we set in Process/FinalizeBlock calls before (which were referring to the previous block)

calbera (approved):
LGTM. Just in a future PR we should maintain/explain the distinction so its clear for future between SlotData (PrepareProposal) and ConsensusBlock (ProcessProposal).

PR #2096, opened 2024-10-24
fix(state-transition): Validate execution payload timestamp

abi87:
Enforce that execution payload timestamp is bounded


mod/state-transition/pkg/core/state_processor_payload.go
```diff
@@ -96,14 +100,25 @@ func (sp *StateProcessor[
...
 	_, _, _, _, _, _, _, _, _, _, _, _, _,
-]) validateStatelessPayload(blk BeaconBlockT) error {
+]) validateStatelessPayload(
+	blk BeaconBlockT,
+	consensusTime math.U64,
+) error {
 	body := blk.GetBody()
 	payload := body.GetExecutionPayload()
 
+	timeBound := consensusTime + math.U64(sp.cs.TargetSecondsPerEth1Block())
```
abi87, line 110: I believe the bound does not have to be strict, i.e. we could enforce being not farther than $consensusTime + N \times Target$ seconds in the future

abi87: THIS IS CURRENTLY WRONG.
We set timestamp to `max(consensusTime)+TargetSecondsPerEth1Block, parentPayloadTime+1)`, so the timestamp does not have to be smaller than consensus + N * Target.

Such a validation can be enforced if timestamp == consensusTime

itsdevbear (approved):

PR #2104, opened 2024-10-26
bug(cli): hardened convert amount to avoid truncation

abi87:
Make sure big int can be turned into a uint64

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved validation for input amounts to ensure they are valid unsigned 64-bit integers, enhancing overall data integrity.


nidhi-singh02 (approved):
lgtm!

PR #2108, opened 2024-10-28
fix(beacon): Hardened sidecars validation

abi87:
Hardened sidecars validation by checking that their block header matches with the header of the block gossiped and verified with the sidecars.

Merged in https://github.com/berachain/beacon-kit/pull/2121 to fix e2e tests

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `ConsensusSidecars` structure for enhanced handling of consensus-related data.
  - Added an `Equals` method to the `BeaconBlockHeader` struct for header comparison.
  - Updated various service and middleware components to support new consensus sidecar types.

- **Bug Fixes**
  - Improved error handling in the `verifySidecars` method to ensure consistency between sidecars and block headers.

- **Documentation**
  - Enhanced type definitions and method signatures across multiple components for better clarity and usability.


mod/da/pkg/blob/verifier.go
```diff
@@ -72,18 +76,26 @@ func (bv *Verifier[_, _, BlobSidecarsT]) VerifySidecars(
 		bv.proofVerifier.GetImplementation(),
 	)
 
+	// check that sideracs block headers match with header of the
+	// corresponding block
+	for i, s := range sidecars.GetSidecars() {
+		if !reflect.DeepEqual(s.GetBeaconBlockHeader(), blkHeader) {
```
calbera, line 82: are we sure `reflect.DeepEqual` is good here? otherwise might make sense to add a Equality function on the blockHeader type

calbera (approved):
nit comments, LGTM

mod/da/pkg/blob/processor.go
```diff
@@ -84,10 +92,16 @@ func NewProcessor[
...
+	cs ConsensusSidecarsT,
 ) error {
-	startTime := time.Now()
+	var (
+		startTime = time.Now()
+		sidecars  = cs.GetSidecars()
+		blkHeader = cs.GetHeader()
+	)
 	defer sp.metrics.measureVerifySidecarsDuration(
 		startTime, math.U64(sidecars.Len()),
```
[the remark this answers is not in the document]
abi87: Done here and in other places where this is used

build/scripts/testing.mk
```diff
@@ -202,7 +202,7 @@ start-erigon: ## start an ephemeral `erigon` node
 	docker run \
     --rm -v $(PWD)/${TESTAPP_FILES_DIR}:/${TESTAPP_FILES_DIR} \
     -v $(PWD)/.tmp:/.tmp \
-    thorax/erigon:latest init \
+    erigontech/erigon:latest init \
```
abi87, line 205: merged https://github.com/berachain/beacon-kit/pull/2121 to unlock CI compilation

mod/da/pkg/blob/verifier.go
abi87: replaced with equals (and added simple UTs to avoid surprises)

PR #2110, opened 2024-10-28
perf(state-transition): reduce amount of withdrawals processed

abi87:
No need to process for withdrawal validators that are not fully or partially withdrawable

Note: this PR (and those built on top) seems to break once we update erigon from 2.60.1 to 2.60.9.
Geth seems to give the following error:
```sh
WARN [11-02|12:18:20.162] Served engine_newPayloadV3               conn=192.168.65.1:43721 reqid=1 duration="93.25µs" err="Invalid parameters" errdata="{\"err\":\"nil withdrawals post-shanghai\"}"
```

Note: keep on investigating with @dezzeus. It looks like withdrawals are duly set to an empty slice when the block is built, but as soon as the block is returned from consensus to middleware for verification, the slice is nil. SSZ serialization seems to be responsible for this. Investigation ongoing
Investigation ongoing

Note: found a solution: expliciting make Withdrawals slice when SSZ unmarshalling.


mod/state-transition/pkg/core/state_processor_staking_test.go
```diff
@@ -0,0 +1,183 @@
...
+		Slot:          genBlockHeader.GetSlot() + 1,
+		ProposerIndex: genBlockHeader.GetProposerIndex(),
+		ParentRoot:    genBlockHeader.HashTreeRoot(),
+		StateRoot:     common.Root{},
+		Body: &types.BeaconBlockBody{
+			ExecutionPayload: &types.ExecutionPayload{
+				Timestamp:     10,
+				ExtraData:     []byte("testing"),
+				Transactions:  [][]byte{},
+				Withdrawals:   []*engineprimitives.Withdrawal{}, // no withdrawals
```
abi87, line 153: this is the point of the test [here](https://github.com/berachain/beacon-kit/blob/41a76575535de5adb77811248bada0a2add084c0/mod/state-transition/pkg/core/state_processor_staking_test.go#L153), which is different from the base branch here: the list of withdrawals is empty instead of containing two withdrawals with zero amount.

mod/consensus-types/pkg/types/payload.go
```diff
@@ -128,6 +128,11 @@ func (p *ExecutionPayload) DefineSSZ(codec *ssz.Codec) {
 		constants.MaxBytesPerTx,
 	)
 	ssz.DefineSliceOfStaticObjectsContent(codec, &p.Withdrawals, 16)
+
+	// TODO: hack to avoid failure of EL checks on withdrawals
+	if p.Withdrawals == nil {
+		p.Withdrawals = make([]*engineprimitives.Withdrawal, 0)
+	}
```
abi87, line 135: post Shangai an EL explicitly check that Withdrawals are not nil (instead empty slices are fine).
Currently BeaconKit duly builds a block with Withdrawals set to empty slice (if not withdrawals are available) but as soon as the block is returned by CometBFT for verification, the SSZ decoding sets the empty slice to nil.
This code change solves the issue.

calbera (approved):
makes sense!

PR #2115, opened 2024-10-30
fix(state-transitions): verify deposits against contract

abi87:
Deposit list in execution payload body should be validated against the deposit contracts events to avoid creation of fake validators or censorship of real validators.
To do that this PR:
- Let state-transition processor access the deposit store, similarly to what block builder does
- Let state-transition processor compare the deposits list from payload with contracts one
- Adds logging to state-transition processor and deposit store
- Fixes `Eth1DepositIndex` to duly point to the latest stored deposit index

What this PR does not (yet)

- Fix deposit pruner (see error below, likely messing `(start, end)` indexes to prune with `(start, range)`)
- e2e broken if nethermind is in. Replacing nethermind from e2e tests with say reth, makes `-32001` disappear

Note:
- e2e broken with errors like
```
2024-11-05 19:03:38 2024-11-05T18:03:38Z WARN Failed to get deposits from block(s), retrying... service=deposit num_blocks=["0x5","0xb","0xc"]
2024-11-05 19:03:38 2024-11-05T18:03:38Z ERRR Failed to read deposits service=deposit error=Error -32001 (Receipt not available for From block 5.)
2024-11-05 19:03:38 2024-11-05T18:03:38Z ERRR Failed to read deposits service=deposit error=Error -32001 (Receipt not available for From block 11.)
2024-11-05 19:03:38 2024-11-05T18:03:38Z ERRR Failed to read deposits service=deposit error=Error -32001 (Receipt not available for From block 12.)
```
and several like
```
2024-11-05 19:15:34 2024-11-05T18:15:34Z ERRR ‼️ error pruning index ‼️ service=deposit-store-pruner error=DepositKVStore Prune start: 20, end: 16: range start greater than end
```

## Release Notes

- **New Features**
	- Added methods for validating genesis and non-genesis deposits, enhancing deposit management.
	- Introduced a new `Equals` method for comparing deposit instances.
	- Implemented a new `DepositStore` interface for managing deposit data.

- **Improvements**
	- Enhanced logging and error handling in deposit-related components.
	- Updated deposit retrieval logic to improve accuracy and efficiency.
	- Streamlined error messages for better clarity during deposit processing.
	- Improved logging capabilities in the deposit store and related components.
	- Replaced static constant with a dynamic reference for minting EVM tokens.

- **Configuration Changes**
	- Adjusted node configuration settings to optimize deposit checks.

- **Documentation**
	- Updated comments and documentation for improved clarity on recent changes.


mod/beacon/validator/block_builder.go
```diff
@@ -87,7 +87,8 @@ func (s *Service[
 	envelope, err := s.retrieveExecutionPayload(ctx, st, blk)
 	if err != nil {
 		return blk, sidecars, err
-	} else if envelope == nil {
```
abi87, line 90: not strictly necessary but OCD kicking in here 😅

mod/state-transition/pkg/core/state_processor_staking.go
```diff
@@ -38,26 +39,36 @@ func (sp *StateProcessor[
...
+	deposits := blk.GetBody().GetDeposits()
+	if len(stateDeposits) != len(deposits) {
+		return fmt.Errorf("deposits mismatched lengths, state: %d, payload: %d",
+			len(stateDeposits),
+			len(deposits),
+		)
+	}
+
+	for i, sd := range stateDeposits {
+		if !reflect.DeepEqual(sd, deposits[i]) {
```
calbera, line 65: i think we should write a `.Equal(` function on the Deposit struct as well. Not a fan of using DeepEqual especially when the types are interfaces

calbera (approved):

mod/state-transition/pkg/core/state_processor_staking.go
abi87: Not as simple as expected to define this given current error types.

mod/state-transition/pkg/core/state_processor_staking.go
```diff
@@ -38,26 +39,37 @@ func (sp *StateProcessor[
...
-	// of deposits.
-	deposits := blk.GetBody().GetDeposits()
-	index, err := st.GetEth1DepositIndex()
+	// Verify that outstanding deposits matches those listed by contract
+	depositIndex, err := st.GetEth1DepositIndex()
 	if err != nil {
 		return err
 	}
-	eth1Data, err := st.GetEth1Data()
+	depositIndex++
```
abi87, line 47: @calbera while adding unit tests spot this bug.
`st.GetEth1DepositIndex()` store the latest processed deposit, hence we need to move it up to process next.
This is similar to what we do in genesis

mod/node-core/pkg/components/state_processor.go
```diff
@@ -32,27 +33,32 @@ import (
 // StateProcessorInput is the input for the state processor for the depinject
 // framework.
 type StateProcessorInput[
+	LoggerT any,
```
abi87, line 36: This is much needed to have visibility over state transitions

mod/beacon/validator/block_builder.go
```diff
@@ -295,7 +295,7 @@ func (s *Service[
 
 	// Dequeue deposits from the state.
 	deposits, err := s.sb.DepositStore().GetDepositsByIndex(
-		depositIndex,
+		depositIndex+1, // pick expected next deposit index
```
abi87, line 298: this come from properly validating and fixing `GetEth1DepositIndex`. Now `GetEth1DepositIndex` returns the index of the latest stored deposit

mod/execution/pkg/deposit/sync.go
```diff
@@ -37,6 +37,15 @@ func (s *Service[
 	BeaconBlockT, _, _, _, _,
 ]) depositFetcher(ctx context.Context, event async.Event[BeaconBlockT]) {
 	blockNum := event.Data().GetBody().GetExecutionPayload().GetNumber()
+	if blockNum < s.eth1FollowDistance {
+		s.logger.Info(
+			"depositFetcher, nothing to fetch",
+			"block num", blockNum,
+			"eth1FollowDistance", s.eth1FollowDistance,
+		)
```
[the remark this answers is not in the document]
abi87: Not strictly necessary but wanted to rule out possibility of crazy stuff while debugging

mod/state-transition/pkg/core/state_processor_genesis.go
```diff
@@ -102,7 +97,18 @@ func (sp *StateProcessor[
 	}
 
 	for _, deposit := range deposits {
-		if err := sp.processDeposit(st, deposit); err != nil {
+		prevDepositIndex, err := st.GetEth1DepositIndex()
```
abi87, line 100: the reason deposit checks have been moved here is that for non-genesis blocks we already validate deposits by comparing them with deposits in store.
So we check genesis deposits are in the right sequence here (no other checks are needed, other than signature, which is done later on) and we simplify the `processDeposit` function

mod/state-transition/pkg/core/state_processor_staking.go
```diff
@@ -40,26 +40,46 @@ func (sp *StateProcessor[
...
-	index, err := st.GetEth1DepositIndex()
+	// Verify that outstanding deposits matches those listed by contract
+	depositIndex, err := st.GetEth1DepositIndex()
 	if err != nil {
 		return err
 	}
-	eth1Data, err := st.GetEth1Data()
+
+	stateDeposits, err := sp.ds.GetDepositsByIndex(
+		depositIndex+1, // pick expected next deposit index
```
abi87, line 50: coherent with block_builder implementation, we expect to process in this block the next deposit following latest one (whose index is returned by `GetEth1DepositIndex`)

mod/state-transition/pkg/core/state_processor_staking.go
```diff
@@ -87,24 +107,7 @@ func (sp *StateProcessor[
 	st BeaconStateT,
 	dep DepositT,
 ) error {
-	var nextDepositIndex uint64
```
abi87, line 90: for genesis this check has been moved to the client
for non-genesis block this check is redundant (caller already checks everything there is to check about deposits (but signature, done later))

mod/state-transition/pkg/core/types.go
```diff
@@ -129,15 +129,20 @@ type Context interface {
...
 ] interface {
+	// Equals returns true if the Deposit is equal to the other.
+	Equals(DepositT) bool
 	// GetAmount returns the amount of the deposit.
 	GetAmount() math.Gwei
 	// GetPubkey returns the public key of the validator.
 	GetPubkey() crypto.BLSPubkey
 	// GetWithdrawalCredentials returns the withdrawal credentials.
 	GetWithdrawalCredentials() WithdrawlCredentialsT
+	// GetIndex returns deposit index
```
abi87, line 144: needed to check sequence in genesis deposits

mod/storage/pkg/deposit/store.go
```diff
@@ -37,12 +39,18 @@ const KeyDepositPrefix = "deposit"
 // the deposit indexes are tracked outside of the kv store.
 type KVStore[DepositT Deposit[DepositT]] struct {
 	store sdkcollections.Map[uint64, DepositT]
-	mu    sync.RWMutex
+
+	// mu protects store for concurrent access
+	mu sync.RWMutex
+
+	// logger is used for logging information and errors.
+	logger log.Logger
```
abi87, line 47: not sure I want to keep logger here, or at least I should move `Info`s to `Debug`s. This is because this store is called by both block builder and verified, so we want to avoid double logging, or at least mark the client

mod/storage/pkg/deposit/store.go
```diff
@@ -115,8 +170,20 @@ func (kv *KVStore[DepositT]) Prune(start, end uint64) error {
 	for i := range end {
 		// This only errors if the key passed in cannot be encoded.
 		if err := kv.store.Remove(ctx, start+i); err != nil {
+			kv.logger.Error(
```
abi87, line 173: Note: the pruned is completelly wrong. It seems to return `(start, range)` instead of `(start, end)`.
Will fix in a different PR as soon as this fix is unlocked.
I guess the pruner issue was kinda silent since we don't really stop the node when it happens, just log an error

mod/state-transition/pkg/core/state_processor_staking.go
```diff
@@ -40,26 +40,47 @@
...
+	}
+
+	for i, sd := range stateDeposits {
+		if !sd.Equals(deposits[i]) {
+			sp.logger.Error(
+				ErrDepositMismatch.Error(),
+				"state deposit", sd,
+				"payload deposit", deposits[i],
+			)
+			return ErrDepositMismatch
```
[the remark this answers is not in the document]
abi87: not needed, index is reported inside the deposit

beacond/cmd/defaults.go
```diff
@@ -121,8 +121,8 @@ func DefaultComponents() []any {
 		],
 		components.ProvideStateProcessor[
 			*Logger, *BeaconBlock, *BeaconBlockBody, *BeaconBlockHeader,
-			*BeaconState, *BeaconStateMarshallable, *Deposit, *ExecutionPayload,
-			*ExecutionPayloadHeader, *KVStore,
+			*BeaconState, *BeaconStateMarshallable, *Deposit, *DepositStore,
+			*ExecutionPayload, *ExecutionPayloadHeader, *KVStore,
```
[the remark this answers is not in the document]
abi87: will do in a future PR. Make it work, make it right, make it fast, in this order. This does not work yet (nethermind fails)

mod/storage/pkg/deposit/store.go
```diff
@@ -61,52 +70,98 @@
...
 	kv.mu.RLock()
 	defer kv.mu.RUnlock()
-	deposits := []DepositT{}
-	for i := range numView {
-		deposit, err := kv.store.Get(context.TODO(), startIndex+i)
-		if errors.Is(err, sdkcollections.ErrNotFound) {
+	var (
+		deposits = []DepositT{}
+		endIdx   = startIndex + depRange
+	)
```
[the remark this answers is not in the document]
abi87: not possible, depRange comes from validated parameters

mod/state-transition/pkg/core/types.go
```diff
@@ -149,6 +152,15 @@ type Deposit[
...
 }
 
+// DepositStore defines the interface for deposit storage.
+type DepositStore[DepositT any] interface {
+	// GetDepositsByIndex returns `numView` expected deposits.
+	GetDepositsByIndex(
+		startIndex uint64,
+		numView uint64,
+	) ([]DepositT, error)
+}
```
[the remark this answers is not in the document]
abi87: not needed

mod/state-transition/pkg/core/state_processor.go
```diff
@@ -92,6 +92,8 @@ type StateProcessor[
 	executionEngine ExecutionEngine[
 		ExecutionPayloadT, ExecutionPayloadHeaderT, WithdrawalsT,
 	]
+	// ds allows checking payload deposits against the deposit contract
+	ds DepositStore[DepositT]
```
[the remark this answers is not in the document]
abi87: No need here

mod/state-transition/pkg/core/state_processor.go
```diff
@@ -148,6 +146,7 @@ func NewStateProcessor[
 	executionEngine ExecutionEngine[
 		ExecutionPayloadT, ExecutionPayloadHeaderT, WithdrawalsT,
 	],
+	ds DepositStore[DepositT],
```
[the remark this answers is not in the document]
abi87: no need

mod/state-transition/pkg/core/state_processor_staking.go
```diff
@@ -87,24 +108,7 @@ func (sp *StateProcessor[
...
-		nextDepositIndex = 0
-	default:
-		// Failed retrieving Eth1DepositIndex outside genesis is an error
-		return fmt.Errorf(
-			"failed retrieving eth1 deposit index outside of processing genesis: %w",
-			err,
-		)
-	}
-
-	if err := st.SetEth1DepositIndex(nextDepositIndex); err != nil {
```
abi87, line 107: btw, this was wrong conceptually. `Eth1DepositIndex` should hold the index of the latest deposit stored, not the expected next one, in order to adhere to Eth 2.0 specs.
Fixed in this PR (which was broken at some point due to `Eth1DepositIndex` being wrongly set)

testing/e2e/config/config.go
```diff
@@ -182,17 +182,17 @@ func defaultValidators() NodeSet {
...
 				KZGImpl:  "crate-crypto/go-kzg-4844",
 			},
 			{
 				ElType:   "geth",
-				Replicas: 2, //nolint:mnd // bet.
+				Replicas: 2, //nolint:mnd // we want two replicas here
 				KZGImpl:  "crate-crypto/go-kzg-4844",
 			},
 			{
 				ElType:   "reth",
```
abi87, line 194: debugging changes. It seems that:

- Dropping nethermind make e2e tests pass (nethermind seems to miss some blocks containing deposits)
- A network made of nethermind only nodes cannot even start

If confirmed, we may have to drop nethermind from e2e tests (simiarly to what we do for Besu) and re-evaluate later on if/how reactivate them

abi87: Filed an issue https://github.com/berachain/beacon-kit/issues/2177 to track Nethermind issue (this happens on main too)

mod/state-transition/pkg/core/state_processor_genesis.go
```diff
@@ -102,7 +97,18 @@ func (sp *StateProcessor[
...
+		if err != nil {
+			// currently this only happen on the first genesis deposit
+			if deposit.GetIndex() != 0 {
+				return nil, ErrDepositMismatch
+			}
```
calbera, line 105: should we ignore the err if its non-nil `&& deposit.GetIndex() == 0` ?

calbera: should still return err imo

mod/state-transition/pkg/core/state_processor_staking.go
calbera: Do we need to keep thsi for bartio as a special case?

mod/state-transition/pkg/core/state_processor_genesis.go
abi87: For the very first deposit, where `Eth1DepositIndex` has not been set yet, we have a ErrNotFound kind of error. This is fine (we haven't even set a single deposit) but it should only happen for the very first deposit (so the check on deposit index)
On the other hand, I agree that we should check the kind of error we receive, and we should behave like that only for ErrNotFound and not others but:
- ErrNotFound is a cosmosSDK error and we don't want that dependency here (discussed with Dev at some point)
- We do this in another place (see `applyDeposit` and how we do create validators
- I have an old PR that start fixing it (https://github.com/berachain/beacon-kit/pull/2065) and I am willing to refresh that and resubmit for review as soon as this in in

mod/state-transition/pkg/core/state_processor_staking.go
abi87: Done for all good measures. I would cause issues if syncing from scratch. Boonet should be fine

mod/config/pkg/spec/special_cases.go
```diff
@@ -28,3 +30,9 @@ const (
 	//nolint:lll // temporary.
 	BArtioValRoot = "0x9147586693b6e8faa837715c0f3071c2000045b54233901c2e7871b15872bc43"
 )
+
+const (
+	BoonetFork1Height uint64 = 69420
+
+	BoonetFork2Height uint64 = math.MaxUint64
```
abi87, line 37: @calbera I think we should temporarily list Boonet forks here, just to avoid spreading them across the codebase

mod/state-transition/pkg/core/state_processor_staking.go
```diff
@@ -41,26 +41,10 @@ func (sp *StateProcessor[
...
 	}
-	depositCount := min(
-		sp.cs.MaxDepositsPerBlock(),
-		eth1Data.GetDepositCount().Unwrap()-index,
-	)
-	_ = depositCount
-	// TODO: Update eth1data count and check this.
-	// if uint64(len(deposits)) != depositCount {
-	// 	return errors.New("deposit count mismatch")
-	// }
```
abi87, line 63: this was the validation done before

mod/state-transition/pkg/core/state/statedb.go
```diff
@@ -31,7 +31,7 @@ const (
 	// EVMMintingSlot is the slot at which we force a single withdrawal to
 	// mint EVMMintingAmount EVM tokens to EVMMintingAddress. No other
 	// withdrawals are inserted at this slot.
-	EVMMintingSlot uint64 = 69420
+	EVMMintingSlot = spec.BoonetFork1Height
```
abi87, line 34: not strlctly related to this PR, just trying to move all forks together

mod/state-transition/pkg/core/state_processor.go
```diff
@@ -219,6 +221,23 @@ func (sp *StateProcessor[
...
+				return nil, fmt.Errorf(
+					"failed retrieving deposit index at slot %d: %w",
+					slot, err,
+				)
+			}
+			fixedDepositIdx := idx - 1
+			if err = st.SetEth1DepositIndex(fixedDepositIdx); err != nil {
+				return nil, err
+			}
+		}
```
abi87, line 239: the Eth1DepositIndex was wrongly set to +1 the expected value. We fix this at fork.

mod/state-transition/pkg/core/state_processor.go
```diff
@@ -219,6 +221,23 @@ func (sp *StateProcessor[
...
+			slot == math.U64(spec.BoonetFork2Height) {
+			var idx uint64
+			idx, err = st.GetEth1DepositIndex()
+			if err != nil {
+				return nil, fmt.Errorf(
+					"failed retrieving deposit index at slot %d: %w",
+					slot, err,
+				)
+			}
+			fixedDepositIdx := idx - 1
```
abi87, line 235: no need to check whether idx>1 since we must have some validators (at least genesis ones)

mod/config/pkg/spec/special_cases.go
```diff
@@ -28,3 +30,9 @@ const (
 	//nolint:lll // temporary.
 	BArtioValRoot = "0x9147586693b6e8faa837715c0f3071c2000045b54233901c2e7871b15872bc43"
 )
+
+const (
```
calbera, line 34:
```suggestion
// Planned hard-fork upgrades on boonet.
const (
```
or something similar

mod/consensus-types/pkg/types/deposit.go
```diff
@@ -186,6 +186,14 @@ func (d *Deposit) GetTree() (*fastssz.Node, error) {
 /* -------------------------------------------------------------------------- */
 /*                             Getters and Setters                            */
 /* -------------------------------------------------------------------------- */
+// Equals returns true if the Deposit is equal to the other.
+func (d *Deposit) Equals(rhs *Deposit) bool {
+	return d.Pubkey == rhs.Pubkey &&
+		d.Credentials == rhs.Credentials &&
+		d.Amount == rhs.Amount &&
+		d.Signature == rhs.Signature &&
+		d.Index == rhs.Index
```
calbera, line 195: for the fields that are special types, should we use `bytes.Equal` or the type's `Equals` method?

mod/execution/pkg/deposit/sync.go
```diff
@@ -37,6 +37,15 @@ func (s *Service[
 	BeaconBlockT, _, _, _, _,
 ]) depositFetcher(ctx context.Context, event async.Event[BeaconBlockT]) {
 	blockNum := event.Data().GetBody().GetExecutionPayload().GetNumber()
+	if blockNum < s.eth1FollowDistance {
```
calbera, line 40: correct me If I'm wrong, but just confirming, this case is only triggered when blockNum is 0?

abi87: as the values are set right now yes. However that is not really hit, since we don't emit a finalized block event on genesis

mod/state-transition/pkg/core/deposits_validation.go
```diff
@@ -0,0 +1,140 @@
...
+			// genesis first deposit
+			return ErrDepositMismatch
+		}
+		if len(deposits) == 0 {
+			// there should be at least a validator in genesis
+			return ErrDepositsLengthMismatch
+		}
+		for i, deposit := range deposits {
+			// deposits indexes should be contiguous
+			if deposit.GetIndex() != math.U64(i) {
```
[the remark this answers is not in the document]
abi87: it comes from a loop, all good

mod/execution/pkg/deposit/sync.go
calbera: eth1FollowDistance 
```suggestion
	if blockNum < s.eth1FollowDistance + 1 {
```

Will never need to query the logs from block height 0. So the first block to query for should be 1.

mod/consensus-types/pkg/types/deposit.go
abi87: I would say their equals, but we should be good now

mod/state-transition/pkg/core/deposits_validation.go
```diff
@@ -0,0 +1,140 @@
...
+		return nil
+
+	case sp.cs.DepositEth1ChainID() == spec.BoonetEth1ChainID:
+		// Boonet inherited the bug from Bartio and it may have added some
+		// validators before we activate the fork. So we skip validation
+		// before fork activation
+		return nil
+
+	default:
+		if _, err := st.GetEth1DepositIndex(); err == nil {
```
calbera, line 50: what error is exactly returned here? Should we explicity check that the error contains "not found" ?

abi87: Can't do now without importing cosmos sdk module here, which we really don't want. I wrote a comment about this but maybe got lost in updates. Let me find it

abi87: For the very first deposit, where `Eth1DepositIndex` has not been set yet, we have a ErrNotFound kind of error. This is fine (we haven't even set a single deposit) but it should only happen for the very first deposit (so the check on deposit index)
On the other hand, I agree that we should check the kind of error we receive, and we should behave like that only for ErrNotFound and not others but:
- ErrNotFound is a cosmosSDK error and we don't want that dependency here (discussed with Dev at some point)
- We do this in another place (see `applyDeposit` and how we do create validators
- I have an old PR that start fixing it (https://github.com/berachain/beacon-kit/pull/2065) and I am willing to refresh that and resubmit for review as soon as this in in

mod/state-transition/pkg/core/state_processor.go
calbera: makes sense, can you put this as a comment on this line?

mod/state-transition/pkg/core/state_processor.go
calbera: can you mention this in a comment under line 226?

mod/storage/pkg/deposit/store.go
calbera: Do we have GH issue for this? Not sure, maybe its related to https://github.com/berachain/beacon-kit/issues/2134

mod/state-transition/pkg/core/types.go
calbera: also applied for use in non-genesis. Order of deposits processing, by index, should always be maintained.

mod/storage/pkg/deposit/store.go
```diff
@@ -61,52 +70,98 @@ func NewStore[DepositT Deposit[DepositT]](
...
 // Prune removes the [start, end) deposits from the store.
 func (kv *KVStore[DepositT]) Prune(start, end uint64) error {
+	kv.logger.Info(
+		"Prune request",
+		"start", start,
+		"end", end,
+	)
 	if start > end {
-		return pruner.ErrInvalidRange
+		return fmt.Errorf(
```
calbera, line 161: generally all use of `fmt.Errorf` can be nicely replaced by `errors.Wrapf`

mod/state-transition/pkg/core/deposits_validation.go
```diff
@@ -0,0 +1,158 @@
...
+		if len(localDeposits) != len(deposits) {
+			return errors.Wrapf(
+				ErrDepositsLengthMismatch,
+				"local: %d, payload: %d", len(localDeposits), len(deposits),
+			)
+		}
+
+		for i, sd := range localDeposits {
+			// Deposit indices should be contiguous.
+			if sd.GetIndex().Unwrap() != expectedStartIdx+uint64(i) {
```
calbera, line 139: order should always be maintained. From spec:

```python
def process_deposit(state: BeaconState, deposit: Deposit) -> None:
    ...

    # Deposits must be processed in order
    state.eth1_deposit_index += 1

    apply_deposit(
        state=state,
        pubkey=deposit.data.pubkey,
        withdrawal_credentials=deposit.data.withdrawal_credentials,
        amount=deposit.data.amount,
        signature=deposit.data.signature,
    )
```

calbera (approved):

build/scripts/codegen.mk
```diff
@@ -12,7 +12,7 @@ generate: ## generate all the code
 		(cd $$module && \
 			GETH_PKG_INCLUDE=$(GETH_PKG_INCLUDE) go generate ./...) || exit 1; \
 	done
-	@go run github.com/vektra/mockery/v2@latest
```
abi87, line 15: anchored mockery to avoid frequent updates. We can revisit say once per month

mod/state-transition/pkg/core/deposits_validation.go
abi87: These checks may be redundant, since these data come from our EVM state and I think they are guaranteed to be in-order. Keeping this for the time being, we will revisit all once merkle root checks are in

mod/state-transition/pkg/core/deposits_validation.go
```diff
@@ -0,0 +1,158 @@
...
+					sd.GetIndex().Unwrap(), expectedStartIdx+uint64(i),
+				)
+			}
+
+			if !sd.Equals(deposits[i]) {
+				return errors.Wrapf(
+					ErrDepositMismatch,
+					"local deposit: %d, payload deposit: %d",
+					sd, deposits[i],
+				)
```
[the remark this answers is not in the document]
calbera: @abi87 we should probably accept this suggestion

Issue #2122, opened 2024-11-01
Create-validator cli is broken since alpha.5


chuck-bear:
The create-validator CLI is broken since alpha.5, last time it worked it was with alpha.4.

How to reproduce it ?

Run that command line with alpha.8 binary:
```
beacond deposit create-validator $VAL_WITHDRAW_CREDENTIAL $VAL_DEPOSIT_GWEI_AMOUNT $GENESIS_FORK_VERSION $GENESIS_VALIDATORS_ROOT --private-key $YOUR_ETH_WALLET_PRIVATE_KEY  --home $YOUR_BEACON_HOME_DIR
```
It will return `open : no such file or directory`

And run it with alpha.4 binary.

Let me know if you need more details.

chuck-bear:
Thanks @dezzeus this looks very familiar to https://github.com/berachain/beacon-kit/issues/2123

abi87:
@chuck-bear we do have a proposed solution by @dezzeus on https://github.com/berachain/beacon-kit/pull/2128. Can you try it out?

chuck-bear:
@dezzeus @abi87 just tested, it's working. Thanks!

PR #2128, opened 2024-11-05
fix: let the create-validator command retrieve the loaded configuration.

dezzeus:
Upon CLI bootstrapping, a `PersistentPreRunE` handler is added to the command in order to, among other things, load the configuration in a viper instance and store it into the command's context with a `ViperContextKey` key.

The `create-validator` command was using the wrong, but homonymous, `ViperContextKey` for retrieving the viper configuration from the command context.

As a result, the command implementation was not able to find some configuration values which were silently used as empty strings that, eventually, would halt the execution with unexpected errors.

This commit replaces the source package of the `ViperContextKey` key from `github.com/cosmos/cosmos-sdk/client`
to `github.com/berachain/beacon-kit/mod/cli/pkg/context`

This should also fix the issue #2122 .

## Summary by CodeRabbit

- **New Features**
	- Improved command for validator deposits with enhanced context management.
- **Bug Fixes**
	- Enhanced error handling for input conversions during deposit message creation.

### DIFF
--- a/mod/cli/pkg/commands/deposit/create.go
+++ b/mod/cli/pkg/commands/deposit/create.go
@@ -24,6 +24,7 @@ import (
 	"os"
 
 	"cosmossdk.io/log"
+	clicontext "github.com/berachain/beacon-kit/mod/cli/pkg/context"
 	"github.com/berachain/beacon-kit/mod/cli/pkg/utils/parser"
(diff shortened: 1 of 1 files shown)


chuck-bear (approved):
LGTM

abi87 (approved):

PR #2132, opened 2024-11-06
fix(make): permission error with slither

gummybera:
Makes volume writable for slither to generate artefacts. 

## Summary by CodeRabbit

- **Chores**
	- Updated Docker volume mount option for the Slither tool to change the contracts directory path, improving compatibility with the Docker container environment.


abi87:
Slither failing seems a fluke that recover by itself.
Closing this now, but keeping this in history to revisit it the issue happens again.
Appreciate @gummybera for the effort in investigating this

PR #2143, opened 2024-11-09
feature(state-processor): send only validators set diffs to consensus

abi87:
Whenever we turn the epoch, we must inform consensus of the changes in the validator set composition.
It's best to send only diffs of the validator set wrt previous epoch, rather than sending the whole new validator set.
This is what this PR does.

## Summary by CodeRabbit

- **Documentation**
	- Updated the documentation for validator handling to clarify conditions for returning validators to the consensus engine.

- **Bug Fixes**
	- Improved clarity of error messages related to proposer mismatches.

- **New Features**
	- Introduced a mechanism to track previous epoch validators, enhancing the efficiency of validator updates.
	- Added a new method for processing validator set updates, optimizing the consensus mechanism.

- **Refactor**
	- Modified methods to improve the handling of validator updates and streamline processing logic.

- **Tests**
	- Simplified assertions in test cases related to validator transitions, enhancing clarity and focus.


mod/state-transition/pkg/core/state_processor.go
```diff
@@ -94,6 +94,11 @@ type StateProcessor[
 	// processingGenesis allows initializing correctly
 	// eth1 deposit index upon genesis
 	processingGenesis bool
+
+	// prevEpochValidators tracks the set of validators active during
+	// previous epoch. This is useful at the turn of the epoch to send to
+	// consensus only the diffs rather than the full updated validator set.
+	prevEpochValidators []*transition.ValidatorUpdate
```
abi87, line 101: note that we only keep active validators here, i.e. those whose effective balance is larger than ejection one.
I do not expect this list to be beefy, although we should probably track its size and decide whether this is the best way to calculate diffs.

A less memory consuming approach is to track chances wherever we do them (e.g.  when updating effective balances, or creating validators or removing it). I prefer current approach is it is more easily verifiable for correctness for me, but we should measure how much of a performance improvement these two alternatives bring

mod/state-transition/pkg/core/state_processor_genesis.go
```diff
@@ -166,5 +166,5 @@ func (sp *StateProcessor[
 		return nil, err
 	}
 
-	return sp.processSyncCommitteeUpdates(st)
```
abi87, line 169: `processSyncCommitteeUpdates` does not really make sense to me.
It seems to me we're trying to squeeze a pure BeaconKit operation (updating CometBFT val set) into an Eth 2.0 one

mod/state-transition/pkg/core/state_processor_staking_test.go
```diff
@@ -214,15 +213,7 @@ func TestTransitionUpdateValidator(t *testing.T) {
...
-	require.Len(t, newEpochVals, len(genDeposits)) // just topped up one validator
-
-	// Assuming genesis order is preserved here which is not necessary
-	// TODO: remove this assumption
-
-	// all genesis validators other than the last are unchanged
-	for i := range len(genDeposits) - 1 {
-		require.Equal(t, genVals[i], newEpochVals[i], fmt.Sprintf("idx: %d", i))
-	}
+	require.Len(t, newEpochVals, 1) // just topped up one validator
```
abi87, line 216: only send diffs

mod/state-transition/pkg/core/state_processor_staking_test.go
```diff
@@ -400,15 +391,7 @@ func TestTransitionCreateValidator(t *testing.T) {
...
-	require.Len(t, newEpochVals, len(genDeposits)+1)
-
-	// Assuming genesis order is preserved here which is not necessary
-	// TODO: remove this assumption
-
-	// all genesis validators are unchanged
-	for i := range len(genDeposits) {
-		require.Equal(t, genVals[i], newEpochVals[i], fmt.Sprintf("idx: %d", i))
-	}
+	require.Len(t, newEpochVals, 1) // just added 1 validator
```
abi87, line 394: only send diff.

mod/state-transition/pkg/core/helpers_test.go
```diff
@@ -219,3 +220,18 @@ func buildNextBlock(
...
+	)
+
+	dummyExecutionPayload = &types.ExecutionPayload{
+		Timestamp:     0,
+		ExtraData:     []byte("testing"),
+		Transactions:  [][]byte{},
+		Withdrawals:   []*engineprimitives.Withdrawal{},
+		BaseFeePerGas: math.NewU256(0),
+	}
+)
```
abi87, line 237: minor UT consolidation

mod/state-transition/pkg/core/state_processor_validators.go
```diff
@@ -58,4 +60,45 @@ func (sp *StateProcessor[
...
+
+	// prevValsSet now contains all evicted validators (and only those)
+	for pkBytes := range prevValsSet {
+		//#nosec:G703 // bytes comes from a pk
+		pk, _ := bytes.ToBytes48([]byte(pkBytes))
+		res = append(res, &transition.ValidatorUpdate{
+			Pubkey:           pk,
+			EffectiveBalance: 0, // signal val eviction to consensus
+		})
+	}
```
abi87, line 99: we do not evict validators yet, so this should not really happen. Still added for completeness

calbera (approved):
logically makes sense, agree with trying to optimize this in future

nidhi-singh02 (approved):

mod/state-transition/pkg/core/state_processor_validators.go
abi87: in fact we may evict them as soon as we introduce a cap on the validators set

fridrik01 (approved):
LGTM

mod/state-transition/pkg/core/state_processor_validators.go
```diff
@@ -0,0 +1,134 @@
...
+	res, err := sp.validatorSetsDiffs(prevEpochVals, activeVals)
+	if err != nil {
+		return nil, err
+	}
+
+	// clear up sets we won't lookup to anymore
+	sp.valSetByEpoch[currEpoch] = activeVals
+	if prevEpoch >= 1 {
+		delete(sp.valSetByEpoch, prevEpoch-1)
+	}
```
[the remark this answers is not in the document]
abi87: Epochs are not skipped

mod/state-transition/pkg/core/state_processor_validators.go
```diff
@@ -58,4 +60,45 @@ func (sp *StateProcessor[
...
+		}
+
+		// consume pre-existing validators
+		delete(prevValsSet, key)
+	}
+
+	// prevValsSet now contains all evicted validators (and only those)
+	for pkBytes := range prevValsSet {
+		//#nosec:G703 // bytes comes from a pk
+		pk, _ := bytes.ToBytes48([]byte(pkBytes))
```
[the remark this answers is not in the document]
abi87: no need, it's fine

PR #2148, opened 2024-11-12
fix(state-transition): Withdrawals backward compatibility with Bartio

abi87:
This PR:

- Makes changes introduced in https://github.com/berachain/beacon-kit/pull/2110 backward compatible
- Hardens Equals methods to handle nil pointer cases (uncovered crash while testing)


mod/consensus-types/pkg/types/header.go
```diff
@@ -176,11 +176,19 @@ func (b *BeaconBlockHeader) GetTree() (*fastssz.Node, error) {
 
 // Equals returns true if the Withdrawal is equal to the other.
 func (b *BeaconBlockHeader) Equals(rhs *BeaconBlockHeader) bool {
-	return b.Slot == rhs.Slot &&
-		b.ProposerIndex == rhs.ProposerIndex &&
-		b.ParentBlockRoot == rhs.ParentBlockRoot &&
-		b.StateRoot == rhs.StateRoot &&
-		b.BodyRoot == rhs.BodyRoot
+	switch {
```
abi87, line 179: not strictly related to the fix, but helpful to harden all the custom Equals functions we wrote

mod/state-transition/pkg/core/state/statedb.go
```diff
@@ -266,6 +269,18 @@ func (s *StateDB[
 
 			// Increment the withdrawal index to process the next withdrawal.
 			withdrawalIndex++
+		} else if s.cs.DepositEth1ChainID() == spec.BartioChainID {
```
abi87, line 272: add empty withdrawals for backward compatibility in Bartio.

calbera (approved):
makes sense

PR #2153, opened 2024-11-13
chore(all): update SSZ dependencies

abi87:
We need to clean up one of the ssz dependencies. This PR achieve this as follow:

- Replaces github.com/itsdevbear/ssz (no longer public) with https://github.com/berachain/karalabe-ssz
- The original dependency use to allow `[]byte` among  extension, which is broken. Dropped that but kept blobs extension. See https://github.com/berachain/karalabe-ssz/releases/tag/v0.2.0-alpha.9
- Carry out code changes related to the updated version of SSZ
- Fix dependency conflict related to update deps

## Summary by CodeRabbit

- **New Features**
	- Simplified method signature for `ToHeader` in the `ExecutionPayload` interface, removing unnecessary parameters.
	- Enhanced size calculation methods across various structs to utilize `ssz.Sizer`, improving flexibility in serialization.
  
- **Bug Fixes**
	- Improved error handling in serialization processes to prevent nil pointer dereferences.

- **Dependency Updates**
	- Updated multiple dependencies to their latest versions, ensuring compatibility and stability across modules. 

- **Documentation**
	- Updated comments and documentation to reflect changes in method signatures and functionality.

PR #2154, opened 2024-11-14
chore(all) - updated fastssz deps

abi87:
Follow up on https://github.com/berachain/beacon-kit/pull/2153


calbera (approved):
We are updating [this diff](https://github.com/itsdevbear/fastssz/compare/main...ferranbt%3Afastssz%3Amain) on the fastssz dep --> LGTM

PR #2158, opened 2024-11-15
feat(withdrawals): Fixed withdrawal for EVM inflation per block

calbera:
**Summary**
- Every block we add a fixed EVM Inflation withdrawal (fixed amount, fixed address, and fixed at position 0 in the list of withdrawals)
- Not processed for use by the CL StateProcessor
- Processed by EL clients to add EVM balance of the fixed amount to the fixed address
- Enforce chain spec `MaxWithdrawalsPerBlock` is at least 2 (to allow 1 for the inflation withdrawal and 1 for validator withdrawal)
- Refactor e2e testing, chain_specs, default values

**Side Effect**:

MaxWithdrawalsPerPayload is 16 according the ETH 2.0 spec. Choosing to enforce the first withdrawal as EVM inflation withdrawal implies only 15 validator withdrawals per block are allowed at most.

**Tests**
- [x] unit test statedb.ExpectedWithdrawals (done through state processor tests for now)
- [x] unit test stateProcessor.processWIthdrawals
- [x] E2E test of EVM balance increasing every block

## Release Notes

- **New Features**
  - Introduced methods for EVM inflation address and per block minting in the chain specification.
  - Added error handling for various chain specification retrieval functions.
  - Enhanced testing capabilities with new setup functions and tests for EVM inflation.

- **Bug Fixes**
  - Improved error reporting for withdrawal discrepancies and validation failures.

- **Chores**
  - Updated dependencies and removed obsolete constants related to deposits.
  - Restructured end-to-end test configurations for better clarity and usability. 

These updates enhance the functionality, reliability, and testability of the application, particularly in relation to EVM inflation and withdrawal processes.


mod/state-transition/pkg/core/state_processor_staking.go
```diff
@@ -289,24 +310,23 @@ func (sp *StateProcessor[
 		return err
 	}
 
-	// Update the next validator index to start the next withdrawal sweep
-	//#nosec:G701 // won't overflow in practice.
-	if numWithdrawals == int(sp.cs.MaxWithdrawalsPerPayload()) {
-		// Next sweep starts after the latest withdrawal's validator index
-		nextValidatorIndex =
-			(expectedWithdrawals[len(expectedWithdrawals)-1].GetIndex() + 1) %
-				math.U64(totalValidators)
```
calbera, line 298: fixing this according to [spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#new-process_withdrawals) (`.GetIndex()` is replaced by `.GetValidatorIndex()`) 

```python
if len(expected_withdrawals) == MAX_WITHDRAWALS_PER_PAYLOAD:
        # Next sweep starts after the latest withdrawal's validator index
        next_validator_index = ValidatorIndex((expected_withdrawals[-1].validator_index + 1) % len(state.validators))
```

mod/state-transition/pkg/core/state/statedb.go
```diff
@@ -283,21 +293,30 @@ func (s *StateDB[
...
-		if len(withdrawals) == int(s.cs.MaxWithdrawalsPerPayload()) {
-			break
-		}
-
 		// Increment the validator index to process the next validator.
-		validatorIndex = (validatorIndex + 1) % math.ValidatorIndex(
-			totalValidators,
-		)
+		validatorIndex++
+		validatorIndex %= math.ValidatorIndex(totalValidators)
```
abi87, line 298: q: I think the original was closer to specs? I am fine with this kind of changes, just what is the preference with BeaconKit (more or less concise)?

mod/state-transition/pkg/core/state/statedb.go
```diff
@@ -283,21 +293,30 @@ func (s *StateDB[
...
+// EVMInflationWithdrawal returns the withdrawal used for EVM balance inflation.
+//
+// NOTE: The withdrawal index and validator index are both set to 0 as they are
+// not used during processing.
+func (s *StateDB[
+	_, _, _, _, _, _, _, _, WithdrawalT, _,
+]) EVMInflationWithdrawal() WithdrawalT {
+	var withdrawal WithdrawalT
+	return withdrawal.New(
+		0,
```
abi87, line 313: the one problem I have with this choice is that 0 is a valid validator. Granted that validator index should not make a difference from the EL perspective, I still wonder if it's better, even just estetically, to pick the block proposer as validator to "withdraw minting to"

mod/state-transition/pkg/core/state_processor_genesis_test.go
```diff
@@ -24,68 +24,45 @@ import (
...
-		mocksSigner,
-		dummyProposerAddressVerifier,
-	)
-
-	// create test inputs
-	kvStore, err := initStore()
-	require.NoError(t, err)
-	beaconState := new(TestBeaconStateT).NewFromDB(kvStore, cs)
+	cs := setupChain(t, components.BetnetChainSpecType)
+	sp, st, _ := setupState(t, cs)
```
abi87, line 38: niiiiiice

mod/state-transition/pkg/core/state_processor_genesis_test.go
```diff
@@ -24,68 +24,45 @@ import (
...
 
 	var (
 		deposits = []*types.Deposit{
 			{
-				Pubkey: [48]byte{0x01},
+				Pubkey: [48]byte{0x00},
```
abi87, line 43: I avoided the 0x00 PK because I believed it was a special, invalid value, but state processor should not really do much with PK except comparing them

mod/state-transition/pkg/core/state_processor_genesis_test.go
```diff
@@ -160,53 +128,35 @@ func checkValidatorNonBartio(
...
-		execEngine,
-		mocksSigner,
-		dummyProposerAddressVerifier,
-	)
-
-	// create test inputs
-	kvStore, err := initStore()
-	require.NoError(t, err)
-	beaconState := new(TestBeaconStateT).NewFromDB(kvStore, cs)
+	cs := setupChain(t, "testnet")
```
abi87, line 131: nit: shouldn't we have a variable for "testnet", like `components.BetnetChainSpecType` above?

mod/state-transition/pkg/core/state_processor_genesis_test.go
```diff
@@ -320,9 +261,3 @@ func commonChecksValidators(
 		require.Equal(t, math.Gwei(0), val.EffectiveBalance)
 	}
 }
-
-// in genesis UTs we don't need to verify proposer address
-// (no one proposes genesis), hence the dummy implementation.
-func dummyProposerAddressVerifier(bytes.B48) ([]byte, error) {
-	return nil, nil
-}
```
abi87, line 328: genesis should not really have a proposer but I guess we don't really care here, so I am in favour of the simplification

mod/state-transition/pkg/core/state_processor_staking.go
```diff
@@ -255,18 +256,37 @@ func (sp *StateProcessor[
 		)
 	}
 
+	// Ensure the EVM inflation withdrawal is the first withdrawal.
+	if numWithdrawals == 0 {
+		return ErrNoWithdrawals
+	}
```
abi87, line 262: is this check needed? We build ExpectedWithdrawals with at least 1 withdrawal and we check the length above.
Consider dropping it

mod/state-transition/pkg/core/state_processor_staking.go
```diff
@@ -255,18 +256,37 @@ func (sp *StateProcessor[
...
+	if !expectedWithdrawals[0].GetAddress().Equals(
+		sp.cs.EVMInflationAddress(),
+	) ||
+		expectedWithdrawals[0].GetAmount() != math.Gwei(
+			sp.cs.EVMInflationPerBlock(),
+		) {
+		return ErrFirstWithdrawalNotEVMInflation
+	}
```
abi87, line 270: nit:  why don't we just use equals here? Looks like we build expected payload setting validator index to zero, but we do let withdrawals in block to set the validator differently. Granted we don't really care, I'd rather Expected and actual withdrawal to fully match.

mod/state-transition/pkg/core/state_processor_staking.go
```diff
@@ -255,18 +256,37 @@ func (sp *StateProcessor[
...
+
 	// Compare and process each withdrawal.
 	for i, wd := range expectedWithdrawals {
 		// Ensure the withdrawals match the local state.
 		if !wd.Equals(payloadWithdrawals[i]) {
 			return errors.Wrapf(
-				ErrNumWithdrawalsMismatch,
-				"withdrawals do not match expected %s, got %s",
-				spew.Sdump(wd), spew.Sdump(payloadWithdrawals[i]),
+				ErrWithdrawalMismatch,
```
abi87, line 277: is the range above correct? I would expect the indexes to be shifted all by one. Something like:
`for i:=1; i <expectedWithdrawals; i++`
In this way we also simplify the loop, since we don't need to test for i==0

abi87 (commented):
some nits, questions and comments, have not finished the review.

mod/chain-spec/pkg/chain/chain_spec.go
```diff
@@ -207,14 +218,27 @@ func NewChainSpec[
...
+// validate ensures that the chain spec is valid, returning error if it is not.
+func (c *chainSpec[
+	DomainTypeT, EpochT, ExecutionAddressT, SlotT, CometBFTConfigT,
+]) validate() error {
+	if c.MaxWithdrawalsPerPayload() <= 1 {
+		return ErrInsufficientMaxWithdrawalsPerPayload
+	}
+
+	// TODO: Add more validation rules here.
+	return nil
```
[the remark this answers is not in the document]
calbera: I think both of these can be the 0 values, no other validation needed. (InflationAddress can be the zero address, Inflation per block can be 0)

mod/config/pkg/spec/testnet.go
```diff
@@ -122,6 +122,12 @@ func BaseSpec() chain.SpecData[
...
 		BytesPerBlob:                     131072,
 		KZGCommitmentInclusionProofDepth: 17,
-		CometValues:                      cmtConsensusParams,
+		// Comet values.
+		CometValues: cmtConsensusParams,
+		// Berachain values.
+		EVMInflationAddress: common.NewExecutionAddressFromHex(
+			"0x6942069420694206942069420694206942069420",
+		),
+		EVMInflationPerBlock: 10e9,
```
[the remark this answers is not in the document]
calbera: These are completely arbitrary values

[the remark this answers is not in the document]
calbera: It's really chain dependent

mod/state-transition/pkg/core/state/statedb.go
```diff
@@ -226,6 +233,11 @@ func (s *StateDB[
 
 	// Iterate through indices to find the next validators to withdraw.
 	for range bound {
+		// Cap the number of withdrawals to the maximum allowed per payload.
+		if uint64(len(withdrawals)) == s.cs.MaxWithdrawalsPerPayload() {
```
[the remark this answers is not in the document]
calbera: yeah agreed I can remove this change since I changed the bound to at least 2

mod/state-transition/pkg/core/state/statedb.go
calbera: yeah originally was exactly from spec I can keep it that way, minimize the diff for now

mod/state-transition/pkg/core/state/statedb.go
calbera: I chose 0 to explicitly state its not for use downstream. We can create a uint64 constant `EVMInflationWithdrawalIndex` and `EVMInflationWithdrawalValidatorIndex` if something like that is more clear

mod/state-transition/pkg/core/state_processor_genesis_test.go
calbera: yea just keeping it consistent with validator index for now

mod/state-transition/pkg/core/state_processor_staking.go
calbera: would agree that the validation should just be for all, as we don't need special check for i==0.

mod/state-transition/pkg/core/state_processor_genesis_test.go
calbera: Right now anything thats not from one of these:
```golang
	DevnetChainSpecType = "devnet"
	BetnetChainSpecType = "betnet"
	BoonetChainSpecType = "boonet"
```
will default to the `TestnetChainSpec()`. I'm personally okay with this for now. We could maybe rename this to `LocalnetChainSpec()` if you like.

mod/config/pkg/spec/testnet.go
```diff
@@ -122,6 +122,17 @@ func BaseSpec() chain.SpecData[
...
 		KZGCommitmentInclusionProofDepth: 17,
-		CometValues:                      cmtConsensusParams,
+		// Comet values.
+		CometValues: cmtConsensusParams,
+		// Berachain values.
+		//
+		// Expected address of the BGT contract on the bArtio testnet.
+		EVMInflationAddress: common.NewExecutionAddressFromHex(
+			"0x289274787bAF083C15A45a174b7a8e44F0720660",
+		),
```
fridrik01, line 132: I don't [see this contract](https://bartio.beratrail.io/address/0x289274787bAF083C15A45a174b7a8e44F0720660) on bArtio, but its [listed in the cArtio](https://github.com/berachain/subgraphs-monorepo/blob/11530b554224ea6d3dd2448bb34369553d093e12/config/cartio.json#L27) config. Are we expecting this contract to be deployed on bArtio?

mod/state-transition/pkg/core/state/statedb.go
```diff
@@ -251,7 +233,7 @@ func (s *StateDB[
 
 	// Iterate through indices to find the next validators to withdraw.
 	for range bound {
-		validator, err = s.ValidatorByIndex(validatorIndex)
+		validator, err = s.ValidatorByIndex(validatorIndex) // 0
```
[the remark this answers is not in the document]
fridrik01: This seems like a false positive, according to [go spec](https://go.dev/ref/spec#For_statements) this syntax should be valid, though I haven't often seen it.

mod/state-transition/pkg/core/state_processor_staking.go
```diff
@@ -308,24 +299,26 @@ func (sp *StateProcessor[
...
 			return err
 		}
 		nextValidatorIndex += math.ValidatorIndex(
 			sp.cs.MaxValidatorsPerWithdrawalsSweep())
 		nextValidatorIndex %= math.ValidatorIndex(totalValidators)
+		return st.SetNextWithdrawalValidatorIndex(nextValidatorIndex)
 	}
 
-	return st.SetNextWithdrawalValidatorIndex(nextValidatorIndex)
+	return nil
```
fridrik01, line 323: If `numWithdrawals == 0`, should that not be an error as we expected to have always the EVM inflation withdrawal first?

fridrik01 (commented):
Lacking some context still, but left a few minor comments

mod/config/pkg/spec/testnet.go
calbera: Yes, making this more explicit. BaseSpec should be defaults, which will be used by e2e. For boonet, will set this specifically to BGT address `0x289274787bAF083C15A45a174b7a8e44F0720660`

mod/state-transition/pkg/core/state_processor_staking.go
calbera: Agree that we should error if numWithdrawals == 0.

mod/state-transition/pkg/core/state_processor.go
```diff
@@ -301,7 +301,7 @@ func (sp *StateProcessor[
 		return err
 	}
 
-	if err := sp.processWithdrawals(st, blk); err != nil {
+	if err := sp.processWithdrawals(st, blk.GetBody()); err != nil {
```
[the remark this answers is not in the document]
calbera: fixed

mod/state-transition/pkg/core/state_processor_staking.go
```diff
@@ -258,34 +258,24 @@ func (sp *StateProcessor[
...
-		if !wd.Equals(payloadWithdrawals[0]) {
-			return fmt.Errorf(
-				"minting withdrawal does not match expected %s, got %s",
-				spew.Sdump(wd), spew.Sdump(payloadWithdrawals[0]),
-			)
-		}
-
-		// No processing needed.
-		return nil
-	}
```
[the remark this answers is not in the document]
calbera: done

mod/chain-spec/pkg/chain/chain_spec.go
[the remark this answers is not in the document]
nidhi-singh02: adding a comment in code would really help here.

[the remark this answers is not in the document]
calbera: done

mod/chain-spec/pkg/chain/chain_spec.go
```diff
@@ -207,14 +218,29 @@ func NewChainSpec[
...
+func (c *chainSpec[
+	DomainTypeT, EpochT, ExecutionAddressT, SlotT, CometBFTConfigT,
+]) validate() error {
+	if c.MaxWithdrawalsPerPayload() <= 1 {
+		return ErrInsufficientMaxWithdrawalsPerPayload
+	}
+
+	// EVM Inflation values can be zero or non-zero, no validation needed.
+
+	// TODO: Add more validation rules here.
```
abi87, line 242: yes, good we have the scaffolding to do this rn.
Note: these checks may be fork dependent too

mod/config/pkg/spec/special_cases.go
```diff
@@ -31,8 +29,10 @@ const (
...
 )
 
-const ( // Planned hard-fork upgrades on boonet.
+// Planned hard-fork upgrades on boonet.
+const (
 	BoonetFork1Height uint64 = 69420
 
-	BoonetFork2Height uint64 = math.MaxUint64
+	// TODO: modify this to be the actual fork height. Avoid overflow.
+	BoonetFork2Height uint64 = 99999999999999
```
abi87, line 37: I'll check the rest of the code, but I am surprised we ever overflow.
Like I don't expect much to be done with fork than comparison.
In UTs we can always use a network where fork is not active to be sure we're safe

calbera: The overflow was happening in tests. Because simulating the slot after the fork upgrade

mod/state-transition/pkg/core/state_processor_staking.go
```diff
@@ -203,25 +208,31 @@ func (sp *StateProcessor[
 		return err
 	}
 	numWithdrawals := len(expectedWithdrawals)
+	if sp.cs.DepositEth1ChainID() == spec.BoonetEth1ChainID &&
+		slot >= spec.BoonetFork2Height {
+		// Enforce that there is at least one withdrawal for EVM inflation.
+		if numWithdrawals == 0 {
+			return ErrZeroWithdrawals
+		}
```
abi87, line 216: This is an extra check on a state that we control. I don't believe it is necessary, but I am fine keeping this as is, at least until we have properly validated all genesis and chain specs.

mod/state-transition/pkg/core/state_processor_staking.go
```diff
@@ -203,25 +208,31 @@ func (sp *StateProcessor[
 		return err
 	}
 	numWithdrawals := len(expectedWithdrawals)
+	if sp.cs.DepositEth1ChainID() == spec.BoonetEth1ChainID &&
+		slot >= spec.BoonetFork2Height {
+		// Enforce that there is at least one withdrawal for EVM inflation.
+		if numWithdrawals == 0 {
```
abi87, line 214: nit: `stateWithdrawals` or `expectedWithdrawals` instead on `numWithdrawals`?

calbera: `numWithdrawals` is derived from length of `expectedWithdrawals`.

mod/state-transition/pkg/core/state_processor_staking.go
abi87: No wait, this is not useless. We should require that we have at least one withdrawal not only on Boonet post fork2 but also in any network that is not Bartio and Boonet pre fork 2 (like the starting with main and ending with net).
Also for this network, we need block building logic to duly include the EVM inflation withdrawal, which I do not think we currently do.

mod/state-transition/pkg/core/state/statedb.go
```diff
@@ -228,6 +219,12 @@ func (s *StateDB[
 		return withdrawals, nil
 	}
 
+	if s.cs.DepositEth1ChainID() == spec.BoonetEth1ChainID &&
+		slot.Unwrap() >= spec.BoonetFork2Height {
```
abi87, line 223: I think this is wrong.
We should add the inflation withdrawal by default and do only two exceptions:
- Bartio
- Boonet pre Fork 2

Instead here any network that is not Cartio post fork 2 won't make the inflation withdrawal, which I think it's wrong

mod/config/pkg/spec/special_cases.go
```diff
@@ -20,19 +20,19 @@
 
 package spec
 
-import "math"
-
 // Special cased Bartio for some ad-hoc handling due to the way
 // some bugs were handled on Bartio. To be removed.
 const (
-	BartioChainID uint64 = 80084
+	BartioChainID = TestnetEth1ChainID
```
abi87, line 26: nit: I avoided doing that because I was not sure Bartio was our testnet really. This is fine with me as long as we have no plans to change `TestnetEth1ChainID` to make it point to a different network

mod/config/pkg/spec/special_cases.go
abi87: We just have to be careful that we don't accidentally activate this fork in tests. I think the value you pick is fine

mod/config/pkg/spec/special_cases.go
calbera: tbh its not really worth having both "Bartio" and "Testnet" in the code when they refer to the same thing

calbera:
https://github.com/coderabbitai pause

abi87 (approved):
LGTM

PR #2160, opened 2024-11-18
fix(beacon, blockchain): Fix payload timestamp build and verification

abi87:
We need the timestamp of the execution payload to fullfil the following constraints:
- it should be strictly monotonic (otherwise EL will reject it)
- it must be capped (otherwise a malicious validator can set it to `math.Uint64` and effectively stop the network, since no further block can be built on top of such a block AND satisfy monotonicity)

https://github.com/berachain/beacon-kit/pull/2095 set timestamp in order to satisfy these constraint BUT it causes block building to fail occasionally, if consensus blocks are finalized rapidly.
The failure works as follows:

- $N^{th}$ block is produced non-optimistically, hence EL payload timestamp is set equal to ProposeBlock $req_N.Time == T_0$
- $N^{th}$ block is processed and finalized rapidly, before 1 second pass:
  - when the block successfully pass processing, the next payload is optimistically built with EL payload timestamp set to $req_N.Time +1 == T_0+1$ sec
- the node propose again $N+1^{th}$ block. Now two facts:
  - because previous block has been finalized less than 1 second ago, ProposeBlock request time is the same as previous block (cause it is rounded to 1 second precision). So $req_N.Time == req_{N+1}.Time == T_0$
  - the payload built optimistically at the previous step is included. Recall that it's timestamp is again $req_N.Time +1$
- $N+1^{th}$ block fails validation because it's not true that EL payload time < $req_N.Time +1$

This PR solve the issue by changing building and validation rules for EL payload timestamp as follows:

- Upon `PrepareBlock`, with EL payload non built optimistically previous rounds, we set EL timestamp to `max(req.Time, parentELpayload.Time+1)`
- Upon `Process/Finalize block`,  we optimistically build the payload setting EL timestampt to `max(parentReq.Time+1, parentELpayload.Time+1)` 
  - Note: `req.Time` will change if multiple rounds are needed to finalized the block. We assume hereonly that `req.Time` will be non-strictly increasing.
- Upon `ProcessBlock` we deem the block valid only if  `ELpayload.Time<= max(req.Time+1, parentELpayload.Time+1` 

The new proposal:
- explicitly guarantees that timestamp is strictly monotonic
- bounds timestamp to be not too far away from either parent EL payload or consensus time

What it lack wrt to current, bugged version, is that it is not anymore true that:

- for every block EL timestamp is smaller or equal to consensus block timestamp.

In case consensus blocks are produced rapidly, EL timestamp will increase of 1 second while consensus time may not increase (if rounded to 1 second precision)

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method `GetConsensusTime` across various components to replace the previous `GetNextPayloadTimestamp`, enhancing clarity in consensus timing.
	- Added functionality for verifying and calculating payload timestamps in the new `payloadtime` package.
	- Implemented unit tests for the `payloadtime` package to ensure timestamp generation and verification logic behaves as expected.

- **Bug Fixes**
	- Enhanced error handling for proposal heights and payload retrieval processes to improve robustness.

- **Refactor**
	- Updated multiple methods to align with new consensus time handling, ensuring consistent logic throughout the codebase.
	- Removed the `minPayloadDelay` field from the middleware, simplifying initialization.
	- Removed the error variable `ErrTooFarInTheFuture` to adjust timestamp validation logic.
	- Updated the `go.mod` files to reflect new dependencies and streamline module management.
	- Adjusted the handling of timestamps in the `ABCIMiddleware` methods for improved accuracy.


mod/consensus/pkg/cometbft/service/middleware/middleware.go
```diff
@@ -75,37 +70,16 @@ func NewABCIMiddleware[
...
-	// finalized) or not (build only after parent block has been finalized).
-	// Hence we need to guarantee that we provide to the execution layer a
-	// strictly increasing timestamp for any ABCI valid call sequence of
-	// Prepare/Propose/FinalizeBlock.
-	// To this purpose, we set next payload timestamp as follows:
-	// - ProposeBlock: req.Time
-	// - Prepare/FinalizeBlock: req.Time + minPayloadDelay
-	// Monotonicity across request sequences is ensured by construction of
-	// minPayloadDelay: no block can be finalized before minPayloadDelay.
-	minPayloadDelay := min(
```
abi87, line 96: so we don't depend anymore on CometBFT config

mod/state-transition/pkg/core/state_processor_payload.go
```diff
@@ -51,7 +51,7 @@ func (sp *StateProcessor[
 		"consensus height", blk.GetSlot().Unwrap(),
 		"payload height", payload.GetNumber().Unwrap(),
 		"payload timestamp", payload.GetTimestamp().Unwrap(),
-		"bound timestamp", ctx.GetNextPayloadTimestamp().Unwrap(),
+		"consensus timestamp", ctx.GetConsensusTime().Unwrap(),
 		"skip payload verification", ctx.GetSkipPayloadVerification(),
 	)
```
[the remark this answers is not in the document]
abi87: Don't think it's necessary rn. Will revisit later on if needed

mod/state-transition/pkg/core/state_processor_payload.go
```diff
@@ -111,23 +115,10 @@ func (sp *StateProcessor[
 	_, _, _, _, _, _, _, _, _, _, _, _, _,
 ]) validateStatelessPayload(
 	blk BeaconBlockT,
-	nextPayloadTimestamp math.U64,
 ) error {
 	body := blk.GetBody()
 	payload := body.GetExecutionPayload()
 
-	// We skip timestamp check on Bartio for backward compatibility reasons
```
abi87, line 119: moved to stateful validation because we need to pick up parent payload timestamp

mod/state-transition/pkg/core/types.go
```diff
@@ -122,10 +122,9 @@ type Context interface {
 	// GetProposerAddress returns the address of the validator
 	// selected by consensus to propose the block
 	GetProposerAddress() []byte
-	// GetNextPayloadTimestamp returns the timestamp proposed by
-	// consensus for the next payload to be proposed. It is also
-	// used to bound current payload upon validation
-	GetNextPayloadTimestamp() math.U64
+	// GetConsensusTime returns the timestamp of current consensus request.
+	// It is used to build next payload and to validate currentpayload.
+	GetConsensusTime() math.U64
```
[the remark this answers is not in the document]
abi87: nah it's fine. There are CometBFT constraints which are pretty well documented

mod/beacon/blockchain/receive.go
```diff
@@ -81,10 +81,17 @@
 		)
 
 		if s.shouldBuildOptimisticPayloads() {
+			var lph ExecutionPayloadHeaderT
+			lph, err = preState.GetLatestExecutionPayloadHeader()
+			if err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
abi87: I don't believe it's needed. LatestExecutionPayloadHeader should be set before any block post genesis is processed.

mod/beacon/payload-time/time.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/berachain/beacon-kit/mod/primitives/pkg/math"
+)
+
+// ErrTooFarInTheFuture is returned when the payload timestamp
+// in a block exceeds the time bound.
+var ErrTooFarInTheFuture = errors.New("timestamp too far in the future")
```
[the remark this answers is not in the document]
abi87: Indeed, kept it here and dropped the other definition

mod/beacon/payload-time/time.go
```diff
@@ -0,0 +1,72 @@
...
+	)
+	if payloadTimestamp <= bound {
+		return nil
+	}
+	return fmt.Errorf(
+		"%w: timestamp bound: %d, got: %d",
+		ErrTooFarInTheFuture,
+		bound, payloadTimestamp,
+	)
+}
```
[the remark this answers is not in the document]
abi87: metrics/logging will be review in depth in some upcoming PRs

mod/beacon/validator/block_builder.go
```diff
@@ -216,39 +216,45 @@ func (s *Service[
...
+	)
+	return s.localPayloadBuilder.RequestPayloadSync(
+		ctx,
+		st,
+		blk.GetSlot(),
+		nextPayloadTimestamp.Unwrap(),
+		blk.GetParentBlockRoot(),
+		lph.GetBlockHash(),
+		lph.GetParentHash(),
+	)
```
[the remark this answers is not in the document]
abi87: metrics/logging will be review in upcoming PRs

mod/beacon/validator/types.go
```diff
@@ -193,10 +193,9 @@ type SlotData[AttestationDataT, SlashingInfoT any] interface {
 	// GetProposerAddress returns the address of the validator
 	// selected by consensus to propose the block
 	GetProposerAddress() []byte
-	// GetNextPayloadTimestamp returns the timestamp proposed by
-	// consensus for the next payload to be proposed. It is also
-	// used to bound current payload upon validation
-	GetNextPayloadTimestamp() math.U64
+	// GetConsensusTime returns the timestamp of current consensus request.
+	// It is used to build next payload and to validate currentpayload.
+	GetConsensusTime() math.U64
```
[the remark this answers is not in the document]
abi87: this is directly CometBFT timestamp which is widely documented

mod/beacon/payload-time/time.go
```diff
@@ -0,0 +1,72 @@
...
+
+func Next(
+	consensusTime,
+	parentPayloadTimestamp math.U64,
+	buildOptimistically bool,
+) math.U64 {
+	delta := math.U64(0)
+	if buildOptimistically {
```
abi87, line 59: Note: we may drop this `buildOptimistically` distinction and always pick `max(consensusTime, parentPayloadTimestamp+1)`

mod/beacon/payload-time/time.go
```diff
@@ -0,0 +1,72 @@
...
+// in a block exceeds the time bound.
+var ErrTooFarInTheFuture = errors.New("timestamp too far in the future")
+
+func Verify(
+	consensusTime,
+	parentPayloadTimestamp,
+	payloadTimestamp math.U64,
+) error {
+	bound := max(
+		consensusTime+1,
```
abi87, line 40: `consensusTime+1` must be here as long as we build optimistically setting delta to 1

mod/beacon/payload-time/time.go
```diff
@@ -0,0 +1,72 @@
...
+		parentPayloadTimestamp+1,
+	)
+	if payloadTimestamp <= bound {
+		return nil
+	}
+	return fmt.Errorf(
+		"%w: timestamp bound: %d, got: %d",
+		ErrTooFarInTheFuture,
+		bound, payloadTimestamp,
+	)
```
calbera, line 50: for readability: personally prefer returning error in the if case and otherwise return nil. The pattern we use throughout mostly.

mod/consensus/pkg/types/slot_data.go
```diff
@@ -45,15 +45,15 @@ func (b *SlotData[AttestationDataT, SlashingInfoT]) New(
...
 ) *SlotData[AttestationDataT, SlashingInfoT] {
 	b = &SlotData[AttestationDataT, SlashingInfoT]{
 		slot:            slot,
 		attestationData: attestationData,
 		slashingInfo:    slashingInfo,
 		commonConsensusData: &commonConsensusData{
-			proposerAddress:      proposerAddress,
-			nextPayloadTimestamp: math.U64(nextPayloadTimestamp.Unix()),
+			proposerAddress: proposerAddress,
+			consensusTime:   math.U64(consensusTime.Unix()),
```
calbera, line 56: just clarifying, do we need to round this before calling .Unix() ?

calbera: or its already handled by comet

abi87: It's handled by Unix, which picks only the second part.
See the docs
```
// Unix returns t as a Unix time, the number of seconds elapsed
// since January 1, 1970 UTC. The result does not depend on the
// location associated with t.
// Unix-like operating systems often record time as a 32-bit
// count of seconds, but since the method here returns a 64-bit
// value it is valid for billions of years into the past or future.
```

mod/beacon/payload-time/time.go
abi87: I agree and I can revert. The reason I pick this version is to try and make clearer that the payloadTime == bond is fine with use, that it to say the bound is loose, which is usually not the case.
Happy to swap if you feel this is not a sufficient reason to break away from convention

mod/state-transition/go.mod
```diff
@@ -6,6 +6,7 @@ replace (
 	cosmossdk.io/api => cosmossdk.io/api v0.7.3-0.20240806152830-8fb47b368cd4
 	cosmossdk.io/core => cosmossdk.io/core v0.0.0-20240806152830-8fb47b368cd4
 	cosmossdk.io/core/testing => cosmossdk.io/core/testing v0.0.0-20240806152830-8fb47b368cd4
+	github.com/berachain/beacon-kit/mod/beacon => ../beacon/
```
abi87, line 9: this is needed because payloadtime package is in a currently untagged commit. Could be remove once we tag

calbera (approved):
lgtm 🚀

mod/beacon/payload-time/time.go
```diff
@@ -0,0 +1,72 @@
...
+		// slower than consensusTime+1 sec), we're still fine as long as
+		// Verify pass which should always to since:
+		// Next.consensusTime <= Verify.consensusTime
+		delta = 1
+	}
+	return max(
+		consensusTime+delta,
+		parentPayloadTimestamp+1,
+	)
+}
```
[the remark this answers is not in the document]
abi87: Following discussions IRL, we will keep this for now. Much is noted in this PR so we have context to revisit at a later time if necessary

mod/beacon/payload-time/time.go
abi87: Done

PR #2171, opened 2024-11-22
Add telemetry for diff btw payload and consensus timestamps

fridrik01:
**Context**
This PR adds a new telemetry gauge value `beacon_kit.state.payload_consensus_timestamp_diff` which represents the time difference (in seconds) between the payload and consensus timestamps. This allows us to detect if either falls behind.

**Test plan**
I started a local node and confirmed the metrics was emitted on http://localhost:26660/metrics. 

image

## Summary by CodeRabbit

- **New Features**
	- Introduced a no-op telemetry sink to avoid unnecessary processing when telemetry is not required.
	- Enhanced state processing capabilities by integrating telemetry metrics.
	- Added functionality for tracking and reporting differences between payload and consensus timestamps.
	- Introduced a new interface for structured metric reporting.

- **Bug Fixes**
	- Improved readability of logging in the state processor by refining timestamp handling.

- **Documentation**
	- Added comments to clarify the purpose of new telemetry-related fields and parameters.

### DIFF
--- a/mod/node-core/pkg/components/metrics/sink.go
+++ b/mod/node-core/pkg/components/metrics/sink.go
@@ -78,3 +78,20 @@ func argsToLabels(args ...string) []metrics.Label {
 	}
 	return labels
 }
+
+// NoOpTelemetrySink is a no-op implementation of the TelemetrySink interface.
+type NoOpTelemetrySink struct{}
+
+// NewNoOpMetricsSink creates a new NoOpTelemetrySink.
+func NewNoOpMetricsSink() NoOpTelemetrySink {
+	return NoOpTelemetrySink{}
+}
+
+// IncrementCounter is a no-op implementation of the TelemetrySink interface.
+func (NoOpTelemetrySink) IncrementCounter(key string, args ...string) {}
+
+// SetGauge is a no-op implementation of the TelemetrySink interface.
+func (NoOpTelemetrySink) SetGauge(key string, value int64, args ...string) {}
+
+// MeasureSince is a no-op implementation of the TelemetrySink interface.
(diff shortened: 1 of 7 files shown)


mod/node-core/pkg/components/metrics/sink.go
```diff
@@ -78,3 +78,20 @@ func argsToLabels(args ...string) []metrics.Label {
 	}
 	return labels
 }
+
+// NoOpTelemetrySink is a no-op implementation of the TelemetrySink interface.
+type NoOpTelemetrySink struct{}
+
+// NewNoOpMetricsSink creates a new NoOpTelemetrySink.
+func NewNoOpMetricsSink() NoOpTelemetrySink {
```
calbera, line 86: nit: would rename to `NewNoOpTelemetrySink`

calbera (approved):
LGTM for now, lets start with this. once deployed, let's add this to our beacon-kit dashboards on grafana

mod/state-transition/pkg/core/metrics.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+}
+
+func (s *stateProcessorMetrics) gaugeTimestamps(
+	payloadTimestamp uint64,
+	consensusTimestamp uint64,
+) {
+	// the diff can be positive or negative depending on whether the payload timestamp is
+	// ahead or behind the consensus timestamp
+	diff := int64(payloadTimestamp) - int64(consensusTimestamp) //#nosec:G701
```
calbera, line 43: generally we add a reason after the #nosec to explain why doing the #nosec is OK

fridrik01: I would assume this is safe but I ain't checking it, will update and  check for overflow before casting to make sure this is ok.

PR #2173, opened 2024-11-24
chore(all) updated mockery to 2.49.0

abi87:
Just run `make generate-check` + forbid alias resolution.

## Summary by CodeRabbit

- **New Features**
	- Introduced new configuration options in `.mockery.yaml` for improved mock generation.
  
- **Bug Fixes**
	- Updated type parameters across multiple mock files for enhanced type specificity.
  
- **Documentation**
	- Updated version comments in various mock files to reflect the upgrade from `mockery v2.48.0` to `v2.49.0`.

- **Chores**
	- Regenerated mock code across numerous files to align with the latest version of the mock generation tool, ensuring consistent functionality.


.mockery.yaml
```diff
@@ -27,6 +27,7 @@ dir: "{{.InterfaceDir}}/mocks"
 mockname: "{{.InterfaceNameCamel}}"
 filename: "{{.InterfaceNameSnake}}.mock.go"
 outpkg: "mocks"
+resolve-type-alias: False # see https://vektra.github.io/mockery/latest/deprecations/#resolve-type-alias
```
abi87, line 30: This seems to me the bare minimum to appease the `make generate` command. However we should consider if we need to add
```
issue-845-fix: True # see https://vektra.github.io/mockery/latest/deprecations/#issue-845-fix
```
too

nidhi-singh02: We don't have `inpackage` parameter I don't think we need fix for issue-845 ? The issue arises only when inpackage is set to true.

> mockery ignored the outpkg: parameter if inpackage: was set to True

fridrik01: I ran `make generate` with this fix and it outputs multiple warnings that we should set the issue-845-fix to True. When I set that option the warnings went away so we probably should add that too?

calbera (approved):
OK with disabling resolve-type-alias

nidhi-singh02 (approved):
let's merge this, I will create an issue for the warnings that are coming, not super important for us rn, to track it.

PR #2175, opened 2024-11-24
fix: kurtosis install script

fridrik01:
There are two problems here:

1. On Ubuntu, the `command -v kurtosis &> /dev/null` fails to detect if kurtosis is installed resulting in it trying to install it again.

2. When extracting the `kurtosis.tar.gz` artifact it generates a `kurtosis` binary in current folder. However we already have a `kurtosis/` folder in our repo so the extraction fails.

This commit fixes 1) so it works on any platform and addresses 2) by removing the kurtosis automatic installation which should be fine since its already a prerequisite in our `kurtosis/README.md`.

## Summary by CodeRabbit

- **New Features**
	- Introduced a verification step to check if the Kurtosis command is installed, enhancing user guidance for setup.
  
- **Bug Fixes**
	- Streamlined the control flow for starting the development network by ensuring proper installation verification.

- **Documentation**
	- Updated prompts to direct users to the README for installation instructions.

### DIFF
--- a/kurtosis/Makefile
+++ b/kurtosis/Makefile
@@ -4,34 +4,11 @@
 ###                                Kurtosis                                 ###
 ###############################################################################
 
-# Installs Kurtosis if not already installed
-install-kurtosis:
-	@echo "Checking for Kurtosis installation..."
-	@if ! command -v kurtosis &> /dev/null; then \
-		echo "Kurtosis could not be found, installing..."; \
-		OS=$$(uname -s | tr A-Z a-z); \
-		if [ "$$OS" = "darwin" ]; then \
-			brew install kurtosis-tech/tap/kurtosis-cli; \
-		elif [ "$$OS" = "linux" ]; then \
-			ARCH=$$(uname -m); \
-			if [ "$$ARCH" = "x86_64" ]; then ARCH="amd64"; \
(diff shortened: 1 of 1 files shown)


kurtosis/Makefile
```diff
@@ -4,34 +4,11 @@
...
-			tar -xzf kurtosis.tar.gz; \
-			rm kurtosis.tar.gz; \
-			chmod +x kurtosis; \
-			sudo mv kurtosis /usr/local/bin/; \
-		else \
-			echo "Unsupported OS $$OS for Kurtosis installation" && exit 1; \
-		fi; \
-	else \
-		echo "Kurtosis is already installed"; \
-	fi
```
fridrik01, line 31: I mentioned during planning today about being fine with fixing the Ubuntu compatibility issue and keeping this install code. I changed my mind as this blob of code is complex and difficult to test and guarantee it works on different platforms. I think we should defer the installation to the user as a prerequisite, its already mentioned in the kurtosis/README.

nidhi-singh02: if we collectively decide we wanna get rid of the install code and let the user ensure kurtosis is installed before spinning up the local network, I am okay @abi87 @calbera wanted to know your thoughts

fridrik01:
Partly replaced with https://github.com/berachain/beacon-kit/pull/2470.

PR #2178, opened 2024-11-25
increase lint max line-length

fridrik01:
Currently running `make lint` runs linting according to the `.golangci.yaml` conf. In that file we set a max line-length of 80 which outputs linting errors for lines longer than that and results in failed CI. 

I propose we increase that limit to 140 as it allows more code to be viewed on screen at once as we usually have more horizontal screen space than vertical.  Go is also [typically not strict on line length](https://google.github.io/styleguide/go/guide#line-length). 

I would be fine with having no upper limit and make it up to the developer to make sure their code is readable, but lets start with this to get a discussion going.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -41,7 +41,7 @@ linters-settings:
     # even if all enum members are not listed.
     # Default: false
     default-signifies-exhaustive: true
-    
+
   exhaustruct:
     # List of regular expressions to exclude struct packages and their names from checks.
     # Regular expressions must match complete canonical struct package/name/structname.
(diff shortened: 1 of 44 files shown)


Gabrielcdecarvalho (commented):
@@ -147,7 +147,7 @@ linters-settings:
    # Max line length, lines longer will be reported.
    # '\t' is counted as 1 character by default, and can be changed with the tab-width option.
    # Default: 120.
    line-length: 150  # Increased maximum line length.
    # Tab width in spaces.
    # Default: 1
    tab-width: 1

calbera (approved):

nidhi-singh02 (approved):
lgtm!

PR #2181, opened 2024-11-27
Check for int64 overflow when computing uint64 diff

fridrik01:
In https://github.com/berachain/beacon-kit/pull/2171 we added the `beacon_kit.state.payload_consensus_timestamp_diff` metric for tracking the diff between the payload and consensus timestamps. 

In that PR we ignored int64 overflow which this PR addresses. Here we explicitly check ahead of time if an overflow will happen and log an error if that is the case.

### DIFF
--- a/mod/state-transition/pkg/core/state_processor_payload.go
+++ b/mod/state-transition/pkg/core/state_processor_payload.go
@@ -51,7 +51,13 @@ func (sp *StateProcessor[
 	payloadTimestamp := payload.GetTimestamp().Unwrap()
 	consensusTimestamp := ctx.GetConsensusTime().Unwrap()
 
-	sp.metrics.gaugeTimestamps(payloadTimestamp, consensusTimestamp)
+	// add telemetry for the diff between payloadTimestamp and
+	// consensusTimestamp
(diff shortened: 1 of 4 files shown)


mod/state-transition/pkg/core/state_processor_payload.go
```diff
@@ -51,7 +51,13 @@ func (sp *StateProcessor[
 	payloadTimestamp := payload.GetTimestamp().Unwrap()
 	consensusTimestamp := ctx.GetConsensusTime().Unwrap()
 
-	sp.metrics.gaugeTimestamps(payloadTimestamp, consensusTimestamp)
+	// add telemetry for the diff between payloadTimestamp and
+	// consensusTimestamp
+	err := sp.metrics.gaugePayloadConsensusTimestampDiff(
+		payloadTimestamp, consensusTimestamp)
```
abi87, line 57: How about instead of all of this we go though time.Duration?
Turn the timestamp into a time.Time objects and get diff as TimeA.Sub(TimeB)?

fridrik01:
Don't think this PR is necessary

PR #2191, opened 2024-11-28
protect valSetByEpoch for concurrent access

abi87:
I recently introduced a `valSetByEpoch` map to track some validator sets by epoch. The map must be protected for concurrent access, since `Transition` function is called both by block verification/finalization routines (sequentially, so no issues) and by block building (concurrently, to check that produced block is valid)

This PR fixes the issue

## Summary by CodeRabbit

- **New Features**
	- Enhanced thread safety for validator set updates with the introduction of a locking mechanism.
- **Bug Fixes**
	- Improved concurrency control to prevent potential race conditions when accessing validator sets.

### DIFF
--- a/mod/state-transition/pkg/core/state_processor.go
+++ b/mod/state-transition/pkg/core/state_processor.go
@@ -23,6 +23,7 @@ package core
 import (
 	"bytes"
 	"fmt"
+	"sync"
 
 	"github.com/berachain/beacon-kit/mod/config/pkg/spec"
 	"github.com/berachain/beacon-kit/mod/consensus-types/pkg/types"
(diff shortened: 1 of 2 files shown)


mod/state-transition/pkg/core/state_processor.go
```diff
@@ -109,6 +110,9 @@ type StateProcessor[
 	// ones.
 	// We prune the map to preserve only current and previous epoch
 	valSetByEpoch map[math.Epoch][]ValidatorT
+
+	// valSetMu protects valSetByEpoch from concurrent accesses
+	valSetMu sync.RWMutex
```
fridrik01, line 115: nit: usually a mutex guarding a struct member is put above it 

```suggestion
	// valSetMu protects valSetByEpoch from concurrent accesses
	valSetMu sync.RWMutex
	// We prune the map to preserve only current and previous epoch
	valSetByEpoch map[math.Epoch][]ValidatorT
```

abi87: Done

PR #2197, opened 2024-12-02
Fork dependent Chain spec

abi87:
This is an attempt at making chain specs fork dependent that is more structural than https://github.com/berachain/beacon-kit/pull/2194 but still incomplete. This PR:
- drops `chan.SpecData` structure and replaces it with `spec.Common` struct, where it can access chainIDs and forks
- adds `Get` prefix to `chain.Spec` interface to avoid conflicts with attributes
- drops chain spec references in `BlockBodyKZGOffset` to avoid import cycles among `chainSpec` and `consensus-types` modules
- drops `common.ChainSpec` type instantiation and uses `spec.Chain` instead.
- TODO: drops `chain-spec` module and just uses `config\spec`


calbera:
outdated.. RIP

PR #2199, opened 2024-12-02
[WIP ]Use single go.mod for whole beacon-kit

fridrik01:
WIP


mod/primitives/pkg/merkle/root_hasher_test.go
```diff
@@ -279,14 +279,14 @@ func requireGoHashTreeEquivalence(
 	// Ensure the lengths are the same
 	require.Equal(
 		t, len(expectedOutput), len(output),
-		fmt.Sprintf("Expected output length %d, got %d",
+		fmt.Errorf("Expected output length %d, got %d",
```
abi87, line 282: @fridrik01 you can actually drop this. This is mostly CodeRabbit allucinating and asking for more context of test failures. However require already gives plenty of info here

.dockerignore
```diff
@@ -11,5 +11,3 @@ testing/
 .git/
 .vscode/
 *.md
-go.sum
-go.mod
```
fridrik01, line 15: If we ignore the go.mod/sum files then its not possible to copy these files when building the docker image as it will hide them. This I had to figure out the hard way, basically after trying everything else until I took a look at this .dockerignore file and tried commenting this out...

mod/primitives/pkg/merkle/root_hasher_test.go
abi87: @fridrik01 what I mean is, you can drop the whole error message and just have `require.Equal(t, len(expectedOutput), len(output))`

mod/node-core/pkg/builder/builder.go
```diff
@@ -116,6 +116,7 @@ func (nb *NodeBuilder[NodeT, LoggerT, LoggerConfigT]) Build(
 	}
 
 	// TODO: so hood
+	//nolint:errcheck // should be safe
```
abi87, line 119: q: why does this come up only now?

kurtosis/src/nodes/consensus/beacond/scripts/multiple-premined-deposits.sh
```diff
@@ -1,4 +1,24 @@
 #!/bin/bash
+# SPDX-License-Identifier: BUSL-1.1
```
abi87, line 2: @fridrik01 nice

.dockerignore
abi87: @fridrik01 can we keep them ignored and just explicitly copying them over as we used to do in Dockerfile?

Dockerfile
```diff
@@ -27,68 +27,7 @@ ARG DB_BACKEND=pebbledb
...
-    go work use ./mod/payload && \
-    go work use ./mod/primitives && \
-    go work use ./mod/state-transition && \
-    go work use ./mod/storage && \
-    go work use ./mod/geth-primitives && \
-    go work use ./mod/chain-spec
-
-RUN --mount=type=cache,target=/root/.cache/go-build \
-    --mount=type=cache,target=/root/go/pkg/mod \
-    go mod download
```
abi87, line 88: don't we need to run this anyhow? How is that we can drop it?

abi87 (changes requested):
a few questions, but overall I think this is what we need

mod/node-core/pkg/node/node.go
```diff
@@ -50,6 +50,7 @@ type node struct {
 // New returns a new node.
 func New[NodeT types.Node](
 	registry *service.Registry, logger log.Logger) NodeT {
+	//nolint:errcheck // should be safe
```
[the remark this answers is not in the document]
fridrik01: I assume its slightly newer dependencies so its catching more now

Dockerfile
fridrik01: Good catch, I was wondering myself if this is needed with migrating to single module. When I compared the build job to `main` branch it went from ~50s ([job](https://github.com/berachain/beacon-kit/actions/runs/12125478611/job/33805523378?pr=2199)) to 3s ([job](https://github.com/berachain/beacon-kit/actions/runs/12125478611/job/33805523378?pr=2199)) so expected some caching to be there already and didn't think we needed this (other jobs were also often faster). I'll try enabling it again, it should not get worse though

.dockerignore
fridrik01: No we can't keep them ignored as then we wont be able to copy them inside the `Dockerfile` as they are kept hidden from it (I probably spent good 1-2 hours debugging that)

mod/primitives/pkg/merkle/root_hasher_test.go
```diff
@@ -277,20 +277,11 @@ func requireGoHashTreeEquivalence(
 	}
 
 	// Ensure the lengths are the same
-	require.Equal(
-		t, len(expectedOutput), len(output),
-		fmt.Sprintf("Expected output length %d, got %d",
-			len(expectedOutput), len(output)))
+	require.Equal(t, len(expectedOutput), len(output))
```
abi87, line 280: Awesome

mod/consensus/pkg/cometbft/service/params/consensus.go
```diff
@@ -47,6 +47,7 @@ func NewConsensusParamsStore(cs ChainSpec) *ConsensusParamsStore {
 // Get retrieves the consensus parameters from the store.
 // It returns the consensus parameters and an error, if any.
 func (s *ConsensusParamsStore) Get() *cmtproto.ConsensusParams {
```
abi87, line 49: @fridrik01 I believe this is fine. I would suggest the following approach to drop the linting:
```
	cfg, _ := s.cs.GetCometBFTConfigForSlot(0).(*cmttypes.ConsensusParams)
	p := cfg.ToProto()
	return &p
```

abi87 (approved):
LGTM

PR #2207, opened 2024-12-03
chore(primitives): add recover to MustToBytes

nidhi-singh02:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling in the hex conversion process by replacing the `MustToBytes` function with `ToBytesSafe`, preventing program crashes and allowing for safer error management.

- **New Features**
	- Enhanced overall robustness of the application by implementing safer hex string to byte conversions across various components, including state processing and execution handling.
	- Updated test suites to reflect the new error handling logic, ensuring consistent validation of hex string conversions.

### DIFF
--- a/mod/da/pkg/kzg/ckzg/ckzg.go
+++ b/mod/da/pkg/kzg/ckzg/ckzg.go
@@ -49,11 +49,11 @@ func NewVerifier(ts *gokzg4844.JSONTrustedSetup) (*Verifier, error) {
 		len(ts.SetupG1Lagrange)*(len(ts.SetupG1Lagrange[0])-2)/2,
 	)
 	for i, g1 := range ts.SetupG1Lagrange {
-		copy(g1s[i*(len(g1)-2)/2:], hex.MustToBytes(g1))
+		copy(g1s[i*(len(g1)-2)/2:], hex.ToBytesSafe(g1))
 	}
 	g2s := make([]byte, len(ts.SetupG2)*(len(ts.SetupG2[0])-2)/2)
 	for i, g2 := range ts.SetupG2 {
-		copy(g2s[i*(len(g2)-2)/2:], hex.MustToBytes(g2))
+		copy(g2s[i*(len(g2)-2)/2:], hex.ToBytesSafe(g2))
 	}
 	if err := ckzg4844.LoadTrustedSetup(g1s, g2s); err != nil {
 		return nil, err
--- a/mod/primitives/pkg/encoding/hex/bytes.go
+++ b/mod/primitives/pkg/encoding/hex/bytes.go
@@ -22,7 +22,6 @@ package hex
 
 import (
 	"encoding/hex"
-	"fmt"
 
 	"github.com/berachain/beacon-kit/mod/errors"
 )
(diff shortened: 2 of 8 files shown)


mod/da/pkg/kzg/ckzg/ckzg.go
```diff
@@ -49,11 +49,11 @@ func NewVerifier(ts *gokzg4844.JSONTrustedSetup) (*Verifier, error) {
 		len(ts.SetupG1Lagrange)*(len(ts.SetupG1Lagrange[0])-2)/2,
 	)
 	for i, g1 := range ts.SetupG1Lagrange {
-		copy(g1s[i*(len(g1)-2)/2:], hex.MustToBytes(g1))
+		copy(g1s[i*(len(g1)-2)/2:], hex.ToBytesSafe(g1))
```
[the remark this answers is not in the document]
nidhi-singh02: seems like coderabbit is having a valid point, what do you think @berachain/core ? It will also happen in other places where `ToBytesSafe()` is used. Are we okay with this ? I am wondering that eventually we would end up in using `ToBytes` and removing `ToBytesSafe` altogether.

mod/primitives/pkg/encoding/hex/bytes.go
```diff
@@ -29,17 +29,23 @@ import (
...
-func MustToBytes(input string) []byte {
+// ToBytesSafe returns the bytes represented by the given hex string.
+// If the input is not a valid hex string, it returns an empty byte slice.
+func ToBytesSafe(input string) []byte {
+	var bz []byte
+	defer func() {
+		if r := recover(); r != nil {
+			bz = []byte{}
+		}
+	}()
```
nidhi-singh02, line 48: I have renamed `MustToBytes` to `ToBytesSafe` to represent the behaviour accurately as it is no loner panic-ing on invalid hex string.

mod/primitives/pkg/encoding/hex/bytes.go
```diff
@@ -29,17 +29,23 @@ import (
...
 	hex.Encode(hexStr[prefixLen:], b)
 	return string(hexStr)
 }
 
-// MustToBytes returns the bytes represented by the given hex string.
-// It panics if the input is not a valid hex string.
-func MustToBytes(input string) []byte {
+// ToBytesSafe returns the bytes represented by the given hex string.
+// If the input is not a valid hex string, it returns an empty byte slice.
+func ToBytesSafe(input string) []byte {
```
fridrik01, line 42: Currently we are only calling `hex.MustToBytes` in about dozen places in our codebase. I think in all places we can propagate any error with marshalling a string to bytes back to the caller to handle it, so we don't need to silently ignore any errors with the conversion.

So I would suggest  changing the function signature to something like:

```
func ToBytesSafe(input string) ([]byte, error)
```

nidhi-singh02: Yeah, Fair point! There's already a func  `ToBytes` doing this. Here I am referring it `https://github.com/berachain/beacon-kit/pull/2207#discussion_r1868763188 , we may end up using `ToBytes` everywhere.

nidhi-singh02:
`MustToBytes()` panic is not really a valid issue as it is fail early pattern that we follow by panic-ing in case genesis data is bad which is good for us. Closing the PR.

PR #2202, opened 2024-12-03
fix(state-transitions): Make setting of eth1 deposit index backward compatible

abi87:
A proposal to replace https://github.com/berachain/beacon-kit/pull/2201
- backward compatility ensured
- added logs for debugging visibility
- unit tests use clean chain spec without fork version mess

## Release Notes

- **New Features**
  - Enhanced logging for deposit processing and state transitions, providing better visibility into the processes.
  
- **Bug Fixes**
  - Improved error handling in various methods, ensuring clearer context in error messages.
  - Adjusted loop structures in the pruning method to prevent unintended behavior.

- **Documentation**
  - Updated comments in test functions for better clarity on expected behaviors and outcomes.

- **Refactor**
  - Removed unused variables and streamlined type definitions in test files to enhance code quality.

PR #2219, opened 2024-12-05
fix(state-transition): restore state sync over boonet

abi87:
Sync is currently broken on Boonet. Bysecting I found:

- https://github.com/berachain/beacon-kit/pull/2158 broke syncing this but making a non-backward compatible change to `MaxValidatorsPerWithdrawalsSweep
- https://github.com/berachain/beacon-kit/pull/2194 fixes syncing  again
- https://github.com/berachain/beacon-kit/pull/2202 broke syncing again

The latest PR breaks sync because if mistakenly changes the indexes for genesis deposits, and only those
These indexes are part of the beacon state that is hashed into blocks. So state sync was broken because:
- Genesis state is subtly modified
- Block 1, already finalized and downloaded from a peer, fails verification because `AppHash` reported by the block does not match the `AppHash` generated by the local state.

The fix is to preserve the way we set indexes over genesis

MOREOVER

We need to properly drop uselass BeaconState ops to preserve appHash

### DIFF
--- a/state-transition/core/state_processor_staking.go
+++ b/state-transition/core/state_processor_staking.go
@@ -84,7 +84,7 @@ func (sp *StateProcessor[
 		// We keep it for backward compatibility.
 		depositIndex++
 	case sp.cs.DepositEth1ChainID() == spec.BoonetEth1ChainID &&
-		slot < math.U64(spec.BoonetFork2Height):
+		slot != 0 && slot < math.U64(spec.BoonetFork2Height):
 		// Boonet pre fork2 has a bug which makes DepositEth1ChainID point to
 		// next deposit index, not latest processed deposit index.
 		// We keep it for backward compatibility.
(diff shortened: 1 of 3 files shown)


calbera (approved):
LGTM

state-transition/core/state_processor_staking.go
```diff
@@ -84,7 +84,7 @@ func (sp *StateProcessor[
 		// We keep it for backward compatibility.
 		depositIndex++
 	case sp.cs.DepositEth1ChainID() == spec.BoonetEth1ChainID &&
-		slot < math.U64(spec.BoonetFork2Height):
+		slot != 0 && slot < math.U64(spec.BoonetFork2Height):
```
abi87, line 87: genesis indexes were set right, so we don't correct them.

calbera (approved):
add one comment then LGTM

PR #2221, opened 2024-12-06
chore(all): update cosmos-sdk deps

abi87:
Imports the awesome work by @alesforz.

- Updates go dependencies to `1.23.4`
- Updates `go.mod` to point to latest cometBFT and cosmosSDK
- Replaces `cosmossdk.io/store/v2/db` with `"github.com/cosmos/cosmos-db"` (see https://github.com/berachain/beacon-kit/pull/2300 for context)
- Fixes compressed BLS keys handling


node-core/components/signer/legacy.go
```diff
@@ -30,7 +30,7 @@ import (
 
 // LegacySigner is a BLS12-381 signer that uses a bls.PrivKey for signing.
 type LegacySigner struct {
-	*bls12381.PrivKey
+	bls12381.PrivKey
```
nidhi-singh02, line 33: Q: is removing the pointer really required ?

testing/e2e/e2e_staking_test.go
```diff
@@ -98,14 +99,22 @@ func (s *BeaconKitE2ESuite) TestDepositRobustness() {
 		// Create a deposit transaction. Use the default validators' pubkeys
 		// if exists, otherwise pubkey is a random 48 byte slice.
 		var pubkey []byte
+		var pk *bls12381.PubKey
+
 		switch i {
 		case 0:
 			pubkey, err = client.GetPubKey(s.Ctx())
 			s.Require().NoError(err)
+			pk, err = bls12381.NewPublicKeyFromBytes(pubkey)
```
calbera, line 108: nit: can move this logic into the e2e testing client package to simplify here

calbera (approved):

cli/commands/server/start.go
```diff
@@ -106,6 +107,10 @@ custom: allow pruning options to be manually specified through 'pruning-keep-rec
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			logger := clicontext.GetLoggerFromCmd[LoggerT](cmd)
 			cfg := clicontext.GetConfigFromCmd(cmd)
+			if cfg.Consensus.TimeoutCommit == 0 {
+				return errors.New("please edit your config.toml file and set timeout_commit to 1s")
+			}
+
```
emostov, line 113: What was the motivation for adding this check?

abi87: >  @abi87 What was the motivation for adding this check? Is this related to the dep updated or just wanting to  prevent footguns and ensure >=1s block times?

Immediate tests following dep updates showed a large number of consensus blocks produced pretty rapidly. We realized some of configs were not properly handled (wrongly picking cometBFT default values instead of BeaconKit) and we added this check as a fail fast kind of solution

emostov: Thank you for the context - this is very helpful!

PR #2223, opened 2024-12-06
fix(e2e): Allow more gas to be spent in deposits e2e test

calbera:
Symptom of using the new deposit contract which requires more gas

### DIFF
--- a/testing/e2e/e2e_staking_test.go
+++ b/testing/e2e/e2e_staking_test.go
@@ -152,13 +152,12 @@ func (s *BeaconKitE2ESuite) TestDepositRobustness() {
 	s.Require().NoError(err)
 
 	// Check that the eth spent is somewhere~ (gas) between
-	// upper bound: 32ether * 500 + 1ether
+	// upper bound: 32ether * 500 + 2ether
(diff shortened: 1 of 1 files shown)


testing/e2e/e2e_staking_test.go
```diff
@@ -152,13 +152,12 @@ func (s *BeaconKitE2ESuite) TestDepositRobustness() {
...
 	// Check that the eth spent is somewhere~ (gas) between
-	// upper bound: 32ether * 500 + 1ether
+	// upper bound: 32ether * 500 + 2ether
 	// lower bound: 32ether * 500
-	oneEther := big.NewInt(1e18)
-	totalAmt := new(big.Int).Mul(
-		oneEther, big.NewInt(NumDepositsLoad*32),
-	)
-	upperBound := new(big.Int).Add(totalAmt, oneEther)
+	twoEther := big.NewInt(1e18)
```
abi87, line 157: These are not two ethers?

calbera: whoops, fixing. the test still passed here

abi87 (approved):

PR #2225, opened 2024-12-07
Remove all async events and replace with direct synchronous calls

fridrik01:
This PR removes all use of async events from the codebase and replaces them with direct synchronous calls through an interface injected by the DI library. The last commit then removes all async related code from the codebase.

STATUS (List of events deprecated by this PR):
- [x] GenDataReceived
- [x] GenDataProcessed
- [x] NewSlot
- [x] BuiltBeaconBlock
- [x] BuiltSidecars
- [x] BeaconBlockReceived
- [x] SidecarsReceived
- [x] BeaconBlockVerified
- [x] SidecarsVerified
- [x] FinalBeaconBlockReceived
- [x] FinalSidecarsReceived
- [x] FinalValidatorUpdatesProcessed


consensus/types/slot_data.go
```diff
@@ -39,15 +39,15 @@ type SlotData[AttestationDataT, SlashingInfoT any] struct {
 	*commonConsensusData
 }
 
-// New creates a new SlotData instance.
-func (b *SlotData[AttestationDataT, SlashingInfoT]) New(
```
abi87, line 43: Nice

primitives/async/id.go
```diff
@@ -1,48 +0,0 @@
...
-// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, NON-INFRINGEMENT, AND
-// TITLE.
-
-package async
-
-// EventID represents the type of a message.
-type EventID string
-
-// event ids, topologically sorted.
-const (
```
abi87, line 27: so long and thanks for all the fish

consensus/cometbft/service/service.go
```diff
@@ -102,9 +112,12 @@ func NewService[
 			db,
 			servercmtlog.WrapSDKLogger(logger),
 		),
-		Middleware: middleware,
-		cmtCfg:     cmtCfg,
-		paramStore: params.NewConsensusParamsStore(cs),
+		Blockchain:    blockchain,
+		BlockBuilder:  blockBuilder,
+		cmtCfg:        cmtCfg,
+		chainSpec:     cs,
```
abi87, line 118: @fridrik01 seems to me you don't need to store `chainSpec` here. This is just used to create paramStore

beacon/blockchain/types.go
```diff
@@ -240,4 +248,18 @@ type TelemetrySink interface {
 	MeasureSince(key string, start time.Time, args ...string)
 }
 
+//nolint:revive // We will remove this once generics are cleaned up
+type BlockchainI[GenesisT any] interface {
```
abi87, line 252: @fridrik01  it seems to me we can drop this parameter. I can compile a non-generic `BlockchainI`. Why is it here?

beacon/validator/types.go
```diff
@@ -237,3 +239,24 @@ type TelemetrySink interface {
...
+	constraints.SSZMarshallable
+	constraints.Empty[T]
+	Len() int
+	Get(index int) BlobSidecarT
+	GetSidecars() []BlobSidecarT
+	ValidateBlockRoots() error
+	VerifyInclusionProofs(kzgOffset uint64) error
+}
+
+type BlockBuilderI[SlotDataT any] interface {
```
abi87, line 254: @fridrik01 it seems to be the generic is not really needed here. Can we drop it?

consensus/cometbft/service/service.go
```diff
@@ -50,12 +55,15 @@ const (
...
+	chainSpec     common.ChainSpec
+	telemetrySink TelemetrySink
 
-	logger     LoggerT
-	sm         *statem.Manager
-	Middleware MiddlewareI
+	logger       LoggerT
+	sm           *statem.Manager
+	Blockchain   blockchain.BlockchainI[any]
+	BlockBuilder validator.BlockBuilderI[any]
```
abi87, line 66: again I think we can drop these generics here.

consensus/cometbft/service/process_proposal.go
```diff
@@ -0,0 +1,89 @@
...
+			fmt.Sprintf("%X", req.Hash),
+			"err",
+			err,
+		)
+		return &cmtabci.ProcessProposalResponse{
+			Status: cmtabci.PROCESS_PROPOSAL_STATUS_REJECT,
+		}, nil
+	}
+
+	return resp, nil
```
abi87, line 88: @fridrik01 is there a way to avoid passing `req` to the Blockchain service and just pass the content?
I think we want to get to that state sooner rather than later (let CosmosSDK types stay within consensus package only), I wonder if we can do this here already

consensus/cometbft/service/prepare_proposal.go
```diff
@@ -0,0 +1,94 @@
...
+	](
+		math.Slot(req.GetHeight()),
+		nil,
+		nil,
+		req.GetProposerAddress(),
+		req.GetTime(),
+	)
+
+	// TODO: can we use the context from the request now that async is gone?
+	//nolint:contextcheck // see todo above
```
abi87, line 73: is this comment still relevant? I don't think we can do that. But we may need to replace the base context with the one from the request, in order to properly handle potential cancellations from CometBFT

consensus/cometbft/service/middleware/errors.go
```diff
@@ -1,66 +0,0 @@
...
-// EXPRESS OR IMPLIED, INCLUDING (WITHOUT LIMITATION) WARRANTIES OF
-// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, NON-INFRINGEMENT, AND
-// TITLE.
-
-package middleware
-
-import "github.com/berachain/beacon-kit/errors"
-
-//nolint:gochecknoglobals // errors
-var (
```
abi87, line 26: all of these errors were related to the async system right? So all gone, nice

consensus/cometbft/service/middleware/constants.go
```diff
@@ -1,34 +0,0 @@
...
-
-import "time"
-
-const (
-	// BeaconBlockTxIndex represents the index of the beacon block transaction.
-	// It is the first transaction in the tx list.
-	BeaconBlockTxIndex uint = iota
-	// BlobSidecarsTxIndex represents the index of the blob sidecar transaction.
-	// It follows the beacon block transaction in the tx list.
-	BlobSidecarsTxIndex
```
abi87, line 31: indexes move to `blockchain/process.go` file, timeout dropped

consensus/cometbft/service/genesis.go
```diff
@@ -1,136 +0,0 @@
...
-	"github.com/berachain/beacon-kit/consensus-types/types"
-	"github.com/berachain/beacon-kit/errors"
-	"github.com/berachain/beacon-kit/primitives/encoding/json"
-	cmtcfg "github.com/cometbft/cometbft/config"
-	"github.com/cometbft/cometbft/node"
-	genutiltypes "github.com/cosmos/cosmos-sdk/x/genutil/types"
-)
-
-// DefaultGenesis returns the default genesis state for the application.
-func (s *Service[_]) DefaultGenesis() map[string]json.RawMessage {
```
abi87, line 36: @fridrik01 we should maybe keep this genesis file? It holds data wrangling info, while init_chain will do the processing of the genesis info

beacon/validator/block_builder.go
```diff
@@ -37,22 +39,24 @@ import (
...
-		sidecars  BlobSidecarsT
-		startTime = time.Now()
-		g, _      = errgroup.WithContext(ctx)
-	)
+	slotData types.SlotData[ctypes.AttestationData, ctypes.SlashingInfo],
+) ([]byte, []byte, error) {
+	// as the async approach built blocks in single thread lets keep the
+	// same approach for now
+	s.blockBuilderMu.Lock()
+	defer s.blockBuilderMu.Unlock()
```
abi87, line 53: @fridrik01 I think we don't really need this mutex anymore, since this function is only called, directly from prepareBlock?

beacon/validator/block_builder.go
```diff
@@ -37,22 +39,24 @@ import (
...
 
+	startTime := time.Now()
 	defer s.metrics.measureRequestBlockForProposalTime(startTime)
 
+	g, _ := errgroup.WithContext(ctx)
```
abi87, line 58: Yesterday we merged https://github.com/berachain/beacon-kit/commit/8ec5187c97300b0d81d2a0eff4401bfcf2ce8a24 from @shotes that removes concurrency here. It is already in main, so please fix this

beacon/validator/service.go
```diff
@@ -80,8 +78,8 @@ type Service[
 	remotePayloadBuilders []PayloadBuilder[BeaconStateT, ExecutionPayloadT]
 	// metrics is a metrics collector.
 	metrics *validatorMetrics
-	// subNewSlot is a channel to hold NewSlot events.
-	subNewSlot chan async.Event[SlotDataT]
+	// mutex for building blocks.
+	blockBuilderMu sync.Mutex
```
abi87, line 82: seems to me this can be dropped

beacon/validator/types.go
fridrik01: Good catch, also the same with BlockchainI

consensus/cometbft/service/prepare_proposal.go
fridrik01: haha, I wrote that before I knew that we were saving the state from the context, will remove comment

consensus/cometbft/service/middleware/errors.go
fridrik01: Yes, these were all the timeout errors for different events, so all gone

consensus/cometbft/service/middleware/constants.go
fridrik01: Yes this moved to `blockchain/process.go` no need for timeouts

consensus/cometbft/service/genesis.go
fridrik01: done!

consensus/cometbft/service/process_proposal.go
fridrik01: Yes I wanted to do the unmarshaling in cometbft.Service but got a bit stuck in getting it to compile, let me try to take a another stab at this (potentially as a followup PR so we can get this merged in quickly)

beacon/validator/block_builder.go
fridrik01: done, rebased PR against main

PR #2231, opened 2024-12-09
fix(state-transition): enforce valid eth1 withdrawal credentials

calbera:
State Transition Changes:
- Ignore creation of validators with a non `0x01` byte prefix in their withdrawal credentials
- In statedb `ExpectedWithdrawals` only convert Credentials to Address if we know for sure it has the `0x01` byte prefix (to avoid erroring out)

No explicit backwards compatibility handling technically required since on bartio and boonet we do not encounter any deposits with a non `0x01` byte prefix in their withdrawal credentials

### DIFF
--- a/state-transition/core/state/statedb.go
+++ b/state-transition/core/state/statedb.go
@@ -268,16 +268,16 @@ func (s *StateDB[
 			return nil, err
 		}
 
-		withdrawalAddress, err = validator.
-			GetWithdrawalCredentials().ToExecutionAddress()
-		if err != nil {
-			return nil, err
-		}
-
 		// Set the amount of the withdrawal depending on the balance of the
 		// validator.
 		//nolint:gocritic // ok.
 		if validator.IsFullyWithdrawable(balance, epoch) {
+			withdrawalAddress, err = validator.
+				GetWithdrawalCredentials().ToExecutionAddress()
+			if err != nil {
+				return nil, err
+			}
+
 			withdrawals = append(withdrawals, withdrawal.New(
 				math.U64(withdrawalIndex),
 				validatorIndex,
@@ -290,6 +290,12 @@ func (s *StateDB[
 		} else if validator.IsPartiallyWithdrawable(
 			balance, math.Gwei(s.cs.MaxEffectiveBalance()),
 		) {
+			withdrawalAddress, err = validator.
+				GetWithdrawalCredentials().ToExecutionAddress()
+			if err != nil {
+				return nil, err
+			}
+
 			withdrawals = append(withdrawals, withdrawal.New(
 				math.U64(withdrawalIndex),
 				validatorIndex,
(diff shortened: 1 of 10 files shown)


state-transition/core/state/statedb.go
```diff
@@ -268,16 +268,16 @@ func (s *StateDB[
 			return nil, err
 		}
 
-		withdrawalAddress, err = validator.
```
abi87, line 271: We need to add `gocognit` to the list of no-lint for this function

state-transition/core/state_processor_staking.go
```diff
@@ -157,6 +158,13 @@ func (sp *StateProcessor[
 	// Get the current epoch.
 	epoch := sp.cs.SlotToEpoch(slot)
 
+	// Verify that the deposit has the ETH1 withdrawal credentials.
+	if dep.GetWithdrawalCredentials()[0] != types.EthSecp256k1CredentialPrefix {
+		return errors.New(
+			"deposit does not have ETH1 withdrawal credentials",
+		)
+	}
```
abi87, line 166: not sure we should err here. Similarly to what we do with faulty signature, I think this deposit should be skipped and the others be duly processed

abi87 (commented):
I have a major concern about handling of deposit with wrong credentails. I think we should skip them, not fail the whole block

state-transition/core/state_processor_staking.go
calbera: done

state-transition/core/state/statedb.go
calbera: done

state-transition/core/state/statedb.go
```diff
@@ -268,16 +268,16 @@ func (s *StateDB[
...
 		// Set the amount of the withdrawal depending on the balance of the
 		// validator.
-		//nolint:gocritic // ok.
+		//nolint:gocritic,nestif // ok.
 		if validator.IsFullyWithdrawable(balance, epoch) {
+			withdrawalAddress, err = validator.
+				GetWithdrawalCredentials().ToExecutionAddress()
+			if err != nil {
+				return nil, err
+			}
```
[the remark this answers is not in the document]
calbera: `ToExecutionAddress()` only errors if the withdrawal credentials are non `0x01` first byte. We know this error is not possible for two reasons:
1. `validator.Is(Fully/Partially)Withdrawable` only returns true if the credentials have a `0x01` first byte. Note: For the bartio case: all validators have credentials with `0x01` first byte.
2. Correct we only accept deposits with the `0x01` first byte credentials.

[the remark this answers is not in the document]
calbera: 2nd question: credentials are never reset and we don't use/check credentials in validator updates.

PR #2234, opened 2024-12-09
chore(all): no more BeaconBlockHeaderT among generics

abi87:
Dropping `BeaconBlockHeaderT` generic and replaced it with concrete type `*types.BeaconBlockHeader`

- [X] No more `grep -nr 'BeaconBlockHeaderT interface' `
- [X] no more `grep -nr 'BeaconBlockHeaderT any' `
- [X] no more `grep -nr 'BeaconBlockHeaderT'`


fridrik01:
See https://github.com/berachain/beacon-kit/pull/2243 which is based against newest `main`.

abi87:
replaced by https://github.com/berachain/beacon-kit/pull/2243

PR #2237, opened 2024-12-10
bug(validator): serialize StateRoot computation and Sidecar production

shotes:
While producing a `BeaconBlock` and `BlobSidecars` for proposal, the function `buildBlockAndSidecars()` starts two goroutines at the same time. These two goroutines have a data race on this `StateRoot` field:
1. `computeAndSetStateRoot()` which computes and sets the `BeaconBlock.StateRoot`
2. `BuildSidecars()` which will produce the sidecars for the given `BeaconBlock` and `BlobsBundle`. This will read the `BeaconBlock.StateRoot` and set its own `BlobSidecar.BeaconBlockHeader.StateRoot` field

This PR resolves this issue.

I was heavily considering an alternate solution where the parallel computation was preserved by doing the following:
- pre-fetch the `BeaconBlock.GetHeader()` and supplying this to `BuildSidecars()`. This contains the old `StateRoot` value and will need to get updated. But this pre-fetch needs to happen so that we aren't calling `GetHeader()` at the same time as setting the `StateRoot` which could result in some undefined behavior.
- running both goroutines
- setting the `StateRoot` in each sidecar after both goroutines return.

I opted for the simpler solution of serializing to be as non-intrusive as possible. If the parallelization is needed for performance, this can be implemented.


abi87 (approved):
LGTM

PR #2239, opened 2024-12-10
Gracefully stop services on SIGTERM and SIGINT signals

fridrik01:
I was looking into how we handle SIGTERM and SIGINT signals and noticed that we had a `cometbft.Service::Close()` doing which explicitly called `Stop` on the cometBFT node and closes the database connection gracefully (see relevant [code](https://github.com/berachain/beacon-kit/blob/main/consensus/cometbft/service/service.go#L161-L173
). 

I however noticed when I send our process a SIGTERM signal that these logs were not printed to STDOUT so I looked more into it looks to me we were not exiting gracefully.
 
With this PR, now once we receive the SIGTERM and SIGINT signals we will give each service the opportunity to close any outstanding sockets/connections it might have.


calbera (approved):
fair enough

nidhi-singh02 (approved):

node-core/services/registry/registry.go
```diff
@@ -85,6 +87,23 @@ func (s *Registry) StartAll(ctx context.Context) error {
...
+	s.logger.Info("Stopping services", "num", len(s.serviceTypes))
+	for _, typeName := range s.serviceTypes {
+		s.logger.Info("Stopping service", "type", typeName)
+		svc := s.services[typeName]
+		if svc == nil {
+			s.logger.Error("service not found", "type", typeName)
+			continue
+		}
+
+		if err := svc.Stop(); err != nil {
```
abi87, line 100: @fridrik01 we are assuming that `StopAll` won't ever be called before `Start` here. Should we explicitly enforce this? With some kind of  `started` atomic bool

node-core/node/node.go
```diff
@@ -73,7 +73,17 @@ func (n *node) Start(
...
-	return g.Wait()
+	err := g.Wait()
+	if err != nil {
+		return err
+	}
+
+	// Stopp each service allowing them the exit gracefully.
+	if err = n.registry.StopAll(); err != nil {
+		return err
+	}
```
abi87, line 84: should we defer here? So to properly stop all in any case?

node-core/services/registry/registry.go
fridrik01: I left a comment that it should be safe to call Stop on a service that was never started.

node-core/node/node.go
fridrik01: On partial startup of services I now call StopAll so make sure everything was cleaned up properly.

PR #2240, opened 2024-12-11
chore(blockchain) : non optimistic client should not request build optimistically

nidhi-singh02:
Non optimistic client currently request for build optimistically, fixing it with this changeset.


nidhi-singh02:
E2E tests are failing, I am debugging it.

shotes:
I believe the check for `shouldBuildOptimisticPayloads()` should be removed. This `sendPostBlockFCU()` function should always call `s.sendNextFCUWithoutAttributes(ctx, blk, lph)` whether it is optimistic or not, since all nodes need to notify the execution layer of the fork choice update.

nidhi-singh02:
> I believe the check for `shouldBuildOptimisticPayloads()` should be removed. This `sendPostBlockFCU()` function should always call `s.sendNextFCUWithoutAttributes(ctx, blk, lph)` whether it is optimistic or not, since all nodes need to notify the execution layer of the fork choice update.
> 
> Do note that removing the call to `s.sendNextFCUWithAttributes(ctx, st, blk, lph)` fixed the issue with triggering optimistic builds - this function should never trigger optimistic builds by sending FCU with attributes.

Makes sense, removing the check worked. Thanks for your valuable pointers @shotes!

beacon/blockchain/execution_engine.go
```diff
@@ -22,14 +22,15 @@ package blockchain
...
-	"github.com/berachain/beacon-kit/config/spec"
 	engineprimitives "github.com/berachain/beacon-kit/engine-primitives/engine-primitives"
 )
 
-// sendPostBlockFCU sends a forkchoice update to the execution client.
+// sendPostBlockFCU sends a forkchoice update to the execution client after a
+// block is finalized.This function should only be used to notify
+// the EL client of the new head and should not request optimistic builds, as:
+// Optimistic clients already request builds in handleOptimisticPayloadBuild()
+// Non-optimistic clients should never request optimistic builds.
```
abi87, line 33: q: this function is still called in its own goroutine. Is this necessary? Can we call it synchronously? @nidhi-singh02  @shotes

beacon/blockchain/execution_engine.go
```diff
@@ -46,69 +47,9 @@ func (s *Service[
...
-	); err != nil {
-		s.logger.Error(
-			"failed to send forkchoice update with attributes in non-optimistic payload",
-			"error",
-			err,
-		)
-	}
+	// Send a forkchoice update without payload attributes to notify
+	// EL of the new head.
+	s.sendNextFCUWithoutAttributes(ctx, blk, lph)
```
abi87, line 52: consider moving `sendNextFCUWithoutAttributes` into this function to have a single function

abi87, line 52: another q: `sendNextFCUWithoutAttributes` can err, in which case we only log the error.
Should we return error instead and, pending the goroutine remova, err on block finalization if we fail to ping the EVM? @nidhi-singh02  @shotes

abi87 (commented):
nits and a couple of questions

beacon/blockchain/execution_engine.go
abi87: My understanding is that we should **not** err over non-deterministic errors like this in the finalize block context, just double-checking here

shotes: In the short term, I think this PR lgtm when it comes to fixing the initial issue.

But @abi87 brings up good questions that I think deserve a lot of attention and potentially some reworking - particularly in how beacon-kit handles staying in sync with its execution client. Here are some questions that I think will drive this discussion:
- What happens when `sendNextFCUWithoutAttributes()` fails because of some ephemeral issue in trying to send the forkchoiceUpdated RPC call? like a onetime networking error
- What happens when `sendNextFCUWithoutAttributes()` continues to fail over and over from network issues? or the EL client is syncing?
- What happens when `sendNextFCUWithoutAttributes()` fails because of an invalid parent block hash? Or an invalid forkchoice state? (I don't think this should *ever* happen in `FinalizeBlock()`, and if it does, there is clearly some chain-breaking bug somewhere.)

In my opinion, beacon-kit cannot simply move on with processing blocks if the forkchoiceUpdate is failing for any reason. At the same time, it is difficult to suggest returning error in `FinalizeBlock()`. A potential path forward would be to start the FCU in a goroutine that:
- retries on ephemeral error, with an exponential backoff that will timeout eventually
- errors when there is a true failure
- prevents the beacon-kit node from participating (not proposing, verifying proposals, or processing new blocks) until the forkchoice is updated on the EL.

abi87: @shotes fine with merging this PR and tackle the retry issues separately, but please make an issue so that we can better track this.
You extended comment may be the issue content.
Once that is done, I am down to merge this

shotes: New issue created in https://github.com/berachain/beacon-kit/issues/2267

nidhi-singh02: sure, moved!

PR #2243, opened 2024-12-11
Remove BeaconBlockHeaderT generic type

fridrik01:
I tried to rebase/merge https://github.com/berachain/beacon-kit/pull/2234 from @abi87 but it was just quicker to do from scratch, not sure if there was a better way.


da/blob/types.go
```diff
@@ -55,18 +55,18 @@ type BeaconBlockBody interface {
 	Length() uint64
 }
 
-type BeaconBlockHeader[BeaconBlockHeaderT any] interface {
+type BeaconBlockHeader interface {
```
abi87, line 58: @fridrik01 We should be able to remove this interface and just use the concrete type.

fridrik01: done!

PR #2244, opened 2024-12-11
fix(blockchain): remove reachable panic in blobsidecar processing

shotes:
The `BlockBodyKZGOffset()` function was written to panic on fork versions that aren't `version.Deneb`. This function is reachable by data from the `BlobSidecars` that have not yet been validated. A malicious proposer could include a `BlobSidecar` with a `slot` that would equate to a fork version that isn't `version.Deneb`, causing a panic on all nodes that process the `BlobSidecar`.

This PR fixes that issue by returning `ErrForkVersionNotSupported` instead.


abi87 (approved):
lgtm

Issue #841, opened 2024-04-22
Block headers in blob sidecars do not carry block signatures


sbudella-gco:
## Description

After the Deneb upgrade, validators of the beacon chain must obtain from
the execution clients supplementary data carried in new transactions
named "blob transactions". The blobs in the transaction are opaque to
the execution layer, and validators are in charge of handling them in
an exceptional way: blobs must be only referenced in beacon blocks but
not therein embedded. Indeed, the blobs are propagated in the consensus
layer network after the validator wraps them in further envelopes named
"sidecars". Blob sidecars must uniquely be linked to the block they were
processed and validated with.

In Beacon-kit, the blobs are fetched and processed at the time of the
block proposal by the builder service; first, a new block structure is
instantiated with `EmptyBeaconBlock`, then the local builder returns the
payload and the blobs bundle obtained from the execution engine by calling
`GetBestPayload`, as seen in `mod/runtime/services/builder/service.go`:

```
func (s *Service) RequestBestBlock(
        ctx context.Context,
        st state.BeaconState,
        slot primitives.Slot,
) (beacontypes.BeaconBlock, *datypes.BlobSidecars, error) {
[..]
        // Create a new empty block from the current state.
        blk, err := beacontypes.EmptyBeaconBlock( 
                slot,
                proposerIndex,
                parentBlockRoot,
                stateRoot,
                s.BeaconCfg().ActiveForkVersionForSlot(slot),
                reveal,
        ) 
        if err != nil { 
                return nil, nil, err 
        } else if blk == nil {
                return nil, nil, beacontypes.ErrNilBlk
        }

        latestExecutionPayload, err := st.GetLatestExecutionPayload()
        if err != nil {
                return nil, nil, err
        }
        parentEth1BlockHash := latestExecutionPayload.GetBlockHash()

        // Get the payload for the block.
        payload, blobsBundle, overrideBuilder, err := s.localBuilder.GetBestPayload(
                ctx,
                st,
                slot,
                parentBlockRoot,
                parentEth1BlockHash,
        )
        if err != nil {
                return blk, nil, fmt.Errorf(
                        "failed to get block root at index: %w",
                        err,
                )
        }
```

Afterwards, the sidecars are built and returned to the invoking `PrepareProposalHandler` handler from
`mod/runtime/abci/proposal.go`, which is the entrypoint to the CometBFT ABCI:

```
[..]
        blobSidecars, err := s.blobFactory.BuildSidecars(blk, blobsBundle)
[...]

PR #2245, opened 2024-12-12
Support for signed beacon block header and signature verification

sbudella-gco:
The modification addresses the long standing issue formerly reported [here](https://github.com/berachain/beacon-kit/issues/841), and revived [here](https://github.com/spearbit-audits/review-berachain-beaconkit-0919/issues/76). Blob sidecars were not strongly tied to the identity of the proposer of the block the sidecars are associated with.

The patch aligns with the [specification](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#signedbeaconblockheader) by introducing the `SignedBeaconBlockHeader` type, which embeds both the header of the block associated with the sidecar and a signature over it. A major divergence from the spec is that the signature is over the beacon block header as opposed to the whole block. But since the header contains the hash tree root of the block body, this should not pose any problem.


consensus-types/types/signed_header.go
```diff
@@ -0,0 +1,133 @@
...
+type SignedBeaconBlockHeader struct {
+	Header    *BeaconBlockHeader  `json:"header"`
+	Signature crypto.BLSSignature `json:"signature"`
+}
+
+/* -------------------------------------------------------------------------- */
+/*                                 Constructor                                */
+/* -------------------------------------------------------------------------- */
+
+// NewBeaconBlockHeader creates a new BeaconBlockHeader.
```
shotes, line 47:
```suggestion
// NewSignedBeaconBlockHeader creates a new BeaconBlockHeader.
```

da/blob/verifier.go
```diff
@@ -46,35 +47,53 @@ type verifier[
...
 
-	// check that sideracs block headers match with header of the
+	// check that sidecars block headers match with header of the
 	// corresponding block
-	for i, s := range sidecars.GetSidecars() {
-		if !s.GetBeaconBlockHeader().Equals(blkHeader) {
-			return fmt.Errorf("unequal block header: idx: %d", i)
+	for _, s := range sidecars.GetSidecars() {
+		var sigHeader = s.GetSignedBeaconBlockHeader()
+		err := verifierFn(
```
shotes, line 91: This removes any checks that validate that the `sigHeader` equals the `blkHeader`. This check could pass with any arbitrary values inside of `sigHeader` aside from the valid signature.

I believe we need to perform one of the two validations:
- ensure that `verifierFn(sigHeader, sigHeader.GetSignature()` also does not error
- ensure `sigHeader` == `blkHeader`

beacon/blockchain/pruning.go
```diff
@@ -38,7 +38,7 @@ func (s *Service[
 	}
 
 	// prune deposit store
-	start, end = depositPruneRangeFn(
+	start, end = depositPrunRangeFn(
```
shotes, line 41: is this from a different branch? intentional?

da/blob/verifier.go
shotes: Also note that the validation performed in this same function lower down can be removed after verification of equality of the two headers:
```go
	g.Go(func() error {
		return sidecars.ValidateBlockRoots()  // this can be removed
	})
```

shotes, line 91: Additionally, should we consider moving this into the goroutine wait groups below? the concurrent verification should be safe in this context, though not necessary.

sbudella-gco: Totally agree.

beacon/blockchain/pruning.go
sbudella-gco: It carried over from Polar's pub/sub cleanup branch, indeed.

state-transition/core/state_processor_sidecar_verifier.go
```diff
@@ -0,0 +1,80 @@
...
+	ctypes "github.com/berachain/beacon-kit/consensus-types/types"
+	"github.com/berachain/beacon-kit/primitives/common"
+	"github.com/berachain/beacon-kit/primitives/crypto"
+	"github.com/berachain/beacon-kit/primitives/version"
+)
+
+func (sp *StateProcessor[
+	_, _, BeaconStateT, _, _, _, _,
+	_, ForkDataT, _, _, _, _, _,
+]) GetSidecarVerifierFn(
```
abi87, line 33: q: I am not sure I understand why we need this function to depend on StateProcessor.
AFAIU, we need 'BeaconState', `ChainSpec` and `signed` in order to verify a sidecar, all elements that we could pass directly to the sidecar verifier.
Can we simplify the implementation, especially after we ditched the pub/sub system?

beacon/blockchain/process_proposal.go
```diff
@@ -83,13 +83,26 @@ func (s *Service[
...
-		c := convertConsensusSidecars[
+		cs := convertConsensusSidecars[
 			ConsensusSidecarsT,
 			BlobSidecarsT,
 		](consensusSidecars)
 
+		// Get the sidecar verification function from the state processor
+		//nolint:govet	// err shadowing
+		sidecarVerifierFn, err := s.stateProcessor.GetSidecarVerifierFn(
+			s.storageBackend.StateFromContext(ctx),
```
abi87, line 94: q: here we verify sidecar first then block.
Does this matter? Would this verification fail potentially if we swapped order?
My guess is yes, since the block processing may change fork for instance.
If so, consider commenting on this.

abi87: Another question: is this change backward compatible? Or should we introduce it past a certain fork.

da/types/sidecar.go
```diff
@@ -109,18 +108,19 @@ func (b *BlobSidecar) DefineSSZ(codec *ssz.Codec) {
 	ssz.DefineStaticBytes(codec, &b.Blob)
 	ssz.DefineStaticBytes(codec, &b.KzgCommitment)
 	ssz.DefineStaticBytes(codec, &b.KzgProof)
-	ssz.DefineStaticObject(codec, &b.BeaconBlockHeader)
+	ssz.DefineStaticObject(codec, &b.SignedBeaconBlockHeader)
```
abi87, line 111: q: doesn't this make the change non-backward compatible?

consensus-types/types/signed_header.go
```diff
@@ -0,0 +1,133 @@
...
+
+import (
+	"github.com/berachain/beacon-kit/primitives/common"
+	"github.com/berachain/beacon-kit/primitives/constraints"
+	"github.com/berachain/beacon-kit/primitives/crypto"
+	"github.com/karalabe/ssz"
+)
+
+// Total size: Header (112) + Signature (96).
+const SignedBeaconBlockHeaderSize = 208
```
abi87, line 31: this does not look used anywhere. Drop it

da/blob/factory.go
```diff
@@ -81,22 +87,46 @@ func (f *SidecarFactory[BeaconBlockT, _]) BuildSidecars(
...
 
 	startTime := time.Now()
 	defer f.metrics.measureBuildSidecarsDuration(
 		startTime, math.U64(numBlobs),
 	)
+
+	// Contrary to the spec, we do not need to sign the full
+	// block, because the header embeds the body's hash tree root
+	// already. We just need a bond between the block signer (already
+	// tied to CometBFT's ProposerAddress) and the sidecars.
```
abi87, line 102: q: can you clarify why we deviate from the specs? Granted this is safe, is there a reason why we should not be fully compliant here?

da/blob/verifier.go
```diff
@@ -92,10 +109,6 @@ func (bv *verifier[_, BlobSidecarsT]) verifySidecars(
 		return bv.verifyKZGProofs(sidecars)
 	})
 
-	g.Go(func() error {
-		return sidecars.ValidateBlockRoots()
```
abi87, line 96: q: why was this dropped?

abi87 (commented):
mostly questions and a nit

da/blob/factory.go
sbudella-gco: Spec-compliant implementations have the beacon block signed from the beginning, therein the beacon block factory always provide a signed envelope. In Berachain, we have never carried out full block signatures from within the app because the signature is done by CometBFT on a CometBFT block, not on the Beacon block (the CometBFT payload). Therefore, it is just a matter of opportunism to sign only the bare minimum (the block header) for the requirement.
Of course, we could sign the full block, given that the HTR of the body is already there in the header (and possible benefit from a tighter bond between the block and the proposer), but at this point the CometBFT signature becomes redundant (and a bloat).

da/blob/verifier.go
sbudella-gco: The function ensures that all sidecars point to the same beacon block header by the latter's HTR, which is something achieved already (among other guarantees) by the equality test in the loop above.

state-transition/core/state_processor_sidecar_verifier.go
sbudella-gco: The BeaconState exposed to the blockchain service is not the full one you'd access to through the state processor of the same service. The former implements the `ReadOnlyBeaconState` which at the moment lacks methods needed by the verifier, e.g. `GetGenesisValidatorsRoot`, etc. I understand we could always extend the interface but I think that would miss the point.

consensus-types/types/signed_header.go
sbudella-gco: Dropped.

beacon/blockchain/process_proposal.go
abi87: as discussed off-line, let's proceed with fixing this and handle differently backward compatibility issues, if any

abi87 (approved):
lgtm

PR #2248, opened 2024-12-12
chore(docs) added some docs around validation

abi87:


state-transition/core/README.md
```diff
@@ -2,12 +2,40 @@
...
 
-[^1]: Technically a validator is made in the BeaconKit state to track the deposit, but such a validator is never returned to the consensus engine.
+We list below a few relevant facts.
+
+## `Balance` and `EffectiveBalance`
+
+BeaconKit distingishes a validator `Balance` and a validator `EffectiveBalance`.
+
+- `Balance` is updated slot by slot, when a deposit in made over the deposit contract and events are subsequently processed by BeaconKit.
+- `Balance` changes in multiples of the `MinDeposit` specified in the deposit contract. There is no cap on the `Balance`.
```
calbera, line 18:
```suggestion
- `Balance` can increase only in multiples of `MinDepositAmount`, which is specified in the deposit contract. There is no cap on the `Balance`.
```

state-transition/core/README.md
```diff
@@ -2,12 +2,40 @@
...
+- `EffectiveBalance` is capped at `MaxEffectiveBalance`. Any `Balance` in excess of `MaxEffectiveBalance` is automatically withdrawn.
+- `EffectiveBalance` is aligned to `Balance`, within the constrains listed above, and accounting for hysteresys, at the turn of every epoch, via [`processEffectiveBalanceUpdates`](./state_processor.go#L491)
+
+## Validator lifecycle
+
+Validators are created by a deposit made to the deposit contract, as soon as the corresponding event is emitted.
+
+Say a deposit is made at slot `S`, epoch `N`, that creates a validator.
+
+The funds deposited will be locked and the validator will stay inactive until the a minimum staking balance is hit. The minimum staking balance is set to `EjectionBalance + EffectiveBalanceIncrement` to account for hysteresys.
```
calbera, line 30:
```suggestion
The funds deposited will be locked and the validator will stay inactive until the a minimum staking balance is hit. The minimum staking balance is set to `EjectionBalance + EffectiveBalanceIncrement` to account for hysteresys.  However if the active validator set has already reached the `ValidatorSetCap`, a new prospective validator must deposit more funds than the effective balance of the active validator with the lowest stake.
```

calbera (approved):

PR #2249, opened 2024-12-12
wip: Remove BeaconBlockBodyT

fridrik01:
NOTE: Unfinished and does not compile, parking here in case I can rebase later

More complex than I thought as it required removing:
- Eth1DataT
- DepositT
- ExecutionPayloadT

Seems also to require removing Withdrawal/s


fridrik01:
Replaced bar smaller incremental PRs, I now know at least what some of the errors were thanks to @abi87 (missing `any` in generic parameter list)

PR #2253, opened 2024-12-12
Fix potential goroutine variable reuse issue in BuildSidecars function

Kalpazanova53:
This commit resolves a potential issue where the variable `i`  used in the `range` loop could be reused across multiple  goroutines, leading to unexpected behavior.

The fix introduces a local `index` variable within each iteration  of the loop to ensure thread safety and prevent race conditions.

Additionally, this change improves the code's robustness and  aligns with best practices for goroutine handling.

### DIFF
--- a/da/blob/factory.go
+++ b/da/blob/factory.go
@@ -87,26 +87,27 @@ func (f *SidecarFactory[BeaconBlockT, _]) BuildSidecars(
 	defer f.metrics.measureBuildSidecarsDuration(
 		startTime, math.U64(numBlobs),
 	)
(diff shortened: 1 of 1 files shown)


shotes:
Variable reuse in loops did indeed used to be a problem. However, since go 1.22, the default behavior for loops changed to "loopvar" which implicitly creates a new variable for each loop iteration. In this case since this project is on go 1.23, each of those new variables are captured in their individual enclosures for the goroutines and this is totally safe.

abi87:
closing following @shotes comment

PR #2256, opened 2024-12-12
1/n: Remove ForkDataT generic type

fridrik01:


state-transition/core/state_processor_staking.go
```diff
@@ -169,7 +170,7 @@ func (sp *StateProcessor[
 	}
 
 	// Verify that the message was signed correctly.
-	var d ForkDataT
+	var d ctypes.ForkData
```
abi87, line 173: we can drop the `d.New` and use the NewForkData constructor here

abi87 (approved):
lgtm

fridrik01:
Please ignore the lint error, its fixed in https://github.com/berachain/beacon-kit/pull/2257

nidhi-singh02 (approved):

PR #2258, opened 2024-12-12
fix(blockchain): sequential blob `Persist()` DB writes

shotes:
The `store.Persist()` function stored `BlobSidecar` updates concurrently using `iter.Map()` from "github.com/sourcegraph/conc/iter" package. However, the underlying RangeDB is not equipped to handle the concurrent writes. This resulted in a race condition that could corrupt the DA store.

This PR fixes this issue and adds a `race` test case. Currently, this test case is disabled until the build flag "race" is added. Please advise on the best way to get that included in unit tests.

This race condition exists in the `db.firstNonNilIndex` field in `RangeDB`. There could also arise another concurrency issue if the same file is written to concurrently, which could happen if two separate blobsidecars in the same slot have the same KzgCommitment (which is currently not prevented).

I was originally considering making the `Set()` function concurrency safe. However, this could be very difficult do to the aforementioned concurrent writes to the same file. In addition, the sequential implementation is onlyl marginally slower than the concurrent implementation. Here are the differences for a test that calls `Persist()` 3 times with 20 blobs each:

    sequential: 259 iterations   4595262 ns/op
    iter.Map:   337 iterations   3631735 ns/op


da/store/store_test.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+func TestStore_PersistRace(t *testing.T) {
+	tmpFilePath := "/tmp/store_test"
+
+	// Make sure we start fresh
+	err := os.RemoveAll(tmpFilePath)
+	if err != nil {
+		t.Fatal(err)
+	}
```
abi87, line 24: do we need this if we have a deferred `RemoveAll`?

da/store/store_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// Remove db when we're done
+	defer os.RemoveAll(tmpFilePath)
+
+	logger := log.NewNopLogger()
+	chainSpec, err := spec.DevnetChainSpec()
+	if err != nil {
+		t.Fatal(err)
+	}
```
abi87, line 33: consider using `require.NoError(t, err)` instead of the if

abi87 (commented):
Two points:
- I would add `-race` flag for `make test-unit` command in `script\build\testing.mk` file. Nota that we do have `-race` enabled for a similar command, so I do not have objection is setting race detector in every test
- If having two separate blobsidecars in the same slot with the same KzgCommitment is illegal, I would add the check in this PR, as a way to fully handle the race. As the PR currently is, I believe we do not fully solve the race

abi87:
Another note: enabling race cause some linker warnings on macs, which should be benign (see https://github.com/golang/go/issues/61229)

shotes:
> * If having two separate blobsidecars in the same slot with the same KzgCommitment is illegal, I would add the check in this PR, as a way to fully handle the race. As the PR currently is, I believe we do not fully solve the race

Having two separate blobsidecars in the same slot with the same KzgCommitment is indeed illegal and is a unique problem to beacon-kit. That is a separate issue that I have halfway written a fix for this in a separate branch in response to an issue brought up in audit. But I have paused that fix so that I could wait for all of the generics stuff to be resolved first.

Having said all of that, even if there were two blobs with the same KzgCommitments currently, the fix in this PR would still prevent it from causing a race condition in the DB.

abi87 (approved):
lgtm

PR #2261, opened 2024-12-13
fix(state-transition): make `EffectiveBalance` update backward compatible

abi87:
in https://github.com/berachain/beacon-kit/pull/2142 we stopped updating EffectiveBalance every slot for existing validators. This is correct but must be done in a backward compatible manner


fridrik01 (approved):
LGTM!

PR #2266, opened 2024-12-15
Remove ExecutionAddressT generic type

fridrik01:
This commit moves the ChainSpec type definitions from primitives/common/consensus.go to chain-spec/chain/chain_spec.go in order to prevent cyclic dependencies.


chain-spec/chain/chain_spec.go
```diff
@@ -20,11 +20,45 @@
...
+	"github.com/berachain/beacon-kit/primitives/math"
+)
+
+type (
+	// Bytes32 defines the commonly used 32-byte array.
+	Bytes32 = bytes.B32
+
+	// ChainSpec defines an interface for chain-specific parameters.
+	//nolint:revive // We can remove this once we have cleaned up generics.
+	ChainSpec = Spec[DomainType, math.Epoch, math.Slot, any]
```
abi87, line 35: it seems to me this is the only type that needs to be moved to avoid import cycles. The others can be kept in common. See https://github.com/berachain/beacon-kit/pull/2279

PR #2271, opened 2024-12-16
Change dependabot to run once a month instead of every day

fridrik01:
It should be frequent enough and less spammy.


abi87:
I would not do this. I'd rather be inform promptly if some critical deps have been updated and then decide if and when to merge.
Not sure what @calbera thinks about this

fridrik01:
cleanup, can revisit if needed

PR #2272, opened 2024-12-16
2/n Remove ExecutionPayloadT generic type

fridrik01:
Had to move `requests.go` and `payload_env.go` to the consensus types package in order to address cyclic dependencies.

We may want to revisit a proper way to handle dependencies in case we don't want to end putting everything into a single package.


consensus-types/types/body.go
```diff
@@ -298,8 +298,7 @@ func (b *BeaconBlockBody) GetTopLevelRoots() []common.Root {
 		common.Root(b.GetGraffiti().HashTreeRoot()),
 		Deposits(b.GetDeposits()).HashTreeRoot(),
 		b.GetExecutionPayload().HashTreeRoot(),
-		// I think this is a bug.
-		common.Root{},
```
abi87, line 302: @fridrik01 why we removed the "I think this is a bug comment?" We should keep it

abi87 (approved):
lgtm with the caveat in the PR heading

PR #2280, opened 2024-12-17
feat(consensus): introduce predictable block times

melekes:
based on https://github.com/berachain/beacon-kit/pull/2221

## Assumptions

* [PBTS](https://docs.cometbft.com/v1.0/explanation/core/proposer-based-timestamps) is used, which can't be attacked by 1/3+ of malicious validators (time warp attack).
* `InitialTime` (genesis time) is persisted across restarts to preserve the desired block cadence.

## Algorithm

The algorithm is as follows (assuming the target block cadence (configurable) is 2s):

```
targetBlockTime = InitialTime + 2s * blockHeight

if actualBlockTime >= targetBlockTime {
  ABCI.FinalizeBlockResponse.NextBlockDelay = 0
} else {
  ABCI.FinalizeBlockResponse.NextBlockDelay = targetBlockTime - actualBlockTime
}
```

## Questions

1. Can I use `CommitMultiStore` to store `InitialTime`? CometBFT only submits genesis time if the app height is 0. If it's greater than 0, we must restore it from somewhere between restarts.
2. Are we okay with exposing a CLI flag to control block time? Could it be useful for testing?
3. Do we need the ability to change the desired block cadence (e.g, 2s)?


calbera:
Followup to questions:
1. Don't think it hurts, generally safer to have it stored on disk vs in memory.
2-3. I think it would be helpful to be configurable for testing. But at some point we'd have to enforce this in the binary so all nodes use the same value?

PR #2281, opened 2024-12-17
Fix BLS12-381 key generation

alesforz:
### Context
The code was incorrectly creating a _compressed_ 48-byte BLS12-381 key by truncating the _uncompressed_ 96-byte long key to 48 bytes. As a consequence, passing the resulting key as argument to BLS methods would return an error because the key was malformed.

### Changes
This PR adds code to create a compressed BLS12-381 using the `Compact` method of the `BLS12-381` library. 
It also calls the appropriate functions in the `cometbft/bls12381` pkg to correctly handle the keys.

Running `make start` with these changes works as expected, i.e., the "invalid BLS key" error does not show up anymore.


melekes (approved):
👍

PR #2274, opened 2024-12-16
4/n Remove BeaconBlockBodyT generic type

fridrik01:
Replaced with https://github.com/berachain/beacon-kit/pull/2282

PR #2282, opened 2024-12-17
4/n Remove BeaconBlockBodyT generic type 

fridrik01:
Instead of resolving all the conflicts in https://github.com/berachain/beacon-kit/pull/2274/  which were due to squashing the branches it was based on I instead created a new one.

I knew there would be no conflicts between `main` and `remove-beaconblockbodyt` but since previous merge squash commits removed all commit history git doesn't know how to deal with the conflicts so we must do it all again (which is super tedious when rebasing over 3 sets of large squashed commits). 

Therefore what I did is that I undid the commit in https://github.com/berachain/beacon-kit/pull/2274 stashed it and then checkout main and then popped the stash, no conflicts and much quicker

The set of operations for reference:
```
git checkout main
git pull
git checkout remove-beaconblockbodyt
git reset HEAD~
git stash
git checkout main
git checkout -b test-merge
git stash pop
git add .
git commit
git push origin test-merge
```

Overall when we work with stacked diffs like this we may not use merge using squashed commits, we need to enable that in this repo


nidhi-singh02 (approved):

state-transition/core/state_processor_genesis.go
```diff
@@ -71,7 +71,7 @@ func (sp *StateProcessor[
 	}
 
 	// TODO: we need to handle chain.Version vs uint32 better.
-	var blkBody BeaconBlockBodyT
+	var blkBody *ctypes.BeaconBlockBody
```
abi87, line 74: we may consolidate these kind of methods, but I guess it's fine to clean them up later on too

PR #2285, opened 2024-12-17
7/n Remove PayloadAttributesT generic type

fridrik01:
This PR also includes [fix: remove err check for nil to address nilaway false positive](https://github.com/berachain/beacon-kit/pull/2285/commits/d216294acee4da9148ce548b1e47d1d99d855468) as we had `nilaway` error. Looking at the potential panic it was never possible to trigger it so it must be a false positive.

### DIFF
--- a/execution/client/errors.go
+++ b/execution/client/errors.go
@@ -57,6 +57,11 @@ var (
 func (s *EngineClient) handleRPCError(
 	err error,
 ) error {
+	if err == nil {
+		//nolint:nilerr // appease nilaway
+		return err
+	}
+
 	// Check for timeout errors.
 	if http.IsTimeoutError(err) {
 		s.metrics.incrementHTTPTimeoutCounter()
--- a/execution/client/errors.go
+++ b/execution/client/errors.go
@@ -54,7 +54,7 @@ var (
 )
 
 // Handles errors received from the RPC server according to the specification.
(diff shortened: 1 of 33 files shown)


execution/client/errors.go
```diff
@@ -54,14 +54,9 @@ var (
 )
 
 // Handles errors received from the RPC server according to the specification.
-func (s *EngineClient[_]) handleRPCError(
+func (s *EngineClient) handleRPCError(
 	err error,
 ) error {
-	// Exit early if there is no error.
```
abi87, line 60: @fridrik01 wouldn't we want to keep this anyhow, as an early return optimization?

fridrik01: @abi87 In general I would be good with this, but it was triggering a `nilaway` false positive error (PR description). I am not sure why it would only happen in this PR and not before though

abi87: `nilaway` is saying that at some point we call `Error()` method over the err, which would panic if we return nil here. Instead all is fine if we return `err` rather than the literal `nil`.

abi87 (approved):
lgtm

execution/client/errors.go
fridrik01: Ahh, ok I see thanks!

PR #2286, opened 2024-12-17
chore(blobs): parallelize blobSidecar header checks

shotes:
Simply moving the header equality check into the same goroutine that verifies the signature and adding some comments.


abi87 (approved):
lgtm

PR #2288, opened 2024-12-17
8/n Remove BlobSidecarT and BlobSidecarsT generic type

fridrik01:


da/types/sidecars.go
```diff
@@ -27,43 +27,40 @@ import (
 	"github.com/sourcegraph/conc/iter"
 )
 
-// BlobSidecars is a slice of blob side cars to be included in the block.
-type BlobSidecars struct {
-	// Sidecars is a slice of blob side cars to be included in the block.
-	Sidecars []*BlobSidecar
-}
```
abi87, line 34: @fridrik01 I wonder if we should keep this as a struct. I expect this to change the way we serialize sidecars, so maybe we want to keep it as it was?

abi87 (commented):
I have a question around the possible change in serialization

da/types/sidecars.go
```diff
@@ -27,43 +27,40 @@ import (
...
 
 func (bs *BlobSidecars) Get(index int) *BlobSidecar {
-	return bs.Sidecars[index]
+	return (*bs)[index]
 }
 
 // IsNil checks to see if blobs are nil.
 func (bs *BlobSidecars) IsNil() bool {
-	return bs == nil || bs.Sidecars == nil
+	return bs == nil || len(*bs) == 0
```
shotes, line 52: This change in the `IsNil()` check changes the functionality of the function. In the old one, a `BlobSidecars` could get unmarshaled by SSZ and not be nil but have length of 0. With this change, 0 sidecars counts as nil.

I propose this change:
```suggestion
	return bs == nil
```

This will impact my other PR #2291 that uses `IsNil` to check if it's ok to dereference data, but explicitly allows an empty list (because there can be proposals that contain 0 sidecars). There is actually a new test case in that PR that tests this functionality explicitly:
```go
func TestZeroSidecarsInBlobSidecarsIsNotNil(t *testing.T) {
	// This test exists to ensure that proposing a BlobSidecars with 0
	// Sidecars is not considered IsNil().
	sidecars := &types.BlobSidecars{Sidecars: make([]*types.BlobSidecar, 0)}
	require.False(t, sidecars.IsNil())
}
```

fridrik01: Ah nice one!

PR #2291, opened 2024-12-18
fix(blobs): Validate sidecar `KzgCommitments` and disallow sidecar omission

shotes:
This PR fixes a few things:
- Ensures that the number of commitments in the BeaconBlock and the number of BlobSidecars are equal.
- Ensures that BlobSidecars do not contain duplicate BlobSidecars
- (this implicitly verifies that the KzgCommitments from the BeaconBlock are 1 to 1 with BlobSidecars due to the inclusion proof)


calbera (approved):
makes sense to me

beacon/blockchain/process_proposal.go
```diff
@@ -72,56 +78,47 @@ func (s *Service[
...
 
-	// In theory, swapping the order of verification between the sidecars
-	// and the incoming block should not introduce any inconsistencies
-	// in the state on which the sidecar verification depends on (notably
-	// the currently active fork). ProcessProposal should only
-	// keep the state changes as candidates (which is what we do in
-	// VerifyIncomingBlock).
+	// Make sure we have the right number of BlobSidecars
+	numCommitments := len(blk.GetBody().GetBlobKzgCommitments())
+	if numCommitments != sidecars.Len() {
```
calbera, line 90: although we already check before calling `.Len()` if `sidecars.IsNil()` I think we should prevent any `sidecars` function from ever panicking on its own

calbera: i.e. here [Len, GetSidecars, Get](https://github.com/berachain/beacon-kit/pull/2291/files#diff-29cacb221d7916f7d9d44e53bc63db6dc41f5e5f816f9634eb5faf5808e98417R38-R48) should all check .IsNil before dereferencing the pointer.

abi87: we do in line 81 right?

abi87 (approved):
lgtm

PR #2296, opened 2024-12-18
feat(state-transition): add deposits root to beacon state

abi87:
NOTE: this change is state breaking with respect to bartio and boonet

Introduce deposit validation in state.

- Include merkle root for all deposits in block.Eth1Data.DepositRoot
- recheck block.Eth1Data.DepositRoot against deposit root recalculated from local state, accounting for deposits coming in the block
- drop deposits pruning to preserve all deposits for the state verification

### DIFF
--- a/consensus-types/types/state.go
+++ b/consensus-types/types/state.go
@@ -60,7 +60,7 @@ type BeaconState struct {
 	TotalSlashing math.Gwei
 
 	// Deposits
-	BlockDepositRoot common.Root
+	DepositsRoot common.Root
 }
 
 // New creates a new BeaconState.
@@ -82,7 +82,7 @@ func (st *BeaconState) New(
 	nextWithdrawalValidatorIndex math.ValidatorIndex,
 	slashings []math.Gwei,
 	totalSlashing math.Gwei,
-	blockDepositRoot common.Root,
+	depositsRoot common.Root,
 ) (*BeaconState, error) {
 	return &BeaconState{
 		Slot:                         slot,
@@ -101,7 +101,7 @@ func (st *BeaconState) New(
 		NextWithdrawalValidatorIndex: nextWithdrawalValidatorIndex,
 		Slashings:                    slashings,
 		TotalSlashing:                totalSlashing,
-		BlockDepositRoot:             blockDepositRoot,
+		DepositsRoot:                 depositsRoot,
 	}, nil
 }
 
--- a/consensus-types/types/state.go
+++ b/consensus-types/types/state.go
@@ -58,9 +58,6 @@ type BeaconState struct {
 	// Slashing
 	Slashings     []math.Gwei
 	TotalSlashing math.Gwei
-
-	// Deposits
-	DepositsRoot common.Root
 }
 
 // New creates a new BeaconState.
@@ -82,7 +79,6 @@ func (st *BeaconState) New(
 	nextWithdrawalValidatorIndex math.ValidatorIndex,
 	slashings []math.Gwei,
 	totalSlashing math.Gwei,
-	depositsRoot common.Root,
 ) (*BeaconState, error) {
 	return &BeaconState{
 		Slot:                         slot,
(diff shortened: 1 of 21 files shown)


state-transition/core/state/statedb.go
```diff
@@ -389,6 +389,11 @@ func (s *StateDB[BeaconStateMarshallableT, _]) GetMarshallable() (BeaconStateMar
 		return empty, err
 	}
 
+	deposits, err := s.GetDeposits()
```
calbera, line 392: should return the deposits of the last processed block

state-transition/core/state_processor_staking.go
```diff
@@ -60,6 +60,10 @@ func (sp *StateProcessor[
 			return err
 		}
 	}
+
+	// Add deposits to beaconState
+	st.SetDepositRoot(deposits)
```
calbera, line 65: UpdateDepositRoot AND set block deposits in the state

consensus-types/types/state.go
```diff
@@ -58,6 +58,9 @@ type BeaconState struct {
 	// Slashing
 	Slashings     []math.Gwei
 	TotalSlashing math.Gwei
+
+	// Deposits
+	DepositsRoot common.Root
```
sbudella-gco, line 63: This is not needed if we stick to `Eth1Data`

consensus-types/types/state.go
```diff
@@ -97,6 +101,7 @@ func (st *BeaconState) New(
 		NextWithdrawalValidatorIndex: nextWithdrawalValidatorIndex,
 		Slashings:                    slashings,
 		TotalSlashing:                totalSlashing,
+		DepositsRoot:                 depositsRoot,
```
sbudella-gco, line 104: same as above.

consensus-types/types/state.go
abi87: Dropped

consensus-types/types/state.go
abi87: Dropped

calbera (commented):
In theory, once changing the `maxItems` in `Deposits` HashTreeRoot this seems correct to me/

state-transition/core/core_test.go
```diff
@@ -227,3 +227,10 @@ func buildNextBlock(
 		Body:          nextBlkBody,
 	}
 }
+
+// We verify DepositRoot in Eth1Data, but we don't really update
+// the store once we store deposits (it's not StateProcessor responsibility).
+// So we just keep using the dummyEth1Data most of the time.
+var dummyEth1Data = &types.Eth1Data{
```
sbudella-gco, line 234: no longer needed

calbera: addressed in https://github.com/berachain/beacon-kit/pull/2304

shotes:
Since we are calling `GetDepositsByIndex` on the whole list of deposits, we may want to consider optimizing that function by pre-allocating [`deposits = []*ctypes.Deposit{}`](https://github.com/berachain/beacon-kit/blob/c08698ebbc62b5f6496ed400622978e10adbd2d2/storage/deposit/store.go#L83) to a length of `endIdx - startIndex` instead of appending a new deposit and causing a bunch of reallocation.

Which brings me to another question regarding that same function `GetDepositsByIndex` - why are we returning no error when a deposit index is not found?

shotes:
The genesis deposits don't seem to be added to the `DepositStore` even though the `Eth1DepositIndex` gets incremented during genesis. This means that `GetDepositsByIndex(0, depositIndex)` would return an empty list, and all deposit roots would not include any of the previous deposits outside of the current new set.

calbera (approved):

primitives/constants/payload.go
```diff
@@ -31,8 +31,12 @@ const (
 	// payload.
 	MaxTxsPerPayload uint64 = 1048576
 
-	// MaxDepositsPerBlock is the maximum number of deposits per block.
-	MaxDepositsPerBlock uint64 = 16
+	// DepositContractDepth is the depth of the deposit contract merkle tree.
+	DepositContractDepth uint64 = 32
```
abi87, line 35: q: @calbera does this require changes to the deposit contract?

PR #2299, opened 2024-12-19
feat(state): include root of deposits in beacon state

calbera:
**NOTE: this change is state breaking with respect to bartio and boonet**

### Summary
- set the combi(prevDepositRoot, HTR of block deposits) in the Eth1Data.DepositRoot field of BeaconState
- now verify that the local deposits root is equivalent to the one in Eth1Data
- removes the need to check full equality of individual deposits
- removes special casing around `Eth1DepositIndex`. It is now set as in spec: increment after processing each deposit.
- keeps pruning on for deposit store and prevents O(n^2) ? complexity growth of hashing historical deposits (n being # of all-time deposits)

#### Questions
- do we need to check the derived genesis deposits root on CL against the EL deposit contract? Can be done [here](https://github.com/berachain/beacon-kit/blob/deposits-in-state/beacon/blockchain/init_chain.go#L34).

### DIFF
--- a/beacon/blockchain/deposit.go
+++ b/beacon/blockchain/deposit.go
@@ -39,7 +39,7 @@ func (s *Service[
 	ctx context.Context,
 	blockNum math.U64,
 ) {
-	if blockNum < s.eth1FollowDistance {
+	if blockNum <= s.eth1FollowDistance {
 		s.logger.Info(
 			"depositFetcher, nothing to fetch",
 			"block num", blockNum,
--- a/beacon/validator/block_builder.go
+++ b/beacon/validator/block_builder.go
@@ -310,6 +310,10 @@ func (s *Service[
 	if err != nil {
 		return ErrNilDepositIndexStart
 	}
+	eth1Data, err := st.GetEth1Data()
+	if err != nil {
+		return err
+	}
 
 	deposits, err := s.sb.DepositStore().GetDepositsByIndex(
 		depositIndex, s.chainSpec.MaxDepositsPerBlock(),
@@ -325,9 +329,8 @@ func (s *Service[
 	)
 	body.SetDeposits(deposits)
 
-	var eth1Data *ctypes.Eth1Data
 	body.SetEth1Data(eth1Data.New(
-		ctypes.Deposits(deposits).HashTreeRoot(),
+		ctypes.Deposits(deposits).CombiHashTreeRoot(eth1Data.DepositRoot),
 		0,                      // Eth1 Deposit Count is not needed.
 		common.ExecutionHash{}, // Eth1 Block Hash is not needed.
 	))
--- a/beacon/validator/block_builder.go
+++ b/beacon/validator/block_builder.go
@@ -26,7 +26,6 @@ import (
 	"time"
 
 	payloadtime "github.com/berachain/beacon-kit/beacon/payload-time"
-	"github.com/berachain/beacon-kit/config/spec"
 	ctypes "github.com/berachain/beacon-kit/consensus-types/types"
 	"github.com/berachain/beacon-kit/consensus/types"
 	"github.com/berachain/beacon-kit/primitives/bytes"
@@ -312,16 +311,8 @@ func (s *Service[
 		return ErrNilDepositIndexStart
 	}
 
-	// Bartio and Boonet pre Fork2 have deposit broken and undervalidated
-	// Any other network should build deposits the right way
-	if !(s.chainSpec.DepositEth1ChainID() == spec.BartioChainID ||
-		(s.chainSpec.DepositEth1ChainID() == spec.BoonetEth1ChainID &&
-			blk.GetSlot() < math.U64(spec.BoonetFork2Height))) {
-		depositIndex++
-	}
 	deposits, err := s.sb.DepositStore().GetDepositsByIndex(
-		depositIndex,
-		s.chainSpec.MaxDepositsPerBlock(),
+		depositIndex, s.chainSpec.MaxDepositsPerBlock(),
 	)
 	if err != nil {
 		return err
--- a/beacon/validator/block_builder.go
+++ b/beacon/validator/block_builder.go
@@ -335,11 +335,10 @@ func (s *Service[
 	body.SetDeposits(deposits)
 
 	var eth1Data *ctypes.Eth1Data
-	// TODO: assemble real eth1data.
 	body.SetEth1Data(eth1Data.New(
-		common.Root{},
-		0,
-		common.ExecutionHash{},
+		ctypes.Deposits(deposits).HashTreeRoot(),
+		0,                      // Eth1 Deposit Count is not needed.
+		common.ExecutionHash{}, // Eth1 Block Hash is not needed.
 	))
 
 	// Set the graffiti on the block body.
--- a/scripts/build/golines.sh
+++ b/scripts/build/golines.sh
@@ -42,9 +42,9 @@ else
 fi
 
 # Find all .go files in the project directory and its subdirectories, ignoring .pb.go and .pb_encoding.go files
-find "${ROOT_DIR}" -type f -name "*.go" ! -name "*.pb.go" ! -name "*.pb_encoding.go" | while read -r file; do
+find "${ROOT_DIR}" -type f -name "*.go" ! -name "*.pb.go" ! -name "*_test.go" ! -name "*.pb_encoding.go" | while read -r file; do
     echo "Processing $file..."
-    golines --reformat-tags --shorten-comments --write-output --max-len=80 "$file"
+    golines --reformat-tags --shorten-comments --write-output --max-len=140 "$file"
 done
 
 echo "✅ All files processed."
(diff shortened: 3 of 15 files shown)


testing/e2e/e2e_staking_test.go
```diff
@@ -152,16 +152,17 @@ func (s *BeaconKitE2ESuite) TestDepositRobustness() {
...
 	depositAmt, _ := big.NewFloat(DepositAmount).Int(nil)
-	totalAmt := new(big.Int).Mul(depositAmt, big.NewInt(NumDepositsLoad))
-	upperBound := new(big.Int).Add(totalAmt, twoEther)
+	lowerBound := new(big.Int).Mul(depositAmt, big.NewInt(NumDepositsLoad))
+	upperBound := new(big.Int).Add(lowerBound, twoEther)
 	amtSpent := new(big.Int).Sub(balance, postDepositBalance)
 
-	s.Require().Equal(amtSpent.Cmp(totalAmt), 1)
-	s.Require().Equal(amtSpent.Cmp(upperBound), -1)
+	s.Logger().Info("balances", "amtSpent", amtSpent.String())
```
calbera, line 163:
```suggestion
```

scripts/build/golines.sh
```diff
@@ -42,9 +42,9 @@ else
 fi
 
 # Find all .go files in the project directory and its subdirectories, ignoring .pb.go and .pb_encoding.go files
-find "${ROOT_DIR}" -type f -name "*.go" ! -name "*.pb.go" ! -name "*.pb_encoding.go" | while read -r file; do
+find "${ROOT_DIR}" -type f -name "*.go" ! -name "*.pb.go" ! -name "*_test.go" ! -name "*.pb_encoding.go" | while read -r file; do
```
abi87, line 45: q: why do we skip tests here?

state-transition/core/state_processor_genesis.go
```diff
@@ -61,9 +61,9 @@ func (sp *StateProcessor[
 
 	var eth1Data *ctypes.Eth1Data
 	eth1Data = eth1Data.New(
-		common.Root{},
-		0,
-		execPayloadHeader.GetBlockHash(),
+		ctypes.Deposits(deposits).HashTreeRoot(),
+		0,                      // Eth1 Deposit Count is not needed.
+		common.ExecutionHash{}, // Eth1 Block Hash is not needed.
```
abi87, line 66: q: is the genesis execution payload header not part of the beacon state anymore with this? If so I believe we should restore this

state-transition/core/state_processor_genesis_test.go
```diff
@@ -152,7 +152,7 @@ func TestInitialize(t *testing.T) {
 	// deposit index is set to the last accepted deposit.
 	latestValIdx, err := st.GetEth1DepositIndex()
 	require.NoError(t, err)
-	require.Equal(t, uint64(len(genDeposits)-1), latestValIdx)
+	require.Equal(t, uint64(len(genDeposits)), latestValIdx)
```
abi87, line 155: it looks like I made the wrong fix for this index. Can you attach here the spec which tells how the index should be set?

state-transition/core/state_processor_staking.go
```diff
@@ -52,6 +52,9 @@ func (sp *StateProcessor[
 			sp.cs.MaxDepositsPerBlock(), len(deposits),
 		)
 	}
+	if err := st.SetEth1Data(blk.GetBody().GetEth1Data()); err != nil {
+		return err
+	}
```
sbudella-gco, line 57: `Eth1Data` here is tied to the deposits coming with the block. We should set it in the state only after the deposit validation (and processing), see [here](https://github.com/berachain/beacon-kit/blob/d43b1bd7516a2ff656cff8f74391f03ebb91afef/state-transition/core/state_processor_staking.go#L64-L65)

state-transition/core/validation_deposits.go
```diff
@@ -154,27 +154,19 @@ func (sp *StateProcessor[
...
-			if !sd.Equals(deposits[i]) {
-				return errors.Wrapf(
-					ErrDepositMismatch,
-					"local deposit: %+v, payload deposit: %+v",
-					sd, deposits[i],
-				)
-			}
+		// Verify that the local deposits have the same root as the block deposits.
+		var eth1Data *ctypes.Eth1Data
+		eth1Data, err = st.GetEth1Data()
```
sbudella-gco, line 159: See my comment above. The `Eth1Data` here was set already to what comes in the block, and it will stay so even if the comparison below fails.

beacon/validator/block_builder.go
```diff
@@ -335,11 +326,10 @@ func (s *Service[
 	body.SetDeposits(deposits)
 
 	var eth1Data *ctypes.Eth1Data
-	// TODO: assemble real eth1data.
 	body.SetEth1Data(eth1Data.New(
-		common.Root{},
-		0,
-		common.ExecutionHash{},
+		ctypes.Deposits(deposits).HashTreeRoot(),
```
sbudella-gco, line 330: You set the `DepositRoot` to the hash of the terminal segment of the list of deposits, but you are not linking it to the hash of the past deposits, i.e. the previous `Eth1Data`.

scripts/build/golines.sh
calbera: golines is formatting lines and comments under the specified line length. But certain tests have byte arrays or literal values that we don’t want to truncate for sake of formatting

state-transition/core/state_processor_genesis.go
calbera: I thought about this but only wanted to make the minimal changes here. Since the deposit count is implicitly guaranteed and the execution block hash is also part of execution payload header already, CL already agrees on these data and we don’t need to explicitly set here

state-transition/core/state_processor_genesis_test.go
calbera: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#deposits 
```python
def process_deposit(state: BeaconState, deposit: Deposit) -> None:
    # Verify the Merkle branch
    assert is_valid_merkle_branch(
        leaf=hash_tree_root(deposit.data),
        branch=deposit.proof,
        depth=DEPOSIT_CONTRACT_TREE_DEPTH + 1,  # Add 1 for the List length mix-in
        index=state.eth1_deposit_index,
        root=state.eth1_data.deposit_root,
    )

    # Deposits must be processed in order
    state.eth1_deposit_index += 1

    apply_deposit(
        state=state,
        pubkey=deposit.data.pubkey,
        withdrawal_credentials=deposit.data.withdrawal_credentials,
        amount=deposit.data.amount,
        signature=deposit.data.signature,
    )
```

state-transition/core/state_processor_staking.go
calbera: sure

beacon/validator/block_builder.go
calbera: Now using combi(historicalRoot, blkDeposits.HTR)

beacon/blockchain/deposit.go
```diff
@@ -39,7 +39,7 @@ func (s *Service[
 	ctx context.Context,
 	blockNum math.U64,
 ) {
-	if blockNum < s.eth1FollowDistance {
+	if blockNum <= s.eth1FollowDistance {
```
abi87, line 42: q: why?

storage/beacondb/eth1.go
```diff
@@ -51,8 +55,13 @@ func (kv *KVStore) SetLatestExecutionPayloadHeader(
 }
 
 // GetEth1DepositIndex retrieves the eth1 deposit index from the beacon state.
+// If the index is not found, it returns 0 and no error.
 func (kv *KVStore) GetEth1DepositIndex() (uint64, error) {
-	return kv.eth1DepositIndex.Get(kv.ctx)
+	index, err := kv.eth1DepositIndex.Get(kv.ctx)
+	if errors.Is(err, collections.ErrNotFound) {
+		return 0, nil
+	}
```
abi87, line 63: I think this should be not necessary if we duly set the index for genesis validators?

calbera: but this is checked in validateGenesisDeposits and right now expects an error. If the error is something other than ErrNotFound we will continue processing even if something else is wrong.

state-transition/core/validation_deposits.go
```diff
@@ -56,9 +56,11 @@ func (sp *StateProcessor[
...
 
 	default:
-		// TODO: improve error handling by distinguishing
-		// ErrNotFound from other kind of errors
-		if _, err := st.GetEth1DepositIndex(); err == nil {
+		index, err := st.GetEth1DepositIndex()
+		if err != nil {
+			return err
+		}
+		if index != 0 {
```
abi87, line 63: this can be fixed by duly initializing the index at genesis

calbera:
closing in favor of #2296 to maintain deposit merkle tree compat

PR #2300, opened 2024-12-19
Fix `go.mod` dependency graph

alesforz:
### Context
The project was defining dependencies in `go.mod` manually, that is, it used relative paths to local copies of the cosmos-sdk modules in the `go.mod` to tell Go where to find the packages on disk rather than downloading them from the internet. This temporary solution was motivated by continuous errors returned by `go mod tidy` when trying to import the necessary modules.

We believe that use of the module `cosmossdk.io/store/v2` was the culprit behind the `go mod tidy` chain of errors. According to the cosmos-sdk [Version Matrix](https://github.com/cosmos/cosmos-sdk?tab=readme-ov-file#version-matrix), `store/v2` is a module available only in versions of the sdk `>= v2`. Given that this project uses the sdk `v0.52.0` the use of this package isn't permitted by the sdk. I believe trying to import it is what caused a chain of incorrect module imports during the `go mod tidy` process.

### Changes
This PR does the following:
- updates the cosmos-sdk to `v0.52.0-rc.1`
- removes the relative paths to local deps; points the `go.mod` to the actual packages.
- drops `cosmossdk.io/store/v2` as a dependency. After inspecting the code, we noticed that there was only [one place](https://github.com/berachain/beacon-kit/blob/b1ca46401bae2004890d7c04876e4525212fb81e/node-core/components/deposit_store.go#L52) in the entire codebase in which the project used `store/v2`: to create a pebble instance to pass to `depositStore.NewStore(...)`. We substituted `store/v2` with the `github.com/cosmos/cosmos-db` module, which also offers a pebble implementation and whose types are compatible with the interfaces that the project uses.


melekes (approved):
🚀

PR #2302, opened 2024-12-19
Fix invalid BLS key length error

alesforz:
### Context
While `make start` duly compiled and ran` beacond`, there was an error when attaching an execution layer.  For example, running 1make start-reth` would return: 
```
Error: error during handshake: error on replay: encoding: invalid key length for bls12_381, got 48, want 96
```
This is an error returned by cometbft.

### Changes
This PR updates `go.mod` to use the latest cometbft version that addresses the above error. For details of the change in cometbft, see this PR https://github.com/cometbft/cometbft/pull/4692

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,8 +3,8 @@ module github.com/berachain/beacon-kit
 go 1.23.4
 
 replace (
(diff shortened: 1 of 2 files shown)


fridrik01 (approved):
👍

PR #2304, opened 2024-12-19
chore(state): cleanup deposits special casing

calbera:
- Cleanup special casing used for processing deposits
- Correctly set `Eth1DepositIndex` in state according to [spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#deposits) `process_deposit`
- Minor performance nits

### DIFF
--- a/beacon/validator/block_builder.go
+++ b/beacon/validator/block_builder.go
@@ -319,7 +319,7 @@ func (s *Service[
 			blk.GetSlot() < math.U64(spec.BoonetFork2Height))) {
 		depositIndex++
 	}
-	deposits, err := s.sb.DepositStore().GetDepositsByIndex(
+	blkDeposits, err := s.sb.DepositStore().GetDepositsByIndex(
 		depositIndex,
 		s.chainSpec.MaxDepositsPerBlock(),
 	)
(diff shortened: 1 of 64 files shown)


shotes:
While we're at it, should we change all instances of `[]*Deposit` or `[]*ctypes.Deposit` to `ctypes.Deposits`?

beacon/validator/block_builder.go
```diff
@@ -312,45 +311,23 @@ func (s *Service[
...
 	)
 	if err != nil {
 		return err
 	}
 
-	// Set the deposits on the block body.
+	var eth1Data *ctypes.Eth1Data
+	body.SetEth1Data(eth1Data.New(deposits.HashTreeRoot(), 0, common.ExecutionHash{}))
+
+	// Set the just the block deposits (after current index) on the block body.
```
shotes, line 325: nit
```suggestion
	// Set just the block deposits (after current index) on the block body.
```

calbera:
> While we're at it, should we change all instances of `[]*Deposit` or `[]*ctypes.Deposit` to `ctypes.Deposits`?

only changed it whereever the .HashTreeRoot() function is needed

beacon/validator/block_builder.go
```diff
@@ -312,45 +311,23 @@ func (s *Service[
...
-	}
-	deposits = append(deposits, blkDeposits...)
-
-	var eth1Data *ctypes.Eth1Data
-	body.SetEth1Data(eth1Data.New(
-		ctypes.Deposits(deposits).HashTreeRoot(),
-		0,
-		common.ExecutionHash{},
-	))
+	body.SetDeposits(deposits[depositIndex:])
```
shotes, line 330: `GetDepositsByIndex` could theoretically return a slice of length less than `depositIndex`, for example if there is a missing index anywhere in the depositStore (which probably shouldn't happen, but it *possibly could*). There should be an index check before indexing into this to prevent panic in this special case.

calbera: done

PR #2312, opened 2024-12-20
fix: timeout_commit defaults to 0 in init

fridrik01:
Now that we have upgraded `cosmossdk`/`cometbft` dependencies it becomes clear that we should improve the way we generate our config files.

In newest `cometbft` version it set the default `timeout_commit` to 0 seconds instead of 1s it was before which which causes us to produce blocks every 30-50us.

To address this quickly I implemented the `InitCmd` in our repo instead of using its implementation in `cosmossdk`. Its almost the same except I check if `timeout_commit` is 0 and if so set it to 1000ms. It would be better to read the config from file but it seems it has special encoding by viper so not easy to unmarshal (I may post an update or new PR as a followup)

But we should definitely fix the underlying issue of not using proper defaults specific to our application.

### DIFF
--- a/cli/commands/initialize/initialize.go
+++ b/cli/commands/initialize/initialize.go
@@ -102,7 +102,11 @@ func InitCmd(mm interface {
 			clientCtx := client.GetClientContextFromCmd(cmd)
 
 			config := client.GetConfigFromCmd(cmd)
-			chainID, _ := cmd.Flags().GetString(flags.FlagChainID)
+			chainID, err := cmd.Flags().GetString(flags.FlagChainID)
+			if err != nil {
+				return errors.New("failed to parse FlagChainID")
+			}
+
 			switch {
 			case chainID != "":
 			case clientCtx.ChainID != "":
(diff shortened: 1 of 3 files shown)


calbera:
Just noting that the `init` command here is being included as a direct alternative to https://github.com/cosmos/cosmos-sdk/blob/main/x/genutil/client/cli/init.go

cli/commands/initialize/initialize.go
```diff
@@ -0,0 +1,239 @@
...
+// LICENSOR OR ITS AFFILIATES (PROVIDED THAT YOU MAY USE A TRADEMARK OR LOGO OF
+// LICENSOR AS EXPRESSLY REQUIRED BY THIS LICENSE).
+//
+// TO THE EXTENT PERMITTED BY APPLICABLE LAW, THE LICENSED WORK IS PROVIDED ON
+// AN “AS IS” BASIS. LICENSOR HEREBY DISCLAIMS ALL WARRANTIES AND CONDITIONS,
+// EXPRESS OR IMPLIED, INCLUDING (WITHOUT LIMITATION) WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, NON-INFRINGEMENT, AND
+// TITLE.
+
+// implementation taken from github.com/cosmos/cosmos-sdk/blob/main/x/genutil/client/cli/init.go
```
calbera, line 21: can we specify exactly what we do differently in a comment here?

fridrik01: done!

abi87 (approved):
lgtm

PR #2315, opened 2024-12-20
chore(test): test setup for a local network with multiple nodes

nidhi-singh02:
Steps to run the network are mentioned in `testing/play_e2e/README.md`

The network won't be deleted on its own, have to delete it manually following the README file.


nidhi-singh02:
Not needed, closing it.

PR #2322, opened 2024-12-24
refactor(engine): improve Engine.Start error handling

lukema95:
## Description

This PR improves the error handling in the `Engine.Start` method. The current implementation uses `panic` to handle errors, which could lead to program crashes and makes error handling difficult for upstream callers.

### Changes

- Replace `panic` with proper error handling using channels
- Add context cancellation handling
- Improve error logging and observability
- Make startup process more reliable and controllable

### DIFF
--- a/execution/engine/engine.go
+++ b/execution/engine/engine.go
@@ -22,6 +22,7 @@ package engine
 
 import (
 	"context"
+	"fmt"
 
 	ctypes "github.com/berachain/beacon-kit/consensus-types/types"
 	engineprimitives "github.com/berachain/beacon-kit/engine-primitives/engine-primitives"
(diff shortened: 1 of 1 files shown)


execution/engine/engine.go
```diff
@@ -58,17 +59,23 @@ func New(
...
+					ee.logger.Error("engine client failed to start", "error", err)
+					errChan <- err
+					return
+			}
+			close(errChan)
 	}()
-	return nil
+
+	select {
+	case err := <-errChan:
```
melekes, line 74: This changes the behavior of `Start`. Previously, we'd exit as soon as goroutine was spawned. Now we're waiting for `Start` to finish (blocking select). What's the point of having a goroutine then?

calbera:
outdated, thanks though!

PR #2323, opened 2024-12-24
fix: Only use reth and geth

MaanavKhaitan:


MaanavKhaitan:
sorry made a PR on the wrong repo, closing

PR #2327, opened 2024-12-28
chore(docs) incorrect usage of `slices.Collect` function

Anon-im:
The `maps.Keys` function returns a slice of the map's keys, so the additional call to `slices.Collect` is redundant and will result in a compilation error: `undefined: slices.Collect`. To obtain the list of keys, it is sufficient to use only `maps.Keys.`


Anon-im:
Hey! @itsdevbear @abi87 my PR is ready for review! Thanks!

abi87:
`maps.Keys` returns an iterator not a slice, hence we need slices.Collect

PR #2329, opened 2024-12-28
Fix incorrect method call on a nil pointer

intls:
In line 171, the `NewWithVersion` method is called on the variable `blk`, which is declared as a pointer to `ctypes.BeaconBlock` and initially set to nil `(var blk *ctypes.BeaconBlock)`. Attempting to call a method on a `nil `pointer will cause a runtime panic.

***The program will crash when attempting to call a method on a nil pointer, which will disrupt the functionality of the entire system.***


abi87:
This is simply not true

PR #2332, opened 2024-12-31
feat: make chainSpec values height dependent

### LINKED ISSUE #2193: Make chainSpecs height dependent
This would allow changing them safely via a hard fork

crStiv:
This PR implements height-dependent chain specification parameters to allow safe changes via hard forks.

Changes:
- Added HeightDependentParams structure
- Implemented height tracking mechanism
- Modified chainSpec to support height-dependent parameters
- Updated parameter getters to use height-dependent values
- Added height parameter updates in FinalizeBlock
- Implemented concrete parameter values for Boonet hardforks

This resolves #2193

### DIFF
--- a/chain-spec/chain/chain_spec.go
+++ b/chain-spec/chain/chain_spec.go
@@ -20,11 +20,21 @@
 
 package chain
 
+import (
+	"github.com/berachain/beacon-kit/primitives/common"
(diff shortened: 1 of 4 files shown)


abi87:
we picked a different way to implement this. Outdated

PR #2333, opened 2024-12-31
chore(boonet): min stake upgrade on boonet fork 4

calbera:
`boonet` branch is currently synced with `v0.6.1`. After merging this PR, future releases for the boonet network can be cut off of the `boonet` branch. 

Effective changes are to validator set processing:
- The min stake (or `EjectionBalance`) is now set to 250k BERA 
- The `EffectiveBalanceIncrement` (not the same as minimum deposit amount on the execution layer deposit contract) is now set to 10k BERA

### DIFF
--- a/state-transition/core/state_processor_genesis.go
+++ b/state-transition/core/state_processor_genesis.go
@@ -179,7 +179,7 @@ func (sp *StateProcessor[
 			)
 		}
 		minEffectiveBalance := math.Gwei(
-			sp.cs.EjectionBalance() + sp.cs.EffectiveBalanceIncrement(),
+			sp.cs.EjectionBalance(false) + sp.cs.EffectiveBalanceIncrement(),
 		)
 
 		var idx math.ValidatorIndex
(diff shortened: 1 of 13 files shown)


state-transition/core/state_processor_genesis.go
```diff
@@ -179,7 +179,7 @@ func (sp *StateProcessor[
 			)
 		}
 		minEffectiveBalance := math.Gwei(
-			sp.cs.EjectionBalance() + sp.cs.EffectiveBalanceIncrement(),
+			sp.cs.EjectionBalance(false) + sp.cs.EffectiveBalanceIncrement(),
```
abi87, line 182: @calbera question: should the Increment change too? I would say so.

abi87 (commented):
LGTM, but I would change the EffectiveBalanceIncrement to a value that make sense given the new min and max

state-transition/core/state_processor_genesis.go
calbera: done

PR #2337, opened 2025-01-03
chore: update finalize_block.go

Anon-im:


abi87:
Not accepting typos only PRs for the time being

PR #2346, opened 2025-01-10
chore(ci): update token variable

chuck-bear:


gummybera:
Thanks @chuck-bear , this is been incorporated in the recent updates to the build workflow. ✅

PR #2349, opened 2025-01-11
fix: correct iteration logic for constants.MaxBytesPerTx and constants.MaxTxsPerPayload

Himess:
### Problem:
The constants `constants.MaxBytesPerTx` and `constants.MaxTxsPerPayload` were previously treated as iterable types with a `range` loop, but they are of type `uint64`. This caused incorrect usage of the `range` keyword, which is incompatible with integer types.

### Solution:
The iteration logic for both constants was corrected by replacing the `range` loop with an explicit integer-based loop using `for i := 0; i < ...; i++`. This ensures proper handling of the uint64 values.

### Test:
To confirm the type of these constants, the following test was performed:

1. A simple Go program was written to print the types of `constants.MaxBytesPerTx` and `constants.MaxTxsPerPayload`.
2. The test revealed that both constants are of type `uint64`.

I have also attached a screenshot of the test I performed below:
![Test Output]
Berachain hata bulma

---

This fix ensures compatibility with Go standards and eliminates potential logical issues in the iteration logic.

### DIFF
--- a/engine-primitives/engine-primitives/transactions_test.go
+++ b/engine-primitives/engine-primitives/transactions_test.go
@@ -72,7 +72,7 @@ var prysmConsistencyTests = []struct {
 		name: "max bytes per tx",
 		txs: func() [][]byte {
 			var tx []byte
-			for i := 0; i < int(constants.MaxBytesPerTx); i++ { // int dönüşümü eklendi
+			for i := 0; i < int(constants.MaxBytesPerTx); i++ {
 				tx = append(tx, byte(i))
 			}
 			return [][]byte{tx}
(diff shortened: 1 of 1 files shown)


Himess:
Current State of the Code
Existing Issue: In transactions_test.go, the usage of range constants.MaxBytesPerTx and range constants.MaxTxsPerPayload works without errors, as confirmed by the attached go test ./... results. However, this approach has the potential for issues when working with uint64 values, especially if they exceed int limits in future scenarios.
Test Verification: The original code without any changes passes all tests successfully.
image
This indicates that the current use of range does not cause any immediate issues.

Purpose of the Change:
Purpose of the Change
Replaced range with Explicit for Loop: Updated iterations to use for i := 0; i < int(...); i++ instead of range.
Added int Conversion: Ensured type compatibility while maintaining explicitness and robustness in the iteration logic.
Future Compatibility: Guarantees that the code will continue to work without errors if the uint64 constants exceed int limits, avoiding potential runtime errors or type mismatches

Outcome of the Change
Tests were re-run after implementing the change.
All tests pass successfully, as seen in the updated go test ./... output.
image

abi87:
This is actually possible following https://tip.golang.org/doc/go1.22

PR #2352, opened 2025-01-15
feat(storage): add batch operations interface to DB

crStiv:
### Added support for batch operations in the DB interface to allow atomic transactions. 

This will help when we need to perform multiple database operations that should either all succeed or all fail together.

### Changes:
- Added Batch() method to create new batch operations
- Added Close() method to properly cleanup DB resources
- Created new Batch interface with basic operations (Set/Delete/Write/Reset)
- Removed old TODO comment

### This should make it easier to handle multiple DB operations safely and improve performance by reducing the number of separate transactions.


abi87:
Does not even compile

PR #2353, opened 2025-01-15
fix: Fix package import path for engineprimitives

mdqst:
I noticed a small issue with the import path for the `engineprimitives` package. The current path duplicates the package name, which seems unintended. I’ve updated it to the correct path:  

```go
engineprimitives "github.com/berachain/beacon-kit/engine-primitives"
```  

This should resolve any potential confusion or errors related to the package naming.


fridrik01:
does not compile

PR #2355, opened 2025-01-15
fix: Fix incorrect value in test_DepositNative() function

0xminds:
I noticed a typo in the `test_DepositNative()` function where `32 gwei` was used instead of `32 ether`. This inconsistency could lead to incorrect test logic since the code above uses `32 ether` for the deposit. I’ve corrected the value to `32 ether` to match the intended behavior.  

**Changes:**  
- Replaced `32 gwei` with `32 ether` in the `emit IDepositContract.Deposit()` call.


calbera:
Its supposed to emit the amount in `gwei` in the Deposit event! Thanks

PR #2359, opened 2025-01-17
fix: Optimize Consensus Configuration and Merkle Tree Benchmarks

crStiv:
1. Consensus Configuration:
   - Removed deprecated `skip_timeout_commit`
   - Set `timeout_commit = "0"` for faster block progress

2. Merkle Tree Tests:
   - Added helper function for test data generation
   - Added benchmarks for different tree sizes (Small/Medium/Large)

### DIFF
--- a/primitives/merkle/tree_test.go
+++ b/primitives/merkle/tree_test.go
@@ -21,6 +21,7 @@
 package merkle_test
 
 import (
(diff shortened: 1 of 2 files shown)


abi87:
Does not make sense

PR #2365, opened 2025-01-18
fix(consensus): explicitly check consensus txs count

abi87:
We only allow two "transactions" to be included into a consensus block: a block and a blob.
This PR explicitly marks as invalid blocks containing more than 2 transactions


fridrik01 (commented):
lgtm

fridrik01 (approved):

PR #2367, opened 2025-01-18
fix(execution-client): duplicated payload duration metric

abi87:
I believe `measureNewPayloadDuration` was wrongly called multiple times:
- In `createContextWithTimeout` which is called by multiple methods (`NewPayload`, `ForkchoiceUpdated` and `GetPayload`)
- In `EngineClient.NewPayload`

This PR fixes this

### DIFF
--- a/execution/client/helpers.go
+++ b/execution/client/helpers.go
@@ -22,7 +22,6 @@ package client
 
 import (
 	"context"
-	"time"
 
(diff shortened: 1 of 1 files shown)


rezbera:
Investigating the metrics current output first

rezbera (approved):
Confirmed that the metric is currently `beacon_kit_execution_client_new_payload_duration` correlates with spikes with `forkchoice_updated` metrics in EL, signalling this metric is incorrect.

PR #2370, opened 2025-01-18
Replace blutgang with spamoor replicas

fridrik01:
This PR replaces the use of a load balancer (blutgang by default but ngingx also suppported) when running a single instance of spamoor with a running multiple replicas of spamoor where each target a single execution client.

This replicates how we conduct testing in our staging env we have had issues with blutgang where it was running unstably and restarting frequently. Also when scanning the blutgang logs it frequently removes nodes from the active set due to them falling behind (and then later adding them back) resulting in spamoor frequently hitting the same rpc node.

NOTE: The e2e CI test is failing, this is fixed in followup PR where we remove blutgang/nginx altogether from all integration tests

### DIFF
--- a/kurtosis/main.star
+++ b/kurtosis/main.star
@@ -171,52 +171,27 @@ def run(plan, network_configuration = {}, node_settings = {}, eth_json_rpc_endpo
     for n, seed_node in enumerate(seed_nodes):
         beacond.dial_unsafe_peers(plan, seed_node.cl_service_name, all_consensus_peering_info)
 
-    # Get only the first rpc endpoint
-    eth_json_rpc_endpoint = eth_json_rpc_endpoints[0]
-    endpoint_type = eth_json_rpc_endpoint["type"]
-    plan.print("RPC Endpoint Type:", endpoint_type)
-    if endpoint_type == "nginx":
-        plan.print("Launching RPCs for ", endpoint_type)
-        nginx.get_config(plan, eth_json_rpc_endpoint["clients"])
-
-    elif endpoint_type == "blutgang":
-        plan.print("Launching blutgang")
-        blutgang_config_template = read_file(
-            constants.BLUTGANG_CONFIG_TEMPLATE_FILEPATH,
-        )
-        blutgang.launch_blutgang(
-            plan,
-            blutgang_config_template,
-            full_node_el_clients,
-            eth_json_rpc_endpoint["clients"],
-            "kurtosis",
-        )
-
-    else:
-        plan.print("Invalid type for eth_json_rpc_endpoint")
-
     # 7. Start additional services
     prometheus_url = ""
     for s_dict in additional_services:
         s = service_module.parse_service_from_dict(s_dict)
         if s.name == "spamoor":
             plan.print("Launching spamoor")
-            ip_spamoor = plan.get_service(endpoint_type).ip_address
-            port_spamoor = plan.get_service(endpoint_type).ports["http"].number
-            spamoor.launch_spamoor(
+            if "replicas" not in s_dict:
+                s.replicas = 1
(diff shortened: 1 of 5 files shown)


kurtosis/src/services/spamoor/launcher.star
```diff
@@ -1,31 +1,34 @@
...
 MIN_CPU = 100
 MAX_CPU = 500
 MIN_MEMORY = 20
 MAX_MEMORY = 300
 
-def launch_spamoor(plan, funding_account, rpc_endpoint):
-    config = get_config(funding_account, rpc_endpoint)
-    plan.add_service(SERVICE_NAME, config)
-    plan.print(config)
+def launch_spamoors(plan, replicas, next_free_prefunded_account, full_node_el_client_configs, full_node_el_clients):
```
rezbera, line 13: m8 ur a kurtosis expert @fridrik01

rezbera (commented):
`2025-02-12 16:02:35 time="2025-02-12T05:02:35Z" level=warning msg="blob tx    110.0 failed: INTERNAL_ERROR: blobs limit in txpool is full" rpc="172.16.0.13:8545" scenario=blob-combined wallet=10`

Maxing out blobs locally - imo we should reduce the blob load a bit as its not sustainable

kurtosis/main.star
```diff
@@ -182,52 +182,27 @@ def run(plan, network_configuration = {}, node_settings = {}, eth_json_rpc_endpo
...
     for s_dict in additional_services:
         s = service_module.parse_service_from_dict(s_dict)
         if s.name == "spamoor":
             plan.print("Launching spamoor")
-            ip_spamoor = plan.get_service(endpoint_type).ip_address
-            port_spamoor = plan.get_service(endpoint_type).ports["http"].number
-            spamoor.launch_spamoor(
+            if "replicas" not in s_dict:
+                s.replicas = 1
+            next_free_prefunded_account = spamoor.launch_spamoors(
```
rezbera, line 193: any idea what this `next_free_prefunded_account` is being used for?

rezbera, line 193: might be worth making it clear this `next_free_prefunded_account` is actually the index into the `PRE_FUNDED_ACCOUNTS` array

fridrik01:
> ```shell
> 2025-02-12 16:02:35 time="2025-02-12T05:02:35Z" level=warning msg="blob tx    110.0 failed: INTERNAL_ERROR: blobs limit in txpool is full" rpc="172.16.0.13:8545" scenario=blob-combined wallet=10
> ```
> 
> ^ log I got from spamoor locally
> 
> Maxing out blobs locally - imo we should reduce the blob load a bit as its not sustainable. beyond that lgtm

Oh, interesting, I was able to replicate this as well. I lowered it from 8 to 4 replicas which it seems to sustain. We should consider lowering tx-fuzz repliaces, currently its firing 16 instances of that at the same time which means we often end up with over 1000 tx per block. But I tend to rather just switch to polygon-cli instead of tuning tx-fuxx rn  which is what we used mostly in testing.

fridrik01:
I merged in #2370   for the CI e2e to pass

testing/e2e/config/defaults.go
```diff
@@ -96,17 +95,17 @@ func defaultFullNodes() NodeSet {
 		Nodes: []Node{
 			{
 				ElType:   "nethermind",
-				Replicas: 1,
+				Replicas: 0,
```
rezbera, line 98: unsure of this change

rezbera (approved):
Unsure of 1 change wrt to defaultFullNodes. I'm not sure if that change is related to which full nodes we run and whether we should set nethermind to `0`. Besides that LGTM, happy to see blutgang gone and bloat reduced

PR #2371, opened 2025-01-18
Remove blutgang/nginx from e2e tests and just randomize clients

fridrik01:
This PR updates our e2e tests to randomize calls to execution clients instead of calling them through a load balancer like `nging`/`blutgang`.

In our staging env we have had issues with `blutgang` where it was running unstably and restarting frequently. Also when scanning the `blutgang` logs it frequently removes nodes from the active set due to them falling behind (and then later adding them back) even if there is almost no load on the network. This resulted in usually always the same node receiving traffic during testing.


testing/e2e/config/defaults.go
```diff
@@ -116,7 +115,7 @@ func defaultFullNodes() NodeSet {
 			},
 			{
 				ElType:   "besu",
-				Replicas: 1,
+				Replicas: 0,
```
rezbera, line 118: agree with killing besu. nethermind is still a pretty good EL though from what I know. not sure we should kill it off support yet

testing/e2e/suite/suite.go
```diff
@@ -89,21 +90,20 @@ func (s *KurtosisE2ESuite) KurtosisCtx() *kurtosis_context.KurtosisContext {
...
-func (s *KurtosisE2ESuite) JSONRPCBalancerType() string {
-	return s.cfg.EthJSONRPCEndpoints[0].Type
+func (s *KurtosisE2ESuite) RandomExecutionClient() *types.ExecutionClient {
+	ecKeys := make([]string, 0, len(s.executionClients))
+	for key := range s.executionClients {
+		ecKeys = append(ecKeys, key)
+	}
+	randomIndex, err := rand.Int(rand.Reader, big.NewInt(int64(len(ecKeys))))
+	if err != nil {
+		panic(err)
```
rezbera, line 104: any reason we panic here instead of 

```go
		s.Require().NoError(err, "..")
```

testing/e2e/suite/types/json_rpc_connection.go
```diff
@@ -35,26 +35,39 @@ type JSONRPCConnection struct {
...
-	)
+func NewJSONRPCConnection(serviceCtx *services.ServiceContext) (*JSONRPCConnection, error) {
+	conn := &JSONRPCConnection{
+		isWebSocket: true,
+	}
 
-	// If the WebSocket port isn't available, try the HTTP port
-	port, ok := serviceCtx.GetPublicPorts()["eth-json-rpc"]
+	// Start by trying to get the public port for the JSON-RPC WebSocket
+	jsonRPC, ok := serviceCtx.GetPublicPorts()["eth-json-rpc-ws"]
```
rezbera, line 44: Do we ever actually use websockets in our tests?

rezbera (approved):
supportive of removing blutgang and just randomizing the clients. 

I was a bit concerned about race conditions between the various execution layers updating could cause flakey tests but it seems from that perspective the behaviour is no different to how it was before.

few nits, no blockers

PR #2377, opened 2025-01-19
fix: Add upper bound validation for MaxWithdrawalsPerPayload

VolodymyrBg:
Description:
This PR adds an important security enhancement to the chain specification validation by implementing an upper bound check for MaxWithdrawalsPerPayload. This prevents potential DoS attacks and resource exhaustion that could occur if an extremely large number of withdrawals were allowed per payload.

Changes:
- Added ErrExcessiveMaxWithdrawalsPerPayload error constant
- Implemented upper bound validation (16384 withdrawals) in chainSpec.validate()
- Improved code readability with better variable naming
- Added comments explaining the rationale for the limits

Security Impact:
This change prevents potential DoS attacks by limiting the maximum number of withdrawals that can be processed in a single payload, while still maintaining reasonable operational capacity for the network.

Testing:
The existing test suite covers the validation logic, and the new limit is set to a practical value that shouldn't impact normal operation while preventing abuse.


abi87:
No plans to change MaxEffectiveBalance nor MaxValidatorsPerWithdrawalsSweep

PR #2381, opened 2025-01-19
Wrap pebble DB to use SetSync always with deposit store.

GrizzlyBera:
We see scenarios where an unclean shutdown leaves the deposit data in an inconsistent state. Specifically, the stored index has been incremented but the the deposit does not seems to be persisted. This may be because the backing store is PebbleDB and it does not flush to disk very often.
This change wraps PebbleDB and re-directs the Set(...) method to always use the SetSync method, which always flushes. Given that deposits are relatively rare, this should not have any noticeable impact on performance.


fridrik01:
Great idea! Confirmed that this addresses the deposit store mismatch on SIGKILL

fridrik01 (approved):
💯

abi87:
Also confirmed I could get missed deposits roots mismatch consistently when running a node in the IDE and killing it.
This PR never shows up the deposit root mismatch

node-core/components/deposit_store.go
```diff
@@ -42,6 +43,48 @@ type DepositStoreInput[
 	AppOpts config.AppOptions
 }
 
+type syncedPDB struct {
```
calbera, line 46: can move to its own package in storage/

abi87 (approved):
lgtm

calbera (approved):

PR #2383, opened 2025-01-20
chore(kurtosis): kurtosis setup to get deposit contract storage slots and fill genesis files

nidhi-singh02:
As the cli changes related to genesis set-deposit-storage are merged into main, spinning up a kurtosis network works successfully.


calbera (commented):
Add a e2e that checks the EL deposit contract state

calbera:
Also want to add this setup in the make start (entrypoint.sh) procedure, so that a local chain also has it correctly set

nidhi-singh02:
> Add a e2e that checks the EL deposit contract state. should call
> 
> * `depositCount()(uint64)` and expect it matches # of vals
> * `genesisDepositsRoot()(bytes32)` and [optionally] expect it equal to HashTreeRoot of genesis deposits
> 
> Can do this at the beginning of e2e staking test. The eth calls should be done with block height specified to 0.

Added checks in e2e_staking_test.go for calling both the above functions. Added a TODO comment for calculating the hashtree root of genesis deposits (as it is optional)

nidhi-singh02:
> Also want to add this setup in the make start (entrypoint.sh) procedure, so that a local chain also has it correctly set

Yeah, the plan is to add as separate PR, so the PR size doesn't grow bigger.

calbera (approved):
mostly taken from our work on internal, LGTM!

PR #2385, opened 2025-01-20
lower test iterations to address slow sidecar test

fridrik01:
The `test-unit-cover` test takes over 9minutes to run on our CI which is double the next longest CI task (test-e2e). This lowers the maximum iterations in the blob sidecar test so it should take approximately the same time as the test-e2e


fridrik01:
was fixed in https://github.com/berachain/beacon-kit/pull/2459

PR #2388, opened 2025-01-21
fix: fix the error handling issue in the ```Has``` function of ```coreKVStore```

MelodyFish:
Potential problems:

Hide potential errors: ```store.kvStore.Has(key)``` may return an error, but the current code does not handle it. Returning nil directly as an error may cover up possible problems with the underlying storage.


fridrik01 (changes requested):
`store.kvStore.Has()` returns only a bool so this PR does not compile

abi87:
Likely the usual, meaningless, AI generate PR

PR #2391, opened 2025-01-21
feat: print results from create-validator in json for easier parsing

fridrik01:
Example output:

```
> beacond deposit create-validator ZOM+GGN51wz3G9Ic+bLlMzXxfKw7QYz37JIYDu9gElI= 10000000000000 0x04000000 0xc811a46d522917e51f1a2b71886864edd0e9a3fedd403c47009b978c4797e5fa --home=/var/folders/77/zm_q7xw16dd5c658yts2g4xr0000gn/T/deposit-injection/fb6a8def06
{"pubkey":"0x81d29656e3080ab53ab258a72e25d1f0f8d557abb2e0613cac9f3b263a0a99aa697642aa48b97e528600f8ab7d057089","credentials":"0x0100000000000000000000000000000000000000000000000000000000000000","amount":"0x9184e72a000","signature":"0x8c4b63c94c761b118a55c917c97ed3c52e2bcfed2094cfd86b0b9f889ea24a232d838bed7fbb140d7d15492d8c79344a0ab8df1cbc86868e81888e362bb766608d324d5c2216fc2eae4f857f936c932f1a797a5df25fe6c8ad7444547f9ae13a","index":0}
```


cli/commands/deposit/create.go
```diff
@@ -119,15 +115,24 @@ func createValidatorCmd(
...
-			"withdrawal credentials", depositMsg.Credentials.String(),
-			"amount", depositMsg.Amount,
-			"signature", signature.String(),
-		)
+		type depositOutput struct {
+			Pubkey      string `json:"pubkey"`
+			Credentials string `json:"withdrawal_credentials"`
+			Amount      string `json:"amount"`
+			Signature   string `json:"signature"`
+		}
```
abi87, line 123: @fridrik01 if we export this struct, we could reuse it whenever we need to parse the output

fridrik01: We have a `types.Deposit` struct we could reuse here. I has the amount in gwei but the caller can convert to wei if needed

abi87 (approved):
LGTM

PR #2392, opened 2025-01-21
2/2 chore(cli): improve tooling around deposits

calbera:
## Quality of life tooling improvements to `beacond deposit`:
- `create-validator [withdrawal-address] [amount] ?[beacond/genesis.json]`: Creates a validator deposit with the necessary credentials. The arguments are expected in the order of withdrawal address, deposit amount, and **optionally the beacond genesis file. If the genesis validator root flag is NOT set, the beacond genesis file MUST be provided as the last argument**. If the broadcast flag is set to true, a private key must be provided to sign the transaction..
- `validate [pubkey] [withdrawal-credentials] [amount] [signature] ?[beacond/genesis.json]`: Validates a deposit message (public key, withdrawal credentials, deposit amount) for creating a new validator. The args taken are in the order of the public key, withdrawal credentials, deposit amount, signature, and **optionally the beacond genesis file. If the genesis validator root flag is NOT set, the beacond genesis file MUST be provided as the last argument.**

### Mainnet value for genesis validator root
`"0xdf609e3b062842c6425ff716aec2d2092c46455d9b2e1a2c9e32c6ba63ff0bda"`. **This never changes.** Will document this in all staking guides & specs for new validators.

**General notice:** the env var `CHAIN_SPEC` determines which chain's values are being to used in ALL of beacond cli cmds. "mainnet" is the default if env var is NOT set.


abi87 (commented):
Still kinda torn about this.
I think ideally we want to extend the tool in the direction of making it easy to use even in e2e tests.
What would be great is being able to feed BLS keys from the CLI and get the signing we want.
This seems to go in the direction of having this tool to work for specific chains only instead (mainnet)?

Also we should not keep adding code without UTs I think.
Will keep reviewing this later on today

calbera:
> Still kinda torn about this. I think ideally we want to extend the tool in the direction of making it easy to use even in e2e tests. What would be great is being able to feed BLS keys from the CLI and get the signing we want. This seems to go in the direction of having this tool to work for specific chains only instead (mainnet)?
> 
> Also we should not keep adding code without UTs I think. Will keep reviewing this later on today

This PR just adds the optional flag to compute genesis validaotr root from genesis file as opposed to requiring the value as an argument itself.

codingwithmanny (approved):
Relatively LGTM with this

shotes:
LGTM. Mostly just shifting code around, makes it easier to just supply genesis file instead of grabbing the root from some other source.

cli/commands/deposit/create.go
```diff
@@ -30,38 +30,45 @@ import (
...
 // NewCreateValidator creates a new command to create a validator deposit.
 //
-//nolint:lll // reads better if long description is one line
+//nolint:lll // reads better if long description is one line.
 func NewCreateValidator(
 	chainSpec chain.Spec,
 ) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "create-validator",
+		Use:   "create-validator [withdrawal-address] [amount] ?[genesis-validator-root]",
```
rezbera, line 52: A concern is that this pattern encourages copy pasting values that are not reader friendly

rezbera: I would recommend that we encourage genesis-file as the default, i.e. `create-validator [withdrawal-address] [amount] ?[genesis-file]`

 and use a flag for the power users

cli/commands/deposit/utils.go
```diff
@@ -0,0 +1,60 @@
...
+		return common.Root{}, err
+	}
+	if genesisFile != defaultGenesisFile {
+		if genesisValidatorRoot, err = genesis.ComputeValidatorsRootFromFile(
+			genesisFile, chainSpec,
+		); err != nil {
+			return common.Root{}, err
+		}
+	} else {
+		if len(args) < maxArgs {
```
rezbera, line 49:
```suggestion
		if len(args) != maxArgs {
```
more explicit

cli/commands/deposit/deposit.go
```diff
@@ -54,26 +59,29 @@ func Commands(
...
-//nolint:mnd,lll // lots of magic numbers, reads better if long description is one line
+//nolint:lll // reads better if long description is one line.
 func NewValidateDeposit(chainSpec chain.Spec) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "validate",
+		Use:   "validate [pubkey] [withdrawal-credentials] [amount] [signature] ?[genesis-validator-root]",
 		Short: "Validates a deposit message for creating a new validator",
-		Long:  `Validates a deposit message for creating a new validator. The deposit message includes the public key, withdrawal credentials, and deposit amount. The args taken are in the order of the public key, withdrawal credentials, deposit amount, signature, and genesis validator root.`,
-		Args:  cobra.ExactArgs(6),
+		Long:  `Validates a deposit message (public key, withdrawal credentials, deposit amount) for creating a new validator. The args taken are in the order of the public key, withdrawal credentials, deposit amount, signature, and optionally a genesis validator root. If the genesis file flag is NOT set, the genesis validator root MUST be provided as an argument.`,
```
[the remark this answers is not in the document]
calbera: it validates that the deposit message is verifiable (i.e. the BLS signature verifies against the pubkey used to sign)

cli/commands/deposit/create.go
calbera: down for this and @codingwithmanny agrees, will change

abi87 (approved):
lgtm

cli/commands/deposit/utils.go
```diff
@@ -0,0 +1,55 @@
...
+
+	if genesisValidatorRootStr != defaultGenesisValidatorRoot {
+		genesisValidatorRoot, err = parser.ConvertGenesisValidatorRoot(genesisValidatorRootStr)
+	} else {
+		if len(args) != maxArgs {
+			return common.Root{}, errors.New(
+				"genesis validator root is required if not using the genesis file flag",
+			)
+		}
+		genesisValidatorRoot, err = genesis.ComputeValidatorsRootFromFile(args[maxArgs-1], chainSpec)
```
rezbera, line 51:
```suggestion
		genesisValidatorRoot, err = genesis.ComputeValidatorsRootFromFile(args[validateDeposit4], chainSpec)
```
Consider same pattern as the other indexes into the args that use `iota`

rezbera (approved):

fridrik01 (approved):

PR #2393, opened 2025-01-22
1/2 fix(cli): Align `genesis validator-root` cmd with processing of genesis validators

calbera:
Correctly return the `GenesisValidatorRoot`, that is set in the BeaconState during processing of genesis, in the CLI command `genesis validator-root`. 

According to stateProcessor.processGenesisActivations, we now run 
```golang
    // mimic stateProcessor.processGenesisActivations
    if val.GetEffectiveBalance() >= minEffectiveBalance {
        val.SetActivationEligibilityEpoch(0)
        val.SetActivationEpoch(0)
    }
```
before calculating the HashTreeRoot of the Validators list.


rezbera (approved):

cli/commands/genesis/root.go
```diff
@@ -79,6 +79,15 @@ func GetGenesisValidatorRootCmd(cs chain.Spec) *cobra.Command {
 					math.Gwei(cs.EffectiveBalanceIncrement()),
 					math.Gwei(cs.MaxEffectiveBalance()),
 				)
+
+				// mimic processGenesisActivation
+				minEffectiveBalance := math.Gwei(
```
abi87, line 84: this is technically some more code duplication, the same that caused the CLI to break.
I will look for a way to consolidate the code more, so that we reuse code from state_processor

abi87: Also, if we cannot remove code duplication we should have UTs testing this, Ideally with quick. Will work on this

abi87 (changes requested):
remove code duplication completelly ( I don't think it's easy) or add UTs that compare implementations (via quick)

cli/commands/genesis/root.go
```diff
@@ -25,24 +25,33 @@ import (
...
 	"github.com/berachain/beacon-kit/primitives/encoding/json"
 	"github.com/berachain/beacon-kit/primitives/math"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 )
 
+// Deposits, Beacon, AppState and Genesis are code duplications that
+// collectively reproduce part of genesis file structure
+
+type Deposits []struct {
```
abi87, line 38: This cleanup is needed to be able to unit test this

cli/commands/genesis/root_test.go
```diff
@@ -0,0 +1,225 @@
...
+	amount2 = math.U64(5052)
+	idx2    = int(1)
+
+	emptySignature  = crypto.BLSSignature{}
+	expectedValRoot = common.Root{
+		0xa3, 0xfa, 0xd, 0x97, 0x0, 0xeb, 0xdc, 0x2c,
+		0x2, 0x1b, 0x51, 0xa1, 0xb, 0xcb, 0xb4, 0x80,
+		0x5d, 0xe6, 0x13, 0x53, 0x9a, 0x77, 0xdc, 0x65,
+		0xc7, 0x64, 0x85, 0x36, 0x6, 0xde, 0x4f, 0xa2,
+	}
```
abi87, line 76: these are the common inputs we fed to the two unit tests below and the common output we expect from them.
This is a step on the way of introducing a property tests that assert equality of the two way of processing things

cli/commands/genesis/root_test.go
```diff
@@ -0,0 +1,225 @@
...
+	*statedb.StateDB,
+	*depositstore.KVStore,
+	*transition.Context,
+) {
+	t.Helper()
+
+	execEngine := mocks.NewExecutionEngine(t)
+
+	mocksSigner := &cryptomocks.BLSSigner{}
+	mocksSigner.On(
```
abi87, line 154: All of this code and the one below is currently duplicated from state processor.
I will move it to a common, testing package, to make it available to multiple packages without code duplication

state-transition/core/core_test.go
```diff
@@ -21,89 +21,21 @@
...
+	statetransition "github.com/berachain/beacon-kit/testing/state-transition"
 	"github.com/stretchr/testify/require"
 )
 
-type (
-	TestBeaconStateMarshallableT = types.BeaconState
-
-	TestBeaconStateT = statedb.StateDB
-
-	TestStateProcessorT = core.StateProcessor[*transition.Context]
```
abi87, line 63: all the code deleted here has been moved to `testing/state-transition' in order to reuse without duplicating it

cli/commands/genesis/root_test.go
abi87: Done

cli/commands/genesis/root.go
```diff
@@ -66,26 +67,35 @@ func GetGenesisValidatorRootCmd(cs chain.Spec) *cobra.Command {
...
+			validatorHashTreeRoot := ValidatorsRoot(genesis.Deposits, cs)
+			cmd.Printf("%s\n", validatorHashTreeRoot)
 			return nil
 		},
 	}
 
 	return cmd
 }
+
+func ValidatorsRoot(deposits types.Deposits, cs chain.Spec) common.Root {
```
abi87, line 79: extracted the core functionality for unit tests

cli/commands/genesis/root_test.go
abi87: Now replaced with property test

cli/commands/genesis/root.go
```diff
@@ -24,25 +24,26 @@ import (
...
 	"github.com/berachain/beacon-kit/primitives/encoding/json"
 	"github.com/berachain/beacon-kit/primitives/math"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 )
 
+// Beacon, AppState and Genesis are code duplications that
+// collectively reproduce part of genesis file structure
+
+type Beacon struct {
```
abi87, line 37: minor cleanup of these types to unlock UTs.
We really should not need this. We should have a single type representing the genesis file structure and reuse it anywhere.
Stuff for another PR

cli/commands/genesis/root_test.go
```diff
@@ -0,0 +1,123 @@
...
+			Credentials: creds,
+			Amount:      amount,
+			Signature:   sign,
+			Index:       0, // indexes will be set in order in the test
+		}
+	}
+	return reflect.ValueOf(res)
+}
+
+func TestCompareGenesisCmdWithStateProcessor(t *testing.T) {
```
abi87, line 80: a property test comparing the two ways we have to generate the validator genesis root (CLI vs StateProcessor), to make sure they return the same output given the same inputs

cli/commands/genesis/root_test.go
```diff
@@ -0,0 +1,123 @@
...
+	qc := &quick.Config{MaxCount: 1_000}
```
abi87, line 81: 1000 samples seems to take ~30 secs on CLI to complete this test VS the current ~7m30s to complete compare_test

nidhi-singh02 (approved):

cli/commands/genesis/root.go
rezbera, line 37: Bit unclear if this is referring to `BeaconState` or `BeaconBlock`

testing/state-transition/state-transition.go
```diff
@@ -0,0 +1,142 @@
+// SPDX-License-Identifier: BUSL-1.1
```
rezbera, line 1: Consider renaming this file with `_test`-  iirc golang will package this with the prod build if not

testing/state-transition/README.md
```diff
@@ -0,0 +1,7 @@
+# Testing state transitions
+
+This package, `statetransition`, contains code which is helpful to test state transitions.
```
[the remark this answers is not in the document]
rezbera: The deeper problem is that we don't have separation of state and logic, so anytime we want to test logic, we need to have boilerplate state and the surrounding code. This is a bigger refactor outside of the scope of this PR

cli/commands/genesis/root_test.go
```diff
@@ -0,0 +1,121 @@
+// SPDX-License-Identifier: BUSL-1.1
```
rezbera, line 1: Rename to validator_root_test as there is also `depositsRoot` as there are other roots this could be confused with

calbera:
can't approve explicitly, but LGTM for now

rezbera (approved):

PR #2394, opened 2025-01-22
fix(blockchain): modify location and error handling of `forkchoiceUpdated` engine API calls

shotes:
This PR is to resolve issue https://github.com/berachain/beacon-kit/issues/2405

There are several fixes in here:
- call `NotifyForkchoiceUpdate` immediately after `VerifyAndNotifyNewPayload`. This validates that the EL is willing, able, and DOES, update the head of the chain to the proposed payload. This is the primary fix and resolves handling around the `ACCEPTED` status. This also enforces that errors in the forkchoiceUpdated propagate to consensus and aren't ignored.
- Fix a bug where an optimistic builder could send a forkchoice update that would attempt to un-finalize a block under one the following conditions (both of these conditions were bugs before the modifications in this PR with the immediate forkchoiceUpdates, but become much easier to accidentally happen with the new changes):
  1. A forkchoice update is sent at height `B` to set `X` as the final block and `Y` as the head. After sending this forkchoice update (in FinalizeBlock), the round does not reach consensus for *some* reason. Then the optimistic builder also fails a proposal that triggers an optimistic rebuild in `rebuildPayloadForRejectedBlock()`. This will attempt to set `W` (the block before `X`) as the final block, while setting `X` as the head in order to rebuild a block on `X`.
  2. The exact same scenario as 1, except instead of rebuilding optimistically, the proposer is attempting to grab an optimistic build, doesn't find one, and then requests a synchronous build via `RequestPayloadSync()`, which would result in the same issue.

Edit: my concerns about 1 and 2 are incorrect. In the live codebase (not this PR), fork choice updates are only sent in FinalizeBlock, and those block hashes are never to be changed, even if FinalizeBlock is called multiple times.


execution/engine/engine.go
```diff
@@ -89,13 +89,9 @@ func (ee *Engine) NotifyForkchoiceUpdate(
 	switch {
 	// We do not bubble the error up, since we want to handle it
 	// in the same way as the other cases.
-	case errors.IsAny(
-		err,
-		engineerrors.ErrAcceptedPayloadStatus,
```
shotes, line 94: `ErrAcceptedPayloadStatus` does not get returned by `notifyForkchoiceUpdated` engine API calls.

execution/engine/engine.go
```diff
@@ -89,13 +89,9 @@ func (ee *Engine) NotifyForkchoiceUpdate(
...
-		engineerrors.ErrSyncingPayloadStatus,
-	):
+	case errors.Is(err, engineerrors.ErrSyncingPayloadStatus):
```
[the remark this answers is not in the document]
shotes: After thinking about this more, I think this is right. 
- If the EL is syncing, we cannot verify the block. Therefore we should not give it the OK in a proposal.
- If we are a simple node catching up to the network, replaying blocks, then `FinalizeBlock` will be executed sequentially, guaranteeing that the EL always has the previous block. Therefore it should never be possible to get the SYNCING status.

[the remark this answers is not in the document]
abi87: https://github.com/cometbft/cometbft/blob/main/docs/references/rfc/rfc-105-non-det-process-proposal.md

state-transition/core/state_processor_payload.go
```diff
@@ -155,7 +157,31 @@ func (sp *StateProcessor[_]) validateStatefulPayload(
...
+		// by the following forkchoice update, turning the payload into VALID
+		// or INVALID.
+		return err
+	}
+
+	// Since we have single slot finality, the previous block is already
+	// considered final from cometBFT perspective. The newPayload being
+	// submitted from this proposal should be set as the new head of the
+	// chain, and we can update the finalized block with the EL. We send the
+	// FCU now which fully verifies that the block is VALID or INVALID.
```
rezbera, line 171: This bubbles up to the `s.stateProcessor.Transition` call which is also called in `ProcessProposal`.

Previously we would only set the head in the EL after `FinalizeBlock` as we confirmed with Informal that this was never going to revert. 

Now it seems possible that the "head", i.e. "latest" tag could revert given that we are setting it before `FinalizeBlock`. This would require that anyone waiting for finality must wait for the "finalized" tag in the EL which is friction we would have to discuss with DX

state-transition/core/state_processor.go
```diff
@@ -189,15 +188,7 @@ func (sp *StateProcessor[ContextT]) ProcessBlock(
 		return err
 	}
 
-	switch err := sp.processExecutionPayload(ctx, st, blk); {
-	case err == nil:
-		// keep going with the processing
-	case errors.Is(err, engineerrors.ErrAcceptedPayloadStatus):
```
calbera, line 195: agree with removal here. this was useless with the comment being inconsistent

calbera (changes requested):
As discussed we just handle errors for ACCEPTED in this PR. Consider updating FCU calls in future.

PR #2397, opened 2025-01-22
fix(cli): `beacond deposit validate` had wrong number of args

abi87:
This PR:

- Fix the number of arguments for `deposit validate`
- Add a quick test to show that deposits created via `deposit create-validator` do pass `deposit validate`
- Does minor code cleanup


cli/commands/deposit/create.go
```diff
@@ -30,10 +30,20 @@ import (
...
 	"github.com/berachain/beacon-kit/primitives/crypto"
+	"github.com/berachain/beacon-kit/primitives/version"
 	"github.com/spf13/cobra"
 )
 
+const (
+	createValWithdrawalAddrIdx = iota
+	createValDepAmtIdx         = iota
+	createValGenValRootIdx     = iota
+	createValArgsCount         = iota
```
abi87, line 44: this is ugly but it's a way to keep args in sync. Open to suggestiions

cli/commands/deposit/validate.go
```diff
@@ -28,39 +28,27 @@ import (
...
 // NewValidateDeposit creates a new command for validating a deposit message.
 //
-//nolint:mnd,lll // lots of magic numbers, reads better if long description is one line
+//nolint:lll // reads better if long description is one line
 func NewValidateDeposit(chainSpec chain.Spec) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "validate",
 		Short: "Validates a deposit message for creating a new validator",
 		Long:  `Validates a deposit message for creating a new validator. The deposit message includes the public key, withdrawal credentials, and deposit amount. The args taken are in the order of the public key, withdrawal credentials, deposit amount, signature, and genesis validator root.`,
-		Args:  cobra.ExactArgs(6),
```
abi87, line 63: This is the main fix. We don't need 6 arguments here I think. We only use 5

cli/utils/parser/validator.go
```diff
@@ -106,18 +106,6 @@ func ConvertSignature(signature string) (crypto.BLSSignature, error) {
...
-func ConvertVersion(version string) (common.Version, error) {
-	versionBytes, err := hex.ToBytes(version)
-	if err != nil {
-		return common.Version{}, err
-	}
-	if len(versionBytes) != constants.DomainTypeLength {
-		return common.Version{}, ErrInvalidVersionLength
-	}
-	return common.Version(versionBytes), nil
-}
```
abi87, line 119: not needed since we only use genesis version

cli/commands/deposit/create.go
```diff
@@ -30,10 +31,25 @@ import (
...
+
+	privateKey      = "private-key" // does not look really used
```
abi87, line 48: another finding: this `private-key` is not really used.

cli/commands/deposit/commands_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/berachain/beacon-kit/primitives/common"
+	"github.com/berachain/beacon-kit/primitives/crypto"
+	"github.com/berachain/beacon-kit/primitives/math"
+	cmtcfg "github.com/cometbft/cometbft/config"
+	cmtbls12381 "github.com/cometbft/cometbft/crypto/bls12381"
+	"github.com/cometbft/cometbft/privval"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCreateAndValidateCommandsDuality(t *testing.T) {
```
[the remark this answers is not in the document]
abi87: I think it also show why it's hard to test deposits at the moment. We can't just generate a rand BLS key, we do need to store it into a file for the BLSSigner to work properly

cli/commands/deposit/commands.go
```diff
@@ -0,0 +1,47 @@
...
+package deposit
+
+import (
+	"github.com/berachain/beacon-kit/chain"
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/spf13/cobra"
+)
+
+// Commands creates a new command for deposit related actions.
+func Commands(
```
abi87, line 30: just moved it to its own file where we only register the commands

cli/commands/deposit/create.go
```diff
@@ -127,11 +124,48 @@ func createValidatorCmd(
 
 		//nolint:forbidigo // simplifies output parsing
 		fmt.Print(string(val))
-
 		return nil
 	}
 }
 
+func CreateDepositMessage(
```
abi87, line 131: separated from args parsing to test this in `command_test.go`

cli/commands/deposit/validate.go
```diff
@@ -74,45 +64,61 @@ func validateDepositMessage(chainSpec chain.Spec) func(
...
+	}
+}
 
-		return depositMessage.VerifyCreateValidator(
-			types.NewForkData(genesisVersion, genesisValidatorRoot),
-			signature,
-			chainSpec.DomainTypeDeposit(),
-			signer.BLSSigner{}.VerifySignature,
-		)
+func ValidateDeposit(
```
abi87, line 101: separated from args parsing to test it in `commands_test.go`

rezbera (approved):
Some overlap with [CLI improvement](https://github.com/berachain/beacon-kit/pull/2392). LGTM

PR #2398, opened 2025-01-22
chore(e2e): Add verification during deposit e2e test

shotes:
This PR strictly correlates the amount of BERA deposited against the amount of validator power and withdrawal balance at the end of the test. 

It also modifies the node-api implementation for the validator endpoints. They now return correctly formatted data that works with `go-eth2-client`.


testing/e2e/suite/setup.go
```diff
@@ -243,7 +243,7 @@ func (s *KurtosisE2ESuite) FundAccounts() {
 				gasTipCap, big.NewInt(0).SetUint64(TenGwei))
 			nonceToSubmit := nonce.Add(1) - 1
 			//nolint:mnd // 20000 Ether
-			value := new(big.Int).Mul(big.NewInt(20000), big.NewInt(Ether))
+			value := new(big.Int).Mul(big.NewInt(200000), big.NewInt(Ether))
```
shotes, line 246: More funding to the test accounts. They did not have enough BERA to hit the devnet 4000e9 max effective balance for all 5 validators.

testing/e2e/e2e_staking_test.go
```diff
@@ -35,22 +38,38 @@ import (
...
+}
+
 // TestDepositRobustness tests sending a large number of deposits txs to the Deposit Contract.
 // Then it checks whether all the validators' voting power have increased by the correct amount.
 //
 // TODO:
-// 1) Add staking tests for exceeding the max stake.
-// 2) Add staking tests for adding a new validator to the network.
-// 3) Add staking tests for hitting the validator set cap and eviction.
+// 1) Add staking tests for adding a new validator to the network.
```
abi87, line 57: this requires simplifying (and probably fixing) tooling. Working on it

testing/e2e/e2e_staking_test.go
```diff
@@ -35,22 +38,38 @@ import (
...
+// 2) Add staking tests for hitting the validator set cap and eviction.
```
abi87, line 58: <3

testing/e2e/e2e_staking_test.go
```diff
@@ -173,13 +211,13 @@ func (s *BeaconKitE2ESuite) TestDepositRobustness() {
 	s.Logger().Info("Final deposit tx mined successfully", "hash", receipt.TxHash.Hex())
 
 	// Give time for the nodes to catch up.
-	err = s.WaitForNBlockNumbers(BlocksToWait)
+	err = s.WaitForNBlockNumbers(NumDepositsLoad / chainspec.MaxDepositsPerBlock())
```
abi87, line 214: q: given the way our deposit system work, this is the exact number of blocks we need to wait right? Not the minimum number of blocks we need to process all deposit?

testing/e2e/e2e_staking_test.go
```diff
@@ -188,37 +226,44 @@ func (s *BeaconKitE2ESuite) TestDepositRobustness() {
...
-	// s.Require().Equal(power1+increaseAmt, power1After)
-	// s.Require().Equal(power2+increaseAmt, power2After)
-	// s.Require().Equal(power3+increaseAmt, power3After)
-	// s.Require().Equal(power4+increaseAmt, power4After)
+	// Check that all validators' voting power have increased by
+	// (NumDepositsLoad / NumValidators) * depositAmountWei
+	// after the end of the epoch (next multiple of SlotsPerEpoch after receipt.BlockNumber).
+	blkNum, err = s.JSONRPCBalancer().BlockNumber(s.Ctx())
+	s.Require().NoError(err)
+	nextEpochBlockNum := (blkNum/chainspec.SlotsPerEpoch() + 1) * chainspec.SlotsPerEpoch()
```
abi87, line 242: nit: can we use `specs.SlotToEpoch` here?

abi87 (commented):
jsut a couple of questions, then LGTM

testing/e2e/e2e_staking_test.go
shotes: Right, this is a rough outline of the number of blocks we should wait for deposit inclusion. There is another `WaitForNBlockNumbers` afterwards that waits for the next epoch to check that we have processed validator set updates and withdrawals.

testing/e2e/e2e_staking_test.go
shotes: fixed in commit https://github.com/berachain/beacon-kit/pull/2398/commits/9e097d2c32b3b80f6055fd823f0a8fd6ed59f455

abi87 (approved):
lgtm

PR #2400, opened 2025-01-22
wip: deposit contract cleanup for local env

calbera:
- TODO: Remove solidity code for deposit contract, submodule into github.com/berachain/contracts
- Set deposit contract predeploy storage at genesis for local testing `make start` -- handled by https://github.com/berachain/beacon-kit/pull/2402


calbera:
waiting on https://github.com/berachain/contracts/pull/4

contracts/foundry.toml
dezzeus, line 1: In order to freeze the generated bytecode, you may want to also set other [configuration options](https://book.getfoundry.sh/reference/config/solidity-compiler) whose default value may change in future.

PR #2402, opened 2025-01-23
chore(genesis deposit) : Change in local network setup with single node

nidhi-singh02:
Using cli command for setting deposit contract storage slot values and using the correct genesis files(the one being modified by the cli command)


testing/files/entrypoint.sh
```diff
@@ -74,7 +75,9 @@ if [[ $overwrite == "y" || $overwrite == "Y" ]]; then
 		./build/bin/beacond genesis add-premined-deposit --home $HOMEDIR \
 			32000000000 0x20f33ce90a13a4b5e7697e3544c3083b8f8a51d4
 		./build/bin/beacond genesis collect-premined-deposits --home $HOMEDIR
-		./build/bin/beacond genesis execution-payload "$ETH_GENESIS" --home $HOMEDIR
+		./build/bin/beacond genesis set-deposit-storage "$ETH_GENESIS" --home $HOMEDIR
```
abi87, line 78: @nidhi-singh02 why do we execute both commands?

abi87 (commented):
I am not sure this change is complete.
Shouldn't we make sure that also `make start-<EL>` pick the right genesis file, as it was modified by the `set-deposit-storage` command?

testing/files/entrypoint.sh
nidhi-singh02: As we want to modify the deposit contract storage slot values for both the genesis files - for nethermind and for others. @calbera I hope my understanding is correct.

nidhi-singh02:
> I am not sure this change is complete. Shouldn't we make sure that also `make start-<EL>` pick the right genesis file, as it was modified by the `set-deposit-storage` command?

Yes, you are right, we are ensuring that as we changed the genesis path values as below - 

```
ETH_GENESIS_PATH = ${HOMEDIR}/eth-genesis.json
NETHER_ETH_GENESIS_PATH = ${HOMEDIR}/eth-nether-genesis.json
```

calbera:
2 things:
- would add the readme note from internal
- would ensure all EL clients startup correctly with the correct deposit contract state

testing/files/entrypoint.sh
calbera: Yeah the idea here is set both eth genesis files so that in case the user wants to start nethermind as the EL, its genesis file is also ready

nidhi-singh02:
I validated by running genesis sanity checks that we have for all the EL clients - geth, reth, erigon, besu and nethermind. 
Facing issues in spinning up a beacond-ethereumjs network, tried on main branch as well - same issue. I have opened up an issue to track it https://github.com/berachain/beacon-kit/issues/2408

I spent some time tweaking few parameters for ethereumjs,  it needs more time to dig into. Fixing something wrong that I see with ethereumjs genesis path in this PR.

calbera (approved):
LGTM. tested that the deposit contract queries correctly on all ELs, even ethereumjs. Regarding ethereumjs not starting up correctly, that is also the case on main, so unrelated to this PR. will keep issue #2408 open

PR #2403, opened 2025-01-23
feat(E2E): Enable e2e tests to run in different networks based on chain spec

nidhi-singh02:
As part of the change set, we will be able to spin up different networks on the basis of chain ID and the chain spec name combination each in its own kurtosis enclave and run the tests.

The instructions to add your tests is given in README.md.

This is the just the foundation of making the e2e tests customizable and being able to run different specs as part of our e2e suite. 

Note : Currently only devnet is supported. I see hardcoding of chainspec in few tests, needs to be refactored.

Feb 20 EDIT -- Converting to draft. Still desired but postponed.


nidhi-singh02:
After merging changes from main, e2e tests started breaking. Debugging it.

testing/e2e/README.md
```diff
@@ -27,15 +27,31 @@ This will automatically build your beacond docker image from the local source
...
 In case you want to configure(change) the validator set, consider doing changes in `defaultValidators`.
 The user can specify the number of replicas they want per type.
 
 All the default configuration are listed in `testing/e2e/config/defaults.go`
 
-Note: Currently the chainID for this local network is 80087, which is our dev network configuration (this is fixed in the kurtosis env setup and will be made configurable in a future version). To make changes to the 80087 chain spec used, modify parameters [here](https://github.com/berachain/beacon-kit/blob/main/config/spec/devnet.go#L40).
+Note: The default chainID for this local network is 80087, which is our dev network configuration. To make changes to the 80087 chain spec used, modify parameters [here](https://github.com/berachain/beacon-kit/blob/main/config/spec/devnet.go#L40).
+
+## Configure the default network configuration
+To change the chainID, modify the `ChainID` field in the `NetworkConfiguration` struct in `defaultNetworkConfiguration`
```
rezbera, line 43: Given that we default to `mainnet` chain spec in beacond - feels unintuitive to have the default for tests be `devnet`. Perhaps worth renaming

testing/e2e/config/defaults.go
```diff
@@ -36,6 +36,16 @@ const (
 	ClientValidator4 = "cl-validator-beaconkit-4"
 )
 
+// default network configuration.
+//
+//nolint:gochecknoglobals // it's a default value
+var (
+	// defaultChainID is the default chain ID for the network.
+	defaultChainID = 80087
```
rezbera, line 44: Should these be `const` given that they are not mutated?

testing/e2e/e2e_staking_test.go
```diff
@@ -57,7 +57,11 @@ type ValidatorTestStruct struct {
 // TODO:
 // 1) Add staking tests for adding a new validator to the network.
 // 2) Add staking tests for hitting the validator set cap and eviction.
-func (s *BeaconKitE2ESuite) TestDepositRobustness() {
+func (s *BeaconKitE2ESuite) runDepositRobustness() {
```
rezbera, line 60: seems unintuitive to have tests not be named as such - affects readability

rezbera: Also my IDE doesn't let me run this test individually now

testing/e2e/suite/suite.go
```diff
@@ -40,83 +43,121 @@ var Run = suite.Run
...
-	enclave *enclaves.EnclaveContext
+	logger log.Logger
+	ctx    context.Context
+	kCtx   *kurtosis_context.KurtosisContext
+
+	// Network management
+	networks  map[string]*NetworkInstance // maps chainSpec to network
+	testSpecs map[string]ChainSpec        // maps testName to chainSpec
+	testFuncs map[string]func()           // maps test names to test functions
+	mu        sync.RWMutex
```
rezbera, line 54: Can you expand on the reasoning for the mutex? it seems the write lock is only in `RegisterTest` which doesn't seem to be called concurrently

testing/e2e/README.md
```diff
@@ -27,15 +27,31 @@ This will automatically build your beacond docker image from the local source
...
+
+To change the chainSpec, modify the `ChainSpec` field in the `NetworkConfiguration` struct in `defaultNetworkConfiguration`
+function in `testing/e2e/config/defaults.go`.
 
 ## Add your tests
-Add your tests in here like how it is done in `TestBasicStartup()`
+To add your tests, you need to do the following:
+1. Create a new file in the `testing/e2e/` directory.
+2. Add your tests in here like how it is done in `runBasicStartup()`
+3. Register your tests in `TestRunE2E()`.
```
rezbera, line 53: Is there any way we can avoid this registration?

rezzmah (commented):
An approach to avoid explicit registration outside of the test is to have each test be "aware" of the chainspec it supports, so that I can look at a test and immediately know what chain spec it's supposed to run on. It would also allow me to run the test directly, without running the whole suite from my IDE (Goland).  A problem with "registration" based approaches is also forgetting to register.

Example
```go
func (s *BeaconKitE2ESuite) TestDepositRobustness() {	
	s.Logger().Info("Running Deposit Robustness on devnet")
	network, chainSpec := s.SetForNetwork("devnet")
```
SetForNetwork would effectively run `t.Skip()` if the test harness is attempting to run the test on a chain spec it doesn't support.

I was initially confused about the modularity goals we aimed to achieve but to be explicit, the user story is as follows:
- I want to be able to run tests that are opinionated for a particular chainspec.

We are NOT trying to do the following:
- Run the same test against multiple chainspecs.

testing/e2e/README.md
rezbera: Idea: https://github.com/berachain/beacon-kit/pull/2403#pullrequestreview-2597051660

calbera:
+1 to Rez's comment. Some things we should try fixing:

- currently broken -- if I try changing just 1 test to run on `mainnet` (and keep the others on `devnet`) the mainnet fails to startup.
- reminder -- we want to ideally run the different chains in parallel
- let's break up this PR into smaller PRs, something like:
  1. the refactor to just the `suite` which includes running multiple chains
  2. the setup of each of the e2e tests (as Rez mentioned above)
  3. the ability to run each of those chains in parallel

nidhi-singh02:
Will address the review comments/resume here once the node api endpoints for validators is complete, which is priority rn.

PR #2407, opened 2025-01-24
fix(deposits): Enforce deposit inclusion

shotes:
This PR enforces that a deposit must be included in the proposed BeaconBlock.

WARNING: I do not believe this is backwards compatible IF a single validator on the chain has ever decided not to include a deposit. Do note that if every single validator so far is using our published binary, this should not be the case and the change should be completely backwards compatible. If this change is to be made, we should either:
1. Sync a node with this change to the chain to ensure no deposits have been excluded. If there has been a single instance of deposit exclusion, then the sync *should* fail. In that case, we have to...
2. Fork it


state-transition/core/validation_deposits.go
```diff
@@ -78,17 +78,18 @@ func (sp *StateProcessor[_]) validateNonGenesisDeposits(
...
-	expectedLocalDepositsLen := depositIndex + uint64(len(blkDeposits))
-	localDeposits, err := sp.ds.GetDepositsByIndex(ctx, 0, expectedLocalDepositsLen)
+	// Grab all previous deposits from genesis up to the current index + max deposits per block.
+	localDeposits, err := sp.ds.GetDepositsByIndex(ctx, 0, depositIndex+sp.cs.MaxDepositsPerBlock())
 	if err != nil {
 		return err
 	}
+	totalProposedDeposits := depositIndex + uint64(len(blkDeposits))
 
 	// First verify that the local store returned all the expected deposits
```
abi87, line 88: I think we need to modify this comment here. The point is verifying that the block includes all the deposits seen in the local store

shotes: Fixed the comment in https://github.com/berachain/beacon-kit/pull/2407/commits/f62217da105e17d229358e65c68555809769e1e9

Also renamed the `totalProposedDeposits` to `totalBlockDeposits` to be more accurate, since this happens during state-transition.

abi87 (commented):
LGTM. Will approve as soon as we verify sync is fine

state-transition/core/validation_deposits.go
```diff
@@ -78,17 +78,18 @@ func (sp *StateProcessor[_]) validateNonGenesisDeposits(
 		return err
 	}
 
-	expectedLocalDepositsLen := depositIndex + uint64(len(blkDeposits))
-	localDeposits, err := sp.ds.GetDepositsByIndex(ctx, 0, expectedLocalDepositsLen)
+	// Grab all previous deposits from genesis up to the current index + max deposits per block.
+	localDeposits, err := sp.ds.GetDepositsByIndex(ctx, 0, depositIndex+sp.cs.MaxDepositsPerBlock())
```
rezbera, line 82: I'm not convinced we handle the error correctly if we query `GetDepositsByIndex` outside of the range of available blocks. 

The check
```go
		case errors.Is(err, sdkcollections.ErrNotFound):
			return deposits, nil
```
Should ideally be sufficient, but I'm not convinced that cosmos is correctly returning ErrNotFound for PebbleDB. 
```go
// Get implements DB.
func (db *PebbleDB) Get(key []byte) ([]byte, error) {
	// fmt.Println("PebbleDB.Get")
	if len(key) == 0 {
		return nil, errKeyEmpty
	}

	res, closer, err := db.db.Get(key)
	if err != nil {
		if err == pebble.ErrNotFound {
			return nil, nil
		}
		return nil, err
	}
	defer closer.Close()

	return cp(res), nil
}
```

Can we have a test for query beyond the available range?

rezbera: Confirmed with @abi87 that the Cosmos SDK collection will wrap the error

abi87 (approved):

PR #2411, opened 2025-01-25
feat(simulated): Testing Framework with Simulated CometBFT (1/n)

rezbera:
This testing framework allows us to test scenarios we previously were not able to test. It enables us to orchestrate the ABCI calls while introspecting state and responses with each individual call.

Start review from `simulated_test.go` to see what we can achieve which will explain the need for surrounding utilities

This PR introduces the following areas of coverage we previously lacked:
- Test handling of an invalid block EL block.

Many more future tests can be added but keeping it lean for now

2/4 - https://github.com/berachain/beacon-kit/pull/2521
3/4 - https://github.com/berachain/beacon-kit/pull/2523
4/4 - https://github.com/berachain/beacon-kit/pull/2534


execution/client/client.go
```diff
@@ -118,6 +118,7 @@ func (s *EngineClient) Start(
 			if err := s.verifyChainIDAndConnection(ctx); err != nil {
 				if errors.Is(err, ErrMismatchedEth1ChainID) {
 					s.logger.Error(err.Error())
+					return err
```
rezbera, line 121: This is a unrecoverable error and should return

cli/commands/errors.go
```diff
@@ -1,30 +0,0 @@
...
-// AN “AS IS” BASIS. LICENSOR HEREBY DISCLAIMS ALL WARRANTIES AND CONDITIONS,
-// EXPRESS OR IMPLIED, INCLUDING (WITHOUT LIMITATION) WARRANTIES OF
-// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, NON-INFRINGEMENT, AND
-// TITLE.
-
-package commands
-
-import "github.com/berachain/beacon-kit/errors"
-
-var (
```
rezbera, line 25: Unusued - cleanup

cli/commands/server/start.go
```diff
@@ -61,19 +61,6 @@ type StartCmdOptions[
 	AddFlags func(cmd *cobra.Command)
 }
 
-// StartCmd runs the service passed in, either stand-alone or in-process with
-// CometBFT.
-func StartCmd[
```
rezbera, line 66: Unusued - cleanup

cmd/beacond/types.go
```diff
@@ -21,122 +21,17 @@
...
-
-	// DepositContract is a type alias for the deposit contract.
-	DepositContract = deposit.WrappedDepositContract
-
-	// Eth1Data is a type alias for the eth1 data.
-	Eth1Data = types.Eth1Data
-
-	// Fork is a type alias for the fork.
-	Fork = types.Fork
-
```
rezbera, line 118: Unusued - cleanup

config/config.go
```diff
@@ -86,16 +86,6 @@ func (c Config) Template() string {
 	return template.TomlTemplate
 }
 
-// MustReadConfigFromAppOpts reads the configuration options from the given
-// application options.
-func MustReadConfigFromAppOpts(opts AppOptions) *Config {
```
rezbera, line 91: Unusued - cleanup

cli/commands/genesis/collect.go
```diff
@@ -43,64 +44,67 @@ func CollectGenesisDepositsCmd() *cobra.Command {
...
+	}
+
+	// create the app state
+	appGenesisState, err := genutiltypes.GenesisStateFromAppGenesis(
+		appGenesis,
+	)
+	if err != nil {
+		return err
+	}
+
```
rezbera, line 69: Refactor to separate function - no change in logic

config/config.go
```diff
@@ -76,6 +76,16 @@ func (c Config) GetEngine() *engineclient.Config {
 	return &c.Engine
 }
 
+// GetPayloadBuilder returns the block store configuration.
+func (c Config) GetPayloadBuilder() *builder.Config {
+	return &c.PayloadBuilder
+}
+
+// GetBlockStoreService returns the block store configuration.
+func (c Config) GetBlockStoreService() *blockstore.Config {
```
rezbera, line 85: Adding getters for consistency with other fields

testing/simulated/components.go
```diff
@@ -0,0 +1,79 @@
...
+		components.ProvideTrustedSetup,
+		components.ProvideValidatorService,
+		components.ProvideShutDownService,
+	}
+	c = append(c,
+		components.ProvideNodeAPIServer,
+		components.ProvideNodeAPIEngine,
+		components.ProvideNodeAPIBackend,
+	)
+	//
```
calbera, line 68: nit remove or comment

testing/simulated/eth-genesis.json
calbera, line 1: which accounts are pre-alloced? which contracts are pre-deployed? it should have at least EIP4788

testing/simulated/execution-clients.go
calbera, line 1: another nit: can `execution` be a package and inside you have `geth.go`, `reth.go`, etc.

testing/simulated/eth-genesis.json
rezbera: which has EIP4788 at 000F3df6D732807Ef1319fB7B8bB8522d0Beac02

testing/simulated/simulated_test.go
calbera, line 1: q: is the plan that all the test scenarios would go in this file? or we would just add in other files in package `simulated_test` ?

rezbera: In https://github.com/berachain/beacon-kit/pull/2523 , i've separated into different files

testing/simulated/execution/geth.go
```diff
@@ -0,0 +1,133 @@
...
+	require.NoError(t, err, "Geth container did not become ready in time")
+
+	return resource, authRPC
+}
+
+// ValidGethImage returns the default Docker image options for the Geth node.
+func ValidGethImage() docker.PullImageOptions {
+	return docker.PullImageOptions{
+		Repository: "ethereum/client-go",
+		Tag:        "latest",
```
calbera, line 131: are we okay with "latest" or should we stick to 1.14.x ?

rezbera: A reasonable question - atm in our e2e tests we use latest with fixed genesis so I'm happy with that. its a good way to get early indication of new releases and their issues

testing/simulated/components.go
```diff
@@ -0,0 +1,78 @@
...
+		components.ProvideTrustedSetup,
+		components.ProvideValidatorService,
+		components.ProvideShutDownService,
+	}
+	c = append(c,
+		components.ProvideNodeAPIServer,
+		components.ProvideNodeAPIEngine,
+		components.ProvideNodeAPIBackend,
+	)
+	c = append(c, components.ProvideNodeAPIHandlers,
```
abi87, line 68: nit: never understood why we append here instead of having a single list

testing/simulated/execution/geth.go
abi87: I also agree with tests being on latest. They should be our canary in the colamine for any potential issue (see Geth 1.15)

testing/simulated/homedir.go
```diff
@@ -0,0 +1,135 @@
...
+// the computed genesis validators root.
+func InitializeHomeDir(t *testing.T, tempHomeDir string) (*cmtcfg.Config, common.Root) {
+	t.Helper()
+
+	t.Logf("Initializing home directory: %s", tempHomeDir)
+	// Create the default CometBFT configuration using the temporary home directory.
+	cometConfig := createCometConfig(t, tempHomeDir)
+
+	// Retrieve the testnet chain specification.
+	chainSpec, err := spec.TestnetChainSpec()
```
abi87, line 55: Wonder if this should be Devnet instead?
It would give us freedom to immediately upgrade to the latest fork while developing.

testing/simulated/homedir.go
```diff
@@ -0,0 +1,135 @@
...
+	// Run initialization command to mimic 'beacond init'
+	initCommand(t, cometConfig.RootDir)
+
+	// Retrieve the BLS signer from the configured home directory.
+	blsSigner := GetBlsSigner(cometConfig.RootDir)
+
+	// Set the deposit amount to the maximum effective balance.
+	depositAmount := math.Gwei(chainSpec.MaxEffectiveBalance())
+	// Define an arbitrary withdrawal address.
+	withdrawalAddress := common.NewExecutionAddressFromHex("0x6Eb9C23e4c187452504Ef8c5fD8fA1a4b15BE162")
```
abi87, line 70: where this address come from? Do we control its sk?

testing/simulated/simcomet.go
```diff
@@ -0,0 +1,85 @@
...
+func ProvideSimComet(
+	logger *phuslu.Logger,
+	blockchain blockchain.BlockchainI,
+	blockBuilder validator.BlockBuilderI,
+	db dbm.DB,
+	cmtCfg *cmtcfg.Config,
+	appOpts config.AppOptions,
+	telemetrySink *metrics.TelemetrySink) *SimComet {
+	return &SimComet{
+		cometbft.NewService(
```
abi87, line 56: I guess this is the upside of instantiating the cometBFT node in Start instead of New (I always wondered why we do that)

testing/simulated/components.go
abi87: Also why don't we reuse prod one?
My guess is you need SimComet instead of comet.
Wonder if it's worth finding a way to reuse prod one and just replace comet?
Maybe not

testing/simulated/simulated_test.go
```diff
@@ -0,0 +1,266 @@
...
+	// Allow a short period for services to fully initialize.
+	time.Sleep(2 * time.Second)
+}
+
+// TearDownTest cleans up the test environment.
+func (s *SimulatedSuite) TearDownTest() {
+	if err := s.ElHandle.Close(); err != nil {
+		s.T().Error("Error closing EL handle:", err)
+	}
+	s.CancelFunc()
```
abi87, line 117: IIRC, CometBFT does not currently handle context cancellation.
This may be the one test where we are able to check that

testing/simulated/simulated_test.go
```diff
@@ -0,0 +1,266 @@
...
+		Txs:             proposal.Txs,
+		Height:          blockHeight,
+		ProposerAddress: pubkey.Address(),
+		Time:            proposalTime,
+	})
+	s.Require().NoError(err)
+	s.Require().Equal(types.PROCESS_PROPOSAL_STATUS_ACCEPT, processResp.Status)
+
+	// Wait for the block to be finalized.
+	time.Sleep(finalizeWaitDuration)
```
abi87, line 177: what is this sleep for? Can't we avoid it?

testing/simulated/simulated_test.go
```diff
@@ -0,0 +1,266 @@
...
+		Txs:             proposal.Txs,
+		Height:          blockHeight,
+		ProposerAddress: pubkey.Address(),
+		Time:            proposalTime,
+	})
+	s.Require().NoError(err)
+	s.Require().Equal(types.PROCESS_PROPOSAL_STATUS_REJECT, processResp.Status)
+
+	// Verify that the log contains the expected error message.
+	s.Require().Contains(s.LogBuffer.String(), errors.ErrInvalidPayloadStatus.Error())
```
abi87, line 263: Wow I never thought we could do assertion over logs!

abi87 (commented):
Made a first pass, so far so good!

testing/simulated/components.go
rezbera: > Wonder if it's worth finding a way to reuse prod one and just replace comet?
Probably not

Yea probably tricky. not sure either but I think there was some attempt at logical separation. Can update all of these along with `prod` one in a separate PR for consistency.

testing/simulated/homedir.go
rezbera: Good question. IIRC @calbera wanted to move all end-to-end tests to use testnet spec, so I kept consistent with that intent.

testing/simulated/homedir.go
rezbera: Yes, it's the last address we use in our devnets and we have the private key. happy to change this to something else as well.

testing/simulated/simulated_test.go
rezbera: It's not the cleanest way but I had no other ideas of how I could access the _reason_ for a rejection, which is important for testing and it works.

testing/simulated/simulated_test.go
rezbera: You are right. I previously encountered errors if I didn't wait a bit before finalize but upon re-testing, that is not the case. I have removed the sleep

testing/simulated/homedir.go
rezbera: Have updated the withdrawal address to one which has the pkey available in kurtosis

rezbera:
The genesis file has now been correctly updated to reflect that in https://github.com/berachain/beacon-kit/blob/1ef508ae5c5cca31aed2e6f42894a2ff838ed5ee/testing/files/eth-genesis.json

testing/simulated/simulated_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+// initializeChain sets up the chain using the genesis file.
+func (s *SimulatedSuite) initializeChain() {
+	// Load the genesis state.
+	appGenesis, err := genutiltypes.AppGenesisFromFile(s.HomeDir + "/config/genesis.json")
+	s.Require().NoError(err)
+
+	// Initialize the chain.
+	initResp, err := s.SimComet.Comet.InitChain(s.Ctx, &types.InitChainRequest{
+		ChainId:       "testnet-chain-80069",
```
shotes, line 126: nit: make this configurable by using chain spec. I'm not even sure if the "testnet" string is in the Testnet chain spec. Just throwing this out there, but not dying on this hill.

shotes (commented):
Just one nit, but loving this

testing/simulated/simulated_test.go
rezbera: The chainID here is the CL chainID which AFAIK is not in the ChainSpec. From what I've seen, the CL ChainID is only passed in via `client.toml`, or in our case, as part of the cosmos SDK context. 

However, it should be a constant which I have done here https://github.com/berachain/beacon-kit/pull/2411/commits/c0b9fbcd7885aa7759cf5a231ccb56196e2bd8e7

testing/simulated/homedir.go
calbera: would actually agree with using devnet here for all e2e. but we should try to adjust the devnet values to be consistent with those of mainnet

rezbera: Switching to `devnet` should be straightforward once we bring make devnet values consistent. Happy to do this in a follow up PR

testing/simulated/simulated_test.go
```diff
@@ -0,0 +1,261 @@
...
+	s.Require().NotNil(cometService)
+	s.SimComet = cometService
+
+	// Start the Beacon node in a separate goroutine.
+	go func() {
+		_ = s.TestNode.Start(s.Ctx)
+	}()
+
+	// Allow a short period for services to fully initialize.
+	time.Sleep(2 * time.Second)
```
abi87, line 107: q: is there an health end point we can hit here instead of just sleeping?
Does not have to be in this PR, just noting it down.
We don't implement yet the health beacon API, but maybe the cometBFT one can work?

testing/simulated/simulated_test.go
```diff
@@ -0,0 +1,261 @@
...
+	queryCtx, err := s.SimComet.CreateQueryContext(blockHeight, false)
+	s.Require().NoError(err)
+
+	stateDB := s.TestNode.StorageBackend.StateFromContext(queryCtx)
+	slot, err := stateDB.GetSlot()
+	s.Require().NoError(err)
+	s.Require().Equal(mathpkg.U64(blockHeight), slot)
+
+	stateHeader, err := stateDB.GetLatestBlockHeader()
+	s.Require().NoError(err)
```
abi87, line 197: q: can't we use some APIs to retrieve the header, similarly to what we do in e2e tests?

abi87 (approved):
Some questions and some cleanup to do, at least around node switch off, but this is a great start, so approving, in the interest of moving this forward

testing/simulated/utils.go
```diff
@@ -0,0 +1,224 @@
...
+	maliciousTxs := []*gethprimitives.Transaction{maliciousTx}
+
+	// Extract the current execution payload and compute the withdrawals hash.
+	payload := signedBeaconBlock.GetMessage().GetBody().ExecutionPayload
+	withdrawals := payload.GetWithdrawals()
+	withdrawalsHash := gethprimitives.DeriveSha(withdrawals, gethprimitives.NewStackTrie(nil))
+	parentRoot := signedBeaconBlock.GetMessage().GetParentBlockRoot()
+
+	// Construct a new execution block header and body using the malicious transactions.
+	executionBlock := gethprimitives.NewBlockWithHeader(
```
abi87, line 109: @rezbera I believe I have refactored out this code, so you can reuse `MakeEthBlock` here

PR #2412, opened 2025-01-25
chore(block-builder): a few minor cleanups

abi87:
This PR collects a few minor cleanup I collected while reviewing the block building code:

- Early return from `BuildBlockAndSidecars` in case block building is disabled (as it could be for full nodes)
- Consolidated checks over nil execution payload.
  - This required some changes to the `rpc.Client` to simplify the unit test that proves it 
- drop `remotePayloadBuilders` which are not really used at this point in time
- minor improvement to `app.toml` config


beacon/validator/block_builder.go
```diff
@@ -41,21 +41,19 @@ import (
...
-		blk       *ctypes.BeaconBlock
-		sidecars  datypes.BlobSidecars
-		forkData  *ctypes.ForkData
-	)
-
 	startTime := time.Now()
 	defer s.metrics.measureRequestBlockForProposalTime(startTime)
 
+	// payload builder is disabled, so nothing to do
+	if s.localPayloadBuilder.Enabled() {
```
abi87, line 53: a minor optimization: if block building is disabled (as it should be, e.g. in full nodes) we could return early here

beacon/validator/service.go
```diff
@@ -50,9 +50,6 @@ type Service struct {
 	// Building blocks are done by submitting forkchoice updates through.
 	// The local Builder.
 	localPayloadBuilder PayloadBuilder
-	// remotePayloadBuilders represents a list of remote block builders, these
-	// builders are connected to other execution clients via the EngineAPI.
-	remotePayloadBuilders []PayloadBuilder
```
abi87, line 55: not sure what is this for, but not really used. Dropped

config/template/template.go
```diff
@@ -65,6 +65,8 @@ implementation = "{{.BeaconKit.KZG.Implementation}}"
 
 [beacon-kit.payload-builder]
 # Enabled determines if the local payload builder is enabled.
+# It should be enabled for validators, but it can be disabled
```
abi87, line 68: hopefully this clarifies how this variable should be set

beacon/validator/block_builder.go
```diff
@@ -95,9 +93,6 @@ func (s *Service) BuildBlockAndSidecars(
 	if err != nil {
 		return nil, nil, err
 	}
-	if envelope == nil {
```
abi87, line 98: this check is not necessary since all code path leading here do intercept the condition `envelope == nil` and return an error

execution/client/ethclient/engine.go
```diff
@@ -142,17 +142,14 @@ func (s *Client) GetPayloadV3(
...
 		ExecutionPayload: t.Empty(version.Deneb),
 	}
 
-	if err := s.Call(
-		ctx, result, GetPayloadMethodV3, payloadID,
-	); err != nil {
-		return nil, err
-	}
-	return result, nil
+	return result, s.Call(ctx, result, GetPayloadMethodV3, payloadID)
```
abi87, line 152: nit for readability, don't feel strongly about this.
The only relevant point here is that ExecutionPayload is not nil

payload/builder/payload.go
```diff
@@ -173,18 +173,13 @@ func (pb *PayloadBuilder) RetrievePayload(
 	}
 
 	payload := envelope.GetExecutionPayload()
-	if !payload.IsNil() {
```
abi87, line 176: this check does not make sense. Note that we use payload below, without checking if is nil

payload/builder/payload.go
```diff
@@ -47,26 +47,27 @@ func (pb *PayloadBuilder) RequestPayloadAsync(
 	}
 
 	if payloadID, found := pb.pc.Get(slot, parentBlockRoot); found {
-		pb.logger.Warn(
```
abi87, line 50: not sure why this is warning, I don't think this is an error condition really?

abi87: it should be for validators since if optimistic build is enabled, still we should probably revert to info

beacon/validator/block_builder.go
```diff
@@ -41,21 +41,19 @@ import (
...
-		signedBlk *ctypes.SignedBeaconBlock
-		blk       *ctypes.BeaconBlock
-		sidecars  datypes.BlobSidecars
-		forkData  *ctypes.ForkData
-	)
-
 	startTime := time.Now()
 	defer s.metrics.measureRequestBlockForProposalTime(startTime)
 
+	if !s.localPayloadBuilder.Enabled() {
```
abi87, line 52: a minor optimization: if block building is disabled (as it should be, e.g. in full nodes) we could return early here

beacon/validator/block_builder.go
```diff
@@ -298,7 +291,7 @@ func (s *Service) buildBlockBody(
 	// Dequeue deposits from the state.
 	depositIndex, err := st.GetEth1DepositIndex()
 	if err != nil {
-		return ErrNilDepositIndexStart
```
abi87, line 301: we silence the error here, which I do not like. Improved the error handling

beacon/validator/errors.go
```diff
@@ -23,21 +23,12 @@ package validator
 import "github.com/berachain/beacon-kit/errors"
 
 var (
-
-	// ErrNilPayload is an error for when there is no payload
-	// in a beacon block.
-	ErrNilPayload = errors.New("nil payload in beacon block")
```
abi87, line 29: not used anymore since we do not check in validators anymore that payload is not nil.

beacon/validator/errors.go
```diff
@@ -23,21 +23,12 @@ package validator
...
-
 	// ErrNilBlkBody is an error for when the block body is nil.
 	ErrNilBlkBody = errors.New("nil block body")
 
 	// ErrNilBlobsBundle is an error for when the blobs bundle is nil.
 	ErrNilBlobsBundle = errors.New("nil blobs bundle")
 
-	// ErrNilDepositIndexStart is an error for when the deposit index start is
-	// nil.
-	ErrNilDepositIndexStart = errors.New("nil deposit index start")
```
abi87, line 39: dropped and used the real error returned by KvStore getter instead

execution/client/ethclient/engine_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	engineprimitives "github.com/berachain/beacon-kit/engine-primitives/engine-primitives"
+	"github.com/berachain/beacon-kit/execution/client/ethclient"
+	"github.com/berachain/beacon-kit/execution/client/ethclient/rpc"
+	"github.com/stretchr/testify/require"
+)
+
+// TestGetPayloadV3NeverReturnsEmptyPayload shows that execution payload
+// returned by ethClient is not nil.
+func TestGetPayloadV3NeverReturnsEmptyPayload(t *testing.T) {
```
abi87, line 35: this UT should give confidence that the payload from GetPayload call is never nil in current implementation

execution/client/ethclient/ethclient.go
```diff
@@ -26,11 +26,11 @@ import (
 
 // Client - Ethereum rpc client.
 type Client struct {
-	*rpc.Client
+	rpc.Client
```
abi87, line 29: I replaced the rpc.Client implementation with an interface to ease up unit testing.

execution/client/ethclient/rpc/client.go
```diff
@@ -32,9 +32,17 @@ import (
...
+type Client interface {
+	Start(context.Context)
+	Call(ctx context.Context, target any, method string, params ...any) error
+	Close() error
+}
+
+// client is an Ethereum RPC client that provides a
 // convenient way to interact with an Ethereum node.
-type Client struct {
+type client struct {
```
abi87, line 45: unexported as much as possible

execution/client/ethclient/rpc/client.go
```diff
@@ -67,18 +79,16 @@ func NewClient(url string, options ...func(rpc *Client)) *Client {
 				}
 			},
 		},
-		header: http.Header{"Content-Type": {"application/json"}},
-	}
-
-	for _, option := range options {
-		option(rpc)
+		jwtSecret:          secret,
+		jwtRefreshInterval: jwtRefreshInterval,
```
abi87, line 83: for a limited number of parameters I much prefer passing them explicitly in the New function rather than using options.
Options seems to be strictly related to the implementation of Client interface and I have found an easy way to mock them.

execution/client/ethclient/rpc/options.go
```diff
@@ -1,41 +0,0 @@
...
-package rpc
-
-import (
-	"time"
-
-	"github.com/berachain/beacon-kit/primitives/net/jwt"
-)
-
-// WithJWTSecret sets the JWT secret for the RPC client.
-func WithJWTSecret(secret *jwt.Secret) func(rpc *Client) {
```
abi87, line 30: We do use this approach with options in other parts of the codebase, but I personally don't really see much value to it.
We don't usually build structs in multiple ways, i.e. we don't generally have a struct specifying some options and not others.
If I am missing something, LMK

execution/client/client.go
```diff
@@ -64,17 +64,15 @@ func New(
...
 	return &EngineClient{
-		cfg:    cfg,
-		logger: logger,
-		Client: ethclient.New(
-			ethclientrpc.NewClient(
-				cfg.RPCDialURL.String(),
-				ethclientrpc.WithJWTSecret(jwtSecret),
-				ethclientrpc.WithJWTRefreshInterval(
-					cfg.RPCJWTRefreshInterval,
-				),
```
abi87, line 76: just moved the instantiation of `ethClient` up and replace the options with actual parameter in its `New` function

execution/client/ethclient/engine.go
```diff
@@ -138,21 +138,19 @@ func (s *Client) GetPayload(
...
+	payloadID engineprimitives.PayloadID,
 ) (ctypes.BuiltExecutionPayloadEnv, error) {
 	var t *ctypes.ExecutionPayload
 	result := &ctypes.ExecutionPayloadEnvelope[*engineprimitives.BlobsBundleV1[
-		eip4844.KZGCommitment, eip4844.KZGProof, eip4844.Blob,
+		eip4844.KZGCommitment,
+		eip4844.KZGProof,
+		eip4844.Blob,
 	]]{
 		ExecutionPayload: t.Empty(version.Deneb),
```
abi87, line 150: This is key point of this cleanup: ExecutionPayload is never nil, cause Empty is the default empty `ExecutionPayloadEnvelope` struct.
I added a UT to verify it

payload/builder/payload.go
```diff
@@ -166,29 +167,9 @@ func (pb *PayloadBuilder) RetrievePayload(
 		return nil, err
 	}
 
-	overrideBuilder := envelope.ShouldOverrideBuilder()
```
abi87, line 169: below is just stuff for logging. I consolidated it slightly

execution/client/ethclient/engine.go
fridrik01: I prefer how it was actually, think it was slightly more readable

execution/client/ethclient/ethclient.go
fridrik01: nice!

execution/client/ethclient/rpc/client.go
fridrik01: Agreed, I mainly find using options like this useful when creating an object using the builder pattern which doesn't imo apply here.

payload/builder/payload.go
```diff
@@ -47,26 +47,27 @@ func (pb *PayloadBuilder) RequestPayloadAsync(
...
+		parentBlockRoot,
+	)
 	if err != nil {
 		return nil, err
 	}
 
 	// Submit the forkchoice update to the execution client.
-	var payloadID *engineprimitives.PayloadID
-	payloadID, _, err = pb.ee.NotifyForkchoiceUpdate(
+	payloadID, _, err := pb.ee.NotifyForkchoiceUpdate(
```
fridrik01, line 70: I thought this would shadow the `err` here

fridrik01 (approved):
lgtm

shotes (approved):
lgtm

beacon/validator/block_builder.go
calbera: Q: do we know if PrepareProposal is ever called for a non-validator node?

calbera (approved):
Lgtm

PR #2413, opened 2025-01-25
fix(initchain): Enforce stricter handling on InitChain Requests

rezbera:
The original code for handling `[]ValidatorUpdate` was [copied](https://github.com/cosmos/cosmos-sdk/blob/e9b16a4eae13d5851eb12dd6eda77913c2951826/baseapp/abci.go#L116)  from CosmosSDK. In our case, the `ValidatorUpdate` list is ALWAYS expected to be 0 in initChain as we derive the validator set from the deposits directly in InitChain, which is the source of truth.

Therefore, we can remove the additional handling if the `len([]ValidatorUpdate) > 0` and simply enforce that it is always zero length, reducing confusion around its purpose

### DIFF
--- a/consensus/cometbft/service/init_chain.go
+++ b/consensus/cometbft/service/init_chain.go
@@ -21,10 +21,8 @@
 package cometbft
 
 import (
-	"bytes"
 	"context"
 	"fmt"
-	"sort"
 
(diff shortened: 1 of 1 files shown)


abi87 (approved):
lgtm

PR #2421, opened 2025-01-27
chore: Cleanup devtools.mk make targets

fridrik01:
Most of the make targets in devtools.mk are not relevant anymore or can be simplified now that we have a single go.mod/go.sum file


nidhi-singh02 (approved):
lgtm!

shotes (approved):
Straightforward. LGTM.

Do we need to change anything from the github CI side to make sure it doesn't try to run this target?

fridrik01:
> Straightforward. LGTM.
> 
> Do we need to change anything from the github CI side to make sure it doesn't try to run this target?

Not sure why the CI is waiting for tidy-sync-check, its not in the repo anywhere (including github workflow), will look at this tmr.

calbera (approved):
RIP to my fave `make repo-rinse` but all good

calbera:
Can we keep repo-rinse back actually? helpful for me, even as just reference for handling submodules

fridrik01:
> Can we keep repo-rinse back actually? helpful for me, even as just reference for handling submodules

You want to live on the edge, I would prefer to remove it as its not that relevant and can potentially wipe out all uncommitted data if run accidentally.. But sure whatever I can put it back in

PR #2422, opened 2025-01-27
feat: stable block time

melekes:


consensus/cometbft/service/options.go
```diff
@@ -81,3 +83,13 @@ func SetChainID[
...
 }
+
+// SetTargetBlockTime returns a Service option function that sets the desired
+// block time (e.g., 2s). Note that it CAN'T be lower than the minimal (floor)
+// block time in the network, which is comprised of the time to a) propose a
+// new block b) gather 2/3+ prevotes c) gather 2/3+ precommits.
+func SetTargetBlockTime[
+	LoggerT log.AdvancedLogger[LoggerT],
+](t time.Duration) func(*Service[LoggerT]) {
+	return func(bs *Service[LoggerT]) { bs.setTargetBlockTime(t) }
```
abi87, line 94: note that we do enforce some floor on timeouts in https://github.com/berachain/beacon-kit/blob/9097bf316facfdee7d6adc9461776550fd7c44af/consensus/cometbft/service/configs.go#L127

melekes: It is unclear to me how to do it from the API perspective. Do you have any suggestions @abi87?

consensus/cometbft/service/block_delay.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+// maxDelayBetweenBlocks is the maximum delay between two consequtive blocks.
+// If the last block time minus the previous block time is greater than
+// maxDelayBetweenBlocks, then we reset `FinalizeBlockResponse.NextBlockDelay`
+// to default.
+//
+// This is needed because the network may stall for a long time and we don't
+// want to rush in new blocks as the network resumes its operation.
+const maxDelayBetweenBlocks = 30 * time.Minute
```
melekes, line 37: What do people think is a good default here?

consensus/cometbft/service/block_delay.go
```diff
@@ -0,0 +1,106 @@
...
+
+	// Reset the initial time and height if the time between blocks is greater
+	// than maxDelayBetweenBlocks.
+	if curBlockTime.Sub(d.PreviousBlockTime) > maxDelayBetweenBlocks {
+		d.InitialTime = curBlockTime
+		d.InitialHeight = curBlockHeight - 1
+	}
+
+	t := d.InitialTime.Add(targetBlockTime * time.Duration(curBlockHeight-d.InitialHeight))
+	if curBlockTime.Before(t) {
```
melekes, line 89: Should we use real-time here?

```suggestion
	if now := time.Now(); now.Before(t) {
	   t.Sub(now)
```

consensus/cometbft/service/block_delay.go
```diff
@@ -0,0 +1,106 @@
...
+package cometbft
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/berachain/beacon-kit/primitives/encoding/json"
+)
+
+// maxDelayBetweenBlocks is the maximum delay between two consequtive blocks.
```
alesforz, line 30: small typo: "consecutive"

consensus/cometbft/service/service.go
```diff
@@ -147,6 +154,15 @@ func NewService[
...
 	}
 
+	// Load block delay
+	bz, err := s.sm.LoadBlockDelay()
+	if err != nil {
+		panic(fmt.Errorf("failed loading block delay: %w", err))
+	}
+	if bz != nil {
+		s.blockDelay = blockDelayFromBytes(bz)
+	}
```
alesforz, line 164: If `bz` is `nil`, are we initializing the block delay to some default? 
I see that `initChain` will call `blockDelayUponGenesis()` to set it, to it might be enough.

Is there a scenario in which the block delay is read between the calls to`NewService` and `initChain`?

alesforz (approved):

consensus/cometbft/service/service.go
melekes: Good question. The logic here is that if `height > 0`, we load everything from DB; if `height == 0`, we construct everything from genesis.

`NewService` gets called first (that's why there's a check for `nil`). `InitChain` is only called if `height == 0` (in other cases, CometBFT doesn't call `InitChain`).

For cases where `height > 0`, but DB is not found, or the record is not found, we should probably panic... I need to think about it more. This could be a typical scenario where the chain is live, but you didn't enable this feature.

melekes: [`41364a7` (#2422)](https://github.com/berachain/beacon-kit/pull/2422/commits/41364a7c6e943b0201fcb51b63030d9212dfe2b1)

this needs to be tested

consensus/cometbft/service/block_delay.go
abi87: @berachain/core originally I was leaning towards 2 hours, some kind of incident resolution time. But it probably does not make sense to go that large. 30 mins seems all right.

melekes:
Do we need to enforce a minimum for `NextBlockDelay` https://github.com/berachain/beacon-kit/pull/2422#discussion_r1931152171?

consensus/cometbft/service/service.go
```diff
@@ -96,6 +97,12 @@ type Service struct {
 	minRetainBlocks uint64
 
 	chainID string
+
+	// calculates block delay for the next block
```
fridrik01, line 101: I feel like we should callout here in a comment at least that the pointer may be `nil` until `FinalizeBlock` initializes it (in case `height > 0` and the key was not in db)

node-core/builder/baseapp_options.go
```diff
@@ -65,6 +66,12 @@ func DefaultServiceOptions(
 		}
 	}
 
+	sbtUpgradeHeight := cast.ToInt64(appOpts.Get(beaconflags.SBTUpgradeHeight))
+	if sbtUpgradeHeight == 0 {
+		panic("--beacon-kit.cometbft.sbt-upgrade-height (height where SBT is enabled) must be set")
```
melekes, line 71: Should we force validators to use this flag?

consensus/cometbft/service/options.go
melekes: It's now a const, so enforcing the minimum is unnecessary.

consensus/cometbft/service/block_delay.go
```diff
@@ -0,0 +1,137 @@
...
+func blockDelayUponGenesis(genesisTime time.Time, initialHeight int64) *blockDelay {
+	return &blockDelay{
+		InitialTime:       genesisTime,
+		InitialHeight:     initialHeight,
+		PreviousBlockTime: genesisTime,
+	}
+}
+
+// Panics if it fails to read from the buffer.
+func blockDelayFromBytes(
```
alesforz, line 63: Here I suggest to add to the docs the format that the function expects for the input `[]byte`. 
From the code, I see it must be 24 bytes long, and it's the concatenation of `[initialTime|initialHeight|prevBlockTime]` in little endian, otherwise it panics.

consensus/cometbft/service/block_delay.go
```diff
@@ -0,0 +1,137 @@
...
+	// Update the previous block time.
+	d.PreviousBlockTime = curBlockTime
+
+	return noDelay
+}
+
+// ToBytes converts the blockDelay to bytes.
+//
+// Panics if it fails to write to the buffer.
+func (d *blockDelay) ToBytes() []byte {
```
alesforz, line 120: I suggest to add to the docs the format of the output `[]byte`. 
That is, it's a slice concatenating `InitialTime`, `InitialHeight`, and `PreviousBlockTime` in little endian.

consensus/cometbft/service/options.go
```diff
@@ -68,3 +70,12 @@ func SetInterBlockCache(cache storetypes.MultiStorePersistentCache) func(*Servic
 func SetChainID(chainID string) func(*Service) {
 	return func(s *Service) { s.chainID = chainID }
 }
+
+// SetInterBlockCache provides a Service option function that sets the stable
```
alesforz, line 74: do you mean `SetSBTUpgradeHeightAndTime`?

alesforz (approved):
Left a couple of comments about functions' docs 😄 
Looks good 👍

abi87 (commented):
some commets, review still ongoing

consensus/cometbft/service/block_delay.go
```diff
@@ -0,0 +1,154 @@
...
+package cometbft
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"time"
+)
+
+// MaxDelayBetweenBlocks is the maximum delay between two consecutive blocks.
```
abi87, line 30: nit: fauly patch from me. This comment is meant to go below `const (`

consensus/cometbft/service/block_delay.go
```diff
@@ -0,0 +1,149 @@
...
+	// Reset the initial time and height if the time between blocks is greater
+	// than maxDelayBetweenBlocks.
+	if curBlockTime.Sub(d.PreviousBlockTime) > maxDelayBetweenBlocks {
+		d.InitialTime = curBlockTime
+		d.InitialHeight = curBlockHeight - 1
+	}
+
+	t := d.InitialTime.Add(targetBlockTime * time.Duration(curBlockHeight-d.InitialHeight))
+	if curBlockTime.Before(t) {
+		return t.Sub(curBlockTime)
```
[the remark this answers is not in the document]
abi87: Changed in one of my patches

consensus/cometbft/service/finalize_block.go
```diff
@@ -94,9 +98,66 @@ func (s *Service) finalizeBlockInternal(
...
+			if s.sbtUpgradeTime.IsZero() {
+				panic("SBT (stable block time) upgrade detected. --beacon-kit.cometbft.sbt-upgrade-time (block's time where SBT was enabled) must be set")
+			}
+			s.blockDelay = BlockDelayUponGenesis(
+				s.sbtUpgradeTime,
+				s.sbtUpgradeHeight,
+			)
+			delay = s.blockDelay.Next(req.Time, req.Height)
+		case req.Height == s.sbtUpgradeHeight: // upgrade is happening now
+			s.blockDelay = BlockDelayUponGenesis(
```
abi87, line 127: so:
- if node starts from scratch it will have s.BlockDelay initialized with genesis height and time WHILE
- if node enters this case it will pick req.Time and req.Height

Is this correct? Unless we are guaranteed that req.Time - genesis.Time = req.Height * targetBlockTime I believe these two nodes would end up returning different delays

abi87: If I am correct a solution may be to not initialising blockDelay at genesis

consensus/cometbft/service/finalize_block.go
```diff
@@ -94,9 +98,66 @@ func (s *Service) finalizeBlockInternal(
...
+				req.Time,
```
abi87, line 128: Also wondering if, operationally, it would be smarted here to use `s.sbtUpgradeTime` instead of `req.Time`. Note that in this scenario `req.Height == s.sbtUpgradeHeight`

consensus/cometbft/service/finalize_block.go
```diff
@@ -94,9 +98,66 @@ func (s *Service) finalizeBlockInternal(
...
+				req.Height,
+			)
+			// default: // upgrade is in the future
```
abi87, line 153: I believe if we hit this default, we are picking `delay = s.blockDelay.Next(req.Time, req.Height)` as calculated in line 136 instead of `constBlockDelay`. This is different from what we do in the `s.blockDelay == nil` case where pre fork we use the constDelay.
Wouldn't this result in different delays being used?

consensus/cometbft/service/finalize_block.go
```diff
@@ -94,9 +98,66 @@ func (s *Service) finalizeBlockInternal(
...
+			if s.sbtUpgradeTime.IsZero() {
+				panic("SBT (stable block time) upgrade detected. --beacon-kit.cometbft.sbt-upgrade-time (block's time where SBT was enabled) must be set")
+			}
+			s.blockDelay = BlockDelayUponGenesis(
+				s.sbtUpgradeTime,
+				s.sbtUpgradeHeight,
+			)
+			delay = s.blockDelay.Next(req.Time, req.Height)
+		case req.Height == s.sbtUpgradeHeight: // upgrade is happening now
+			s.blockDelay = BlockDelayUponGenesis(
```
abi87, line 149: unlike the case `s.blockDelay == nil` here we would return `delay = s.blockDelay.Next(req.Time, req.Height)` instead of constBlockDelay right? 

Wouldn't this result in different delays being used?

consensus/cometbft/service/block_delay.go
```diff
@@ -0,0 +1,149 @@
...
+	// `FinalizeBlockResponse.NextBlockDelay` can't be exactly 0. If it's set to
+	// 0, then `timeout_commit` from the config will be used, which is not what
+	// we want since we're trying to control the block time.
+	const noDelay = 1 * time.Microsecond
+
+	// Reset the initial time and height if the time between blocks is greater
+	// than maxDelayBetweenBlocks.
+	if curBlockTime.Sub(d.PreviousBlockTime) > maxDelayBetweenBlocks {
+		d.InitialTime = curBlockTime
+		d.InitialHeight = curBlockHeight - 1
```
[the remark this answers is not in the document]
jmalicevic: I think the answer to this question might lie in line 122 where we substract the current height from the InitialHeight. And InitialHeight has to be set to something before this computation occurs.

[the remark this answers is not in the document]
abi87: @jmalicevic I understand, I just wonder why it isn't initalized to `curBlockHeight`?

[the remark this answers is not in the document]
jmalicevic: So if my understanding is right, in this case (when the delay between blocks is so large that it does not make sense to compute based off it), we reset everything and treat this as if the upgrade happened just now?  
In that case, I don't know why we do the computations below (L122) rather than just returning the default delay and resetting everything here to the current height and time and return the default delay? 

Unless we do want to take the last height into account (the delay between the last height and this height) and we set it all to last height. But even in that case `PrevBlockTime` should represent the time of the last height (when this was last computed), so L123 should never really be true and we are again back to the case of returning the default delay. 

Setting this to `currBlockHeight - 1` would make sense if we did not touch `PreviousBlockTime` so we consider `currBlockHeight -1` to be the initial height and time based off which we want to continue our algorithm. But I don't think this is the case. 

@melekes please correct me if I am wrong.

consensus/cometbft/service/block_delay.go
```diff
@@ -0,0 +1,154 @@
...
+func (d *BlockDelay) Next(curBlockTime time.Time, curBlockHeight int64) time.Duration {
+	// Until `timeout_commit` is removed from the CometBFT config,
+	// `FinalizeBlockResponse.NextBlockDelay` can't be exactly 0. If it's set to
+	// 0, then `timeout_commit` from the config will be used, which is not what
+	// we want since we're trying to control the block time.
+	const noDelay = 1 * time.Microsecond
+
+	// Reset the initial time and height if the time between blocks is greater
+	// than maxDelayBetweenBlocks.
+	if curBlockTime.Sub(d.PreviousBlockTime) > MaxDelayBetweenBlocks {
```
jmalicevic, line 116: There is not a case where `Next` is *not* called on every height? If there is such a case at some point in time, then we will need to account for `PreviousBlockHeight` as well. Maybe leaving a comment would here to remember.

consensus/cometbft/service/block_delay.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	err = binary.Read(buf, binary.LittleEndian, &prevBlockTime)
+	if err != nil {
+		panic(fmt.Sprintf("failed to read PreviousBlockTime: %v", err))
+	}
+
+	return &blockDelay{
+		InitialTime:       time.Unix(0, initialTime),
+		InitialHeight:     initialHeight,
+		PreviousBlockTime: time.Unix(0, prevBlockTime),
```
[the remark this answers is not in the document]
melekes: Depends on whenever we need nanosecond precision. block.Time is `time.Time` - "A Time represents an instant in time with nanosecond precision." So if we want to be precise here, we need to preserve milliseconds and nanoseconds.

consensus/cometbft/service/block_delay.go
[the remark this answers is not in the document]
melekes: > we should update it at every block

yes!

consensus/cometbft/service/block_delay.go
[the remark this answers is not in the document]
melekes: [`2fd27f2` (#2422)](https://github.com/berachain/beacon-kit/pull/2422/commits/2fd27f21831036db4f44f6670bcee625f06f6412)

consensus/cometbft/service/block_delay.go
melekes: > There is not a case where Next is not called on every height?

No, it's not how the new `NextBlockDelay` feature works. The ABCI application is always expected to provide value.

melekes: Please feel free to add a comment, but beware of adding a new source of confusion for developers reading the code.

consensus/cometbft/service/finalize_block.go
melekes: > unlike the case s.blockDelay == nil here we would return delay = s.blockDelay.Next(req.Time, req.Height) instead of constBlockDelay right?

Right. 

Before the first upgrade, we always use `constBlockDelay = 500ms` (before `s.sbtUpgradeHeight + 1` is reached). Then it's whatever you set `TargetBlockTime` to (e.g. 2s). 

If you change the target block time again, the subsequent block delay will be calculated using the previous checkpoint (up until `s.sbtUpgradeHeight + 1`).

melekes: Thinking about this more, we should probably save `TargetBlockTime` to the db as well.

consensus/cometbft/service/finalize_block.go
melekes: answered here https://github.com/berachain/beacon-kit/pull/2422#discussion_r2031626507

consensus/cometbft/service/finalize_block.go
melekes: Side note: If nodes provide different values for `s. sbtUpgradeHeight `, they will have different delays. You have to ensure consistent values across the validators.

Now, to your question:

There are two possible scenarios.

1. Node 1 starts from genesis (no snapshot) and catches up to the rest of the network. In that case, the delay is ignored.
  a) If Node 1 has `s.sbtUpgradeHeight`, it will switch to the desired height and time as a checkpoint.
  b) If Node 1 doesn't set `s.sbtUpgradeHeight`, it will continue using the genesis time and height.

2. Node 1 starts from the genesis (as does most of the network). Node 2 joins in later and provides `s.sbtUpgradeHeight` that differs from the genesis time and height. Thus, the resulting delays will diverge. But this would be a misconfiguration, IMHO.

**Anyway, this becomes way too complex, and probably the only way to reduce the resulting complexity is to add a new consensus parameter (similar to `vote extensions enable height`), which will guarantee consistency across the whole network.**

consensus/cometbft/service/state/state.go
```diff
@@ -91,3 +91,19 @@ func (sm *Manager) Close() error {
...
 	return sm.cms
 }
+
+var blockDelayKey = []byte("blockDelay")
+
+// LoadBlockDelay loads the block delay bytes from the database.
+//
+// see next_block_delay.go
+func (sm *Manager) LoadBlockDelay() ([]byte, error) {
+	return sm.db.Get(blockDelayKey)
```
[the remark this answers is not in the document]
melekes: > does this mean these values is not accounted for in the appHash?

correct. I don't think it should be part of the app state as it's a configuration parameter, which regulates the speed of block creation.

> how do we ensure that keys spaces of commit multi store and blockDelayKey are separate?

good question. by using a unique prefix + checking if key exists before creating it?

> unlike commitMultistore, blockDelay value is persisted as soon as Set is called, without any Commit step, right?

right. we should fix it.

consensus/cometbft/service/finalize_block.go
abi87: Practically speaking, aren't we forced to provide a reference point different from genesis height/time? I would have to recheck but I believe we have been consistently below 2 second (even just slightly) and that time different gets amplified while the chain keeps going.

consensus/cometbft/service/block_delay.go
jmalicevic: Yes you are right, it might add more harm than value. We can add a comment n the function saying it is expected that this is called at every height and leave it at that. No need to talk abut potential new parameters :D

consensus/cometbft/service/block_delay.go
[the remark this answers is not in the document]
melekes: fixed in [`0a15f45` (#2422)](https://github.com/berachain/beacon-kit/pull/2422/commits/0a15f451374f64affa141b2bd4412fa9b866f1e7)

consensus/cometbft/service/finalize_block.go
melekes: > Practically speaking, aren't we forced to provide a reference point different from genesis height/time?

Yes, it can be enforced through the CLI, but consensus parameters ensure consistency across the whole network, while the CLI does not give you any guarantees.

consensus/cometbft/service/finalize_block.go
```diff
@@ -92,14 +100,65 @@ func (s *Service) finalizeBlockInternal(
...
+		s.blockDelay = &BlockDelay{
+			InitialTime:       req.Time,
+			InitialHeight:     req.Height,
+			PreviousBlockTime: req.Time,
+		}
+		return constBlockDelay
+	}
+
+	// c3. current height > SBTEnableHeight
+	if !(req.Height > s.cmtConsensusParams.Feature.SBTEnableHeight) {
```
abi87, line 138: we return in all cases above. Isn't this if redundant?

consensus/cometbft/service/finalize_block.go
```diff
@@ -92,14 +100,65 @@ func (s *Service) finalizeBlockInternal(
...
+		// Looks like we've skipped SBTEnableHeight (probably restoring from the
+		// snapshot) => reinitialize the block delay. Treat the current block as a
+		// checkpoint.
+		s.blockDelay = &BlockDelay{
+			InitialTime:       req.Time,
+			InitialHeight:     req.Height,
+			PreviousBlockTime: req.Time,
+		}
+		return constBlockDelay
+	}
```
abi87, line 159: Can this case even happen?
Here`req.Height > s.cmtConsensusParams.Feature.SBTEnableHeight` so any node which has duly processed the activation height has `blockDelay` duly stored. If we are restoring the state, must we not find `blockDelay` duly stored?

consensus/cometbft/service/finalize_block.go
```diff
@@ -23,11 +23,19 @@ package cometbft
...
 	cmtabci "github.com/cometbft/cometbft/abci/types"
 	"github.com/sourcegraph/conc/iter"
 )
 
+// Delay to use before the upgrade to SBT.
+const constBlockDelay = 1 * time.Second
+
+// Height to enable SBT. Changes will be applied in the next block after this
+// (10_000_001).
+const sbtEnableHeight = 10_000_000
```
jmalicevic, line 37: Have different heights for consensus parameters updates and sbtEnableHeight
```suggestion
const sbtEnableHeight = 10_000_000
const sbtConsensusParamUpdateHeight = 10_00_0
```

consensus/cometbft/service/finalize_block.go
```diff
@@ -23,11 +23,20 @@ package cometbft
...
 	"context"
 	"fmt"
+	"time"
 
 	cmtabci "github.com/cometbft/cometbft/abci/types"
 	"github.com/sourcegraph/conc/iter"
 )
 
+// Delay to use before the upgrade to SBT.
+const constBlockDelay = 1 * time.Second
```
abi87, line 33: it's 500 ms in our case

consensus/cometbft/service/finalize_block.go
melekes: [`404e20b` (#2422)](https://github.com/berachain/beacon-kit/pull/2422/commits/404e20b4a64c5f7a646d0098a56db8d55fc2d67c)

alesforz (approved):

consensus/cometbft/service/delay/delay.go
```diff
@@ -0,0 +1,137 @@
...
+	// than MaxDelayBetweenBlocks. This makes the current time and height the
+	// initial one as if the upgrade happened just now.
+	if curBlockTime.Sub(d.PreviousBlockTime) > cfg.SbtMaxBlockDelay() {
+		d.InitialTime = curBlockTime
+		d.InitialHeight = curBlockHeight
+		d.PreviousBlockTime = curBlockTime
+		return cfg.SbtTargetBlockTime()
+	}
+
+	d.PreviousBlockTime = curBlockTime
```
fridrik01, line 62: should we validate that `curBlockTime > d.PreviousBlockTime` and handle that appropriately if that is not the case, or can we assume its monotonically increasing?

consensus/cometbft/service/delay/delay.go
```diff
@@ -0,0 +1,137 @@
...
+// FromBytes converts the bytes to a blockDelay.
+//
+// Expected format:
+//
+//	InitialTime (int64) | InitialHeight (int64) | PreviousBlockTime (int64)
+//	(little endian)
+//
+// It returns ErrBlockDelayFromBytes if it fails to read from the buffer.
+func FromBytes(bz []byte) (*BlockDelay, error) {
+	buf := bytes.NewReader(bz)
```
fridrik01, line 81: consider adding a check for valid byte length, like:

```
if len(bz) != 24 {
    return nil, fmt.Errorf("invalid BlockDelay buffer size: %d", len(bz))
}
```

chain/spec.go
```diff
@@ -272,6 +275,22 @@ func (s spec) validate() error {
 	return nil
 }
 
+func (s spec) SbtMaxBlockDelay() time.Duration {
```
fridrik01, line 295: consider adding a check here in `spec.validate()` which checks that `ConsensusUpdateHeight < ConsensusEnableHeight`, or `ConsensusUpdateHeight <= ConsensusEnableHeight` if we are sure they may be done at same height (when testing this in our devnet we had an issue when update and enable height were equal).

consensus/cometbft/service/delay/delay.go
```diff
@@ -0,0 +1,137 @@
...
+	// InitialHeight is a checkpoint in blockchain's height from which we start
+	// calculating the next block time.
+	InitialHeight int64
+
+	// PreviousBlockTime is the time of the previous block.
+	PreviousBlockTime time.Time
+}
+
+// Next returns the duration to wait before proposing the next block.
+func (d *BlockDelay) Next(cfg ConfigGetter, curBlockTime time.Time, curBlockHeight int64) time.Duration {
```
fridrik01: nit: consider renaming `Next()` as it implies to me that it does not mutate state which it does,  something like `ComputeNext()` or `UpdateAndGetNext()`.

consensus/cometbft/service/service.go
```diff
@@ -105,13 +109,19 @@ type Service struct {
 	// TODO: We must use this as a workaround for now until CometBFT properly
 	// generates contexts that inherit from the parent context we provide.
 	ctx context.Context
+
+	// calculates block delay for the next block
+	//
+	// NOTE: may be nil until either InitChain or FinalizeBlock is called.
+	blockDelay *delay.BlockDelay
```
fridrik01, line 116: just to confirm, access to `blockDelay` does not need synchronization (almost 100% sure it doesn't due to comebft but would be good to confirm)?

consensus/cometbft/service/delay/delay.go
abi87: Seems an overkill to me, at least for the time being. This data is never sent over the wire to other nodes nor it is received from any of them. Everything is local, so this kind of validation seems unnecessary

consensus/cometbft/service/delay/delay.go
abi87: Block times of Finalized blocks are guaranteed to be strictly increasing if the microseconds are taken into account, which is what we do here I believe

consensus/cometbft/service/service.go
abi87: I believe so. This attribute is used only in commit and finalize (which are guaranteed not to be concurrently called). Also it is not called by any query method

consensus/cometbft/service/delay/delay.go
fridrik01: That's fair

README.md
```diff
@@ -30,14 +30,10 @@ The framework offers the most user-friendly way to build and operate an EVM bloc
...
+BeaconKit supports the following execution clients:
 
-- [**Geth**](https://geth.ethereum.org/): Official Go implementation of the Ethereum protocol.
-- [**Erigon**](https://erigon.tech/): More performant, feature-rich client forked from `go-ethereum`.
-- [**Nethermind**](https://www.nethermind.io/): .NET based client with full support for Ethereum protocols.
-- [**Besu**](https://www.lfdecentralizedtrust.org/projects/besu): Enterprise-grade client, Apache 2.0 licensed, written in Java.
-- [**Reth**](https://reth.rs/): Rust-based client focusing on performance and reliability.
-- [**Ethereumjs**](https://ethereumjs.readthedocs.io/en/latest/#): Javascript based client managed by the Ethereum Foundation.
+- [**Bera-Geth**](https://github.com/berachain/bera-geth): Official Go implementation of the Berachain protocol.
+- [**Bera-Reth**](https://github.com/berachain/bera-reth): Rust-based client focusing on performance and reliability.
```
fridrik01: nit: [latest patch](https://github.com/berachain/beacon-kit/pull/2422/commits/721f7be12de84843024dc83c785918192d27f329) adds this and other unrelated changes to this PR which would be nice to cleanup

abi87: Solved, it was merge artifact

abi87 (approved):
finally LGTM

PR #2423, opened 2025-01-27
[4/n] refactor(chain-spec): handling values for planned fork upgrades

calbera:
## Changes
- Convert all "DenebPlus" usage to Deneb1 
- Remove base spec (each spec should explicitly show all values) and set defaults constants
- Introduce concept of further hard forks for chain spec value changes 
- Remove unused chain specs and set the following:
  - mainnet
  - public testnet: calling it dartio?
  - devnet

## Next PRs
- [5/n](https://github.com/berachain/beacon-kit/pull/2424): chain spec -> fork version spec changes
- [6/n](https://github.com/berachain/beacon-kit/pull/2425): Deneb 1 values
- revive [WIP 7/n](https://github.com/berachain/beacon-kit-internal/pull/88): remove usage of `CHAIN_SPEC` and allow configuring chain spec values in a file


primitives/version/versions.go
```diff
@@ -20,28 +20,23 @@
...
+	Altair uint32 = 1
+	// Bellatrix is the second hardfork of the Beacon Chain.
+	Bellatrix uint32 = 2
+	// Capella is the third hardfork of the Beacon Chain.
+	Capella uint32 = 3
+	// Deneb is the first version of Deneb, used for genesis of Berachain mainnet.
+	Deneb uint32 = 4
+	// Deneb1 is the first hardfork of Deneb on Berachain mainnet (TBD if used).
+	// There may also be Deneb2, Deneb3, etc. hardforks.
+	Deneb1 uint32 = 260
```
abi87, line 36: So technically this is a breaking change since we are reordering the forks. It should be fine since DenebPlus was never active.
@calbera can you confirm this?

abi87 (commented):
publishing some comments, stil reviewing this

primitives/version/versions.go
calbera: correct

beacon/validator/block_builder.go
```diff
@@ -336,18 +331,9 @@ func (s *Service) buildBlockBody(
 	// Fill in unused field with non-nil value
 	body.SetSyncAggregate(&ctypes.SyncAggregate{})
 
-	// Get the epoch to find the active fork version.
-	epoch := s.chainSpec.SlotToEpoch(blk.GetSlot())
-	activeForkVersion := s.chainSpec.ActiveForkVersionForEpoch(
-		epoch,
-	)
-	if activeForkVersion >= version.DenebPlus {
```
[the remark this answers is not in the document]
calbera: Yeah exactly, this codepath was never triggered so just removing it.

testing/state-transition/state-transition.go
```diff
@@ -135,6 +135,7 @@ func SetupTestState(t *testing.T, cs chain.Spec) (
 	)
 
 	ctx := &transition.Context{
+		Context:                 context.Background(),
```
abi87, line 138: I am surprised we missed this. IIRC UTs fail without this

beacon/validator/block_builder.go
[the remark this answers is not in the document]
calbera: @shotes would like your confirmation here as I believe you added these unused fields.

testing/state-transition/state-transition.go
calbera: i was surprised too. It may be the unit tests aren't getting triggered in CI? I had to manually specify the `tags = test` to get them to run locally.

state-transition/core/validation_deposits_test.go
```diff
@@ -263,8 +261,7 @@ func TestLocalDepositsExceedBlockDeposits(t *testing.T) {
 }
 
 func TestLocalDepositsExceedBlockDepositsBadRoot(t *testing.T) {
-	csData := spec.BaseSpec()
-	csData.DepositEth1ChainID = spec.BoonetEth1ChainID
+	csData := defaultSpecData()
```
abi87, line 264: q: should we use Testnet or devnet values instead of yet another test config?

state-transition/core/core_test.go
```diff
@@ -159,3 +160,70 @@ func moveToEndOfEpoch(
 	}
 	return blk
 }
+
+// defaultSpecData returns a modifiable chain.SpecData before use as chain.Spec.
+func defaultSpecData() *chain.SpecData {
```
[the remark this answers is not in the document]
abi87: q: Can't we pick mainnet or devnet chain spec and modify relevant values only there?

config/spec/mainnet.go
```diff
@@ -22,60 +22,119 @@ package spec
...
-	// Target for block time is 2 seconds on Berachain mainnet.
-	mainnetSpec.TargetSecondsPerEth1Block = 2
-
+const (
 	// BGT contract address.
 	//
 	// A hard fork will occur to set this value as the BGT contract address
 	// when BGT beings to be minted.
-	mainnetSpec.EVMInflationAddress = common.ExecutionAddress{}
+	MainnetEVMInflationAddressDeneb = DefaultEVMInflationAddress
```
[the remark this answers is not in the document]
calbera: I am okay using the predefined specs in UTs but 1) some UTs may require changing values to test special conditions and 2) I still think there's value to having defaults that other specs can pull from.

config/spec/mainnet.go
```diff
@@ -22,60 +22,119 @@ package spec
...
-	mainnetSpec.MaxValidatorsPerWithdrawalsSweep = 31
+	MainnetMaxValidatorsPerWithdrawalsSweep = 31
 
-	// MaxEffectiveBalance (or max stake) is 10 million BERA.
-	mainnetSpec.MaxEffectiveBalance = 10_000_000 * 1e9
+	// MainnetMaxEffectiveBalance is the max stake of 10 million BERA for genesis version Deneb.
+	MainnetMaxEffectiveBalanceDeneb = 10_000_000 * 1e9
 
-	// Effective balance increment is 10k BERA
+	// MainnetEffectiveBalanceIncrementDeneb is 10k BERA for genesis version Deneb
```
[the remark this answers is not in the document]
calbera: can remove Deneb suffix from all values set at genesis. since genesis version is implied to be Deneb

config/spec/defaults.go
```diff
@@ -0,0 +1,88 @@
...
+// TITLE.
+
+package spec
+
+// NOTE: Most of these default values are taken from ETH2.0 spec.
+// Some values (mentioned below) are modified to better suit Berachain's system.
+
+const (
+	// Gwei value constants.
+	DefaultMaxEffectiveBalance       = 32e9
```
shotes, line 28: I think these `const`s should be made into private vars (lowercasing the first letter). This will discourage the use of these values directly, as they could be different than what is in the active spec.

We want to prevent a situation where a dev imports this package and uses these const values directly, instead of pulling them from the active spec.

primitives/version/version_test.go
```diff
@@ -1,151 +0,0 @@
...
-
-import (
-	"encoding/binary"
-	"testing"
-
-	"github.com/berachain/beacon-kit/primitives/version"
-	"github.com/stretchr/testify/require"
-)
-
-func TestFromUint32(t *testing.T) {
```
abi87, line 31: @calbera looks like these tests were moved to bytes package. Can you confirm?

state-transition/core/validation_deposits_test.go
calbera: most UTs are using mainnet spec. This is required if modifications are necessary to test

primitives/version/version_test.go
calbera: confirmed

abi87: Also, why we moved this code to bytes? I don't feel strongly about this but I think usually we err towards adding a new type (even if it's just an alias) and adding there its own methods.

primitives/bytes/b8.go
```diff
@@ -18,7 +18,7 @@
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, NON-INFRINGEMENT, AND
 // TITLE.
 //
-//nolint:dupl // it's okay.
```
abi87, line 21: nice

da/types/sidecar_test.go
```diff
@@ -133,8 +132,7 @@ type InclusionSink struct{}
 func (is InclusionSink) MeasureSince(_ string, _ time.Time, _ ...string) {}
 
 func TestHasValidInclusionProof(t *testing.T) {
-	specVals := spec2.BaseSpec()
-	spec, err := chain.NewSpec(specVals)
+	spec, err := spec.DevnetChainSpec()
```
abi87, line 135: @calbera @shotes IDK how we feel about this, but should we err towards using mainnet values for testing?
We do have other chain specs as well, but in general I wonder if we should test over something as close as possible to mainnet?
This is fine here, we can definitely move discussions to a different PR

consensus-types/types/payload.go
```diff
@@ -561,14 +561,11 @@ func (p *ExecutionPayload) GetExcessBlobGas() math.U64 {
...
 // ToHeader converts the ExecutionPayload to an ExecutionPayloadHeader.
-func (p *ExecutionPayload) ToHeader() (
-	*ExecutionPayloadHeader,
-	error,
-) {
+func (p *ExecutionPayload) ToHeader() (*ExecutionPayloadHeader, error) {
 	txsRoot := p.GetTransactions().HashTreeRoot()
 
 	switch p.Version() {
-	case version.Deneb, version.DenebPlus:
```
abi87, line 571: @calbera again idea here is that we have never used Deneb1 so this can go?

consensus-types/types/body.go
```diff
@@ -313,15 +313,3 @@ func (b *BeaconBlockBody) GetBlobKzgCommitments() eip4844.KZGCommitments[common.
 func (b *BeaconBlockBody) SetBlobKzgCommitments(commitments eip4844.KZGCommitments[common.ExecutionHash]) {
 	b.BlobKzgCommitments = commitments
 }
-
-// SetAttestationData is not implemented.
-// TODO: remove support for this with version handling cleanup.
-func (b *BeaconBlockBody) SetAttestationData(_ []*AttestationData) {
-	panic("not implemented")
```
abi87, line 320: if anything this confirms that we have never used this. Nice

da/types/sidecar_test.go
shotes: I prefer mainnet values as the default and all testing to happen around that.

consensus-types/types/payload.go
calbera: correct

config/spec/mainnet.go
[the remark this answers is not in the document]
abi87: I am just think it'd be nice if out of this package we pull data via the predefined getters only. But I am fine revisiting this later on

testing/state-transition/state-transition.go
abi87: FUUUUU

primitives/version/version_test.go
calbera: There's no unique methods on Version once the generics were removed. The Version type alias is still there

da/types/sidecar_test.go
calbera: all tests using mainnet spec, unless testing a specific condition which requires custom values in the `SpecData` struct itself.

shotes: Though I don't think we should be using the "mainnet" spec directly inside of testing. We should still have a separate "devnet" spec that is identical in values to mainnet, but the chain ID is different. I know none of our tests should be sending data anywhere, but I think it's good practice to never ever use the mainnet chain ID even in local tests.

beacon/validator/block_builder.go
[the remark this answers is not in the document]
shotes: yup, these were never being used anywhere. They were only kept in just in case they were planned to be used in a fork upgrade, but I don't know of any plans we have to use those fields anytime soon.

node-core/components/chain_spec.go
```diff
@@ -46,14 +44,10 @@ func ProvideChainSpec() (chain.Spec, error) {
 
 	// TODO: replace reading env var with config value.
 	switch os.Getenv(ChainSpecTypeEnvVar) {
-	case TestnetChainSpecType:
-		chainSpec, err = spec.TestnetChainSpec()
-	case BetnetChainSpecType:
-		chainSpec, err = spec.BetnetChainSpec()
-	case BoonetChainSpecType:
```
abi87, line 53: bye Boonet and thanks for all the fish

primitives/bytes/b4.go
```diff
@@ -91,3 +104,7 @@ func (h B4) MarshalSSZ() ([]byte, error) {
 func (h B4) HashTreeRoot() (B32, error) {
 	return ToBytes32(ExtendToSize(h[:], B32Size))
 }
+
```
shotes, line 107: nit: seems like this comment block corresponds to the moved functions `FromUint32` and `ToUint32`. This should be moved next to those functions.

primitives/bytes/b4.go
```diff
@@ -91,3 +104,7 @@ func (h B4) MarshalSSZ() ([]byte, error) {
...
+/* -------------------------------------------------------------------------- */
+/*                            uint32 conversion                               */
```
shotes, line 109: nit: seems like this comment block corresponds to the moved functions FromUint32 and ToUint32. This should be moved next to those functions.

config/spec/chain_ids.go
```diff
@@ -21,18 +21,14 @@
...
 
 const (
-	// BoonetEth1ChainID is the chain ID for a local devnet.
-	BoonetEth1ChainID uint64 = 80000
-
-	// BetnetEth1ChainID is the chain ID for a local devnet.
-	BetnetEth1ChainID uint64 = 80088
-
-	// DevnetEth1ChainID is the chain ID for a local devnet.
+	// DevnetEth1ChainID is the chain ID for a local devnet. Used by `make start` and unit tests.
```
shotes, line 24: nit: not used by UTs if we change them to use Mainnet.

On the same note though, IMO we should continue using devnet for UTs and e2e, and just modify devnet spec to have the same values as mainnet, other than chain ID.

config/spec/chain_ids.go
```diff
@@ -21,18 +21,14 @@
 package spec
 
 const (
-	// BoonetEth1ChainID is the chain ID for a local devnet.
-	BoonetEth1ChainID uint64 = 80000
```
abi87, line 25: @chuck-bear @gummybera @waldoisabear @KodiakBera for visibility

state-transition/core/core_test.go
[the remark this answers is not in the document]
calbera: we cant because of the specData type. there is a SpecData struct which is different from Spec interface used everywhere

config/spec/defaults.go
calbera: The only issue is if these are unexported it disallows doing something like https://github.com/berachain/beacon-kit/pull/2423/files#r1930925161. because we need the raw specData to be able to modify certain values

config/spec/chain_ids.go
calbera: lets do it. I hope it doesnt break any existing tests.

config/spec/defaults.go
abi87: How about mocking the interface? Not a super clean solution either, but it would allow to unexport things? Spitballing here

shotes: Couldn't you change that function to something like:
```go
func modifiableSpecData() *chain.SpecData {
    mainnet := spec.MainnetChainSpec()
    return &chain.SpecData{
        // Gwei values constants.
	MaxEffectiveBalance:       mainnet.MaxEffectiveBalance(),
    }
}
```

testing/state-transition/state-transition.go
fridrik01: fixed in https://github.com/berachain/beacon-kit/pull/2428

config/spec/defaults.go
```diff
@@ -0,0 +1,89 @@
...
+	defaultDomainTypeSelectionProof    = 5
+	defaultDomainTypeAggregateAndProof = 6
+	defaultDomainTypeApplicationMask   = 16777216 // "0x00000001" in little endian
+
+	// Eth1-related values.
+	defaultDepositContractAddress    = "0x4242424242424242424242424242424242424242" // Berachain specific.
+	defaultMaxDepositsPerBlock       = 16
+	defaultDepositEth1ChainID        = 1
+	defaultEth1FollowDistance        = 1 // Berachain specific.
+	defaultTargetSecondsPerEth1Block = 2 // Berachain specific.
```
abi87, line 57: @calbera: this was 3 in `base.go` but it was modified to 2 in mainnet specs, so here we set default to 2?

config/spec/defaults.go
```diff
@@ -0,0 +1,89 @@
...
+	defaultElectraForkEpoch = 9999999999999999 // Set as a future epoch as not yet determined.
+
+	// State list length constants.
+	defaultEpochsPerHistoricalVector = 8
+	defaultEpochsPerSlashingsVector  = 8
+	defaultHistoricalRootsLimit      = 8
+	defaultValidatorRegistryLimit    = 1099511627776
+
+	// Slashing.
+	defaultInactivityPenaltyQuotient      = 16777216
```
abi87, line 70: @calbera does not look like this is used, nor I can find it in `main`. Where does it come from?

calbera: removing

config/spec/defaults.go
calbera: yeah just setting default to 2. its also unused but leaving it in in case we want later

config/spec/mainnet.go
```diff
@@ -22,60 +22,121 @@ package spec
...
-	// Target for block time is 2 seconds on Berachain mainnet.
-	mainnetSpec.TargetSecondsPerEth1Block = 2
-
+const (
 	// BGT contract address.
 	//
 	// A hard fork will occur to set this value as the BGT contract address
 	// when BGT beings to be minted.
-	mainnetSpec.EVMInflationAddress = common.ExecutionAddress{}
+	MainnetEVMInflationAddress = defaultEVMInflationAddress
```
shotes, line 34: can these be made unexported as well?

abi87 (approved):

PR #2425, opened 2025-01-27
chore(forks): Deneb1 Values

calbera:
## Deneb1 Hardfork

**Values to set:**
- Deneb1 Fork Activation Epoch: [TBD]
- EVM Inflation Address Deneb1: `0x656b95E550C07a9ffe548bd4085c72418Ceb1dba`
- EVM Inflation Per Block Deneb1: `5.75 * 1e9` (5.75 BERA)

## Next Change
- version management (uint32s to B4)
- revive [WIP 6/n](https://github.com/berachain/beacon-kit-internal/pull/88): remove usage of `CHAIN_SPEC` and allow configuring chain spec values in a file


config/spec/mainnet.go
```diff
@@ -73,6 +73,19 @@ const (
...
+	// TODO: set to the correct epoch.
+	MainnetDeneb1ForkEpoch = 5000
+
+	// MainnetEVMInflationAddressDeneb1 is the address on the EVM which will receive the
+	// inflation amount of native EVM balance through a withdrawal every block in the Deneb1 fork.
+	MainnetEVMInflationAddressDeneb1 = "0x656b95E550C07a9ffe548bd4085c72418Ceb1dba"
+
+	// MainnetEVMInflationPerBlockDeneb1 is the amount of native EVM balance (in Gwei) to be
+	// minted to the EVMInflationAddressDeneb1 via a withdrawal every block in the Deneb1 fork.
+	MainnetEVMInflationPerBlockDeneb1 = 5.75 * 1e9
```
rezbera, line 88: do we have a constant for the gwei multiplier somewhere?

calbera: done

fridrik01 (commented):
lgtm

config/spec/mainnet.go
```diff
@@ -71,6 +72,19 @@ const (
 	// The deposit contract address on mainnet at genesis is the same as the
 	// default deposit contract address.
 	MainnetDepositContractAddress = defaultDepositContractAddress
+
+	// MainnetDeneb1ForkEpoch is the epoch at which the Deneb1 fork occurs.
+	//
+	// TODO: set to the correct epoch.
+	MainnetDeneb1ForkEpoch = 5000
```
abi87, line 79: @calbera I believe we should do something like that for whatever testnet our e2e tests use. So that we check this fork in e2e tests as well

config/spec/mainnet.go
```diff
@@ -129,10 +143,14 @@ func MainnetChainSpecData() *chain.SpecData {
...
 
-		// Berachain values.
+		// Berachain values at genesis.
 		ValidatorSetCap:      MainnetValidatorSetCap,
 		EVMInflationAddress:  common.NewExecutionAddressFromHex(MainnetEVMInflationAddress),
 		EVMInflationPerBlock: MainnetEVMInflationPerBlock,
+
+		// Deneb1 values.
+		EVMInflationAddressDeneb1:  common.NewExecutionAddressFromHex(MainnetEVMInflationAddressDeneb1),
+		EVMInflationPerBlockDeneb1: MainnetEVMInflationPerBlockDeneb1,
```
abi87, line 153: @calbera these are not the values at genesis right?

config/spec/mainnet.go
calbera: Right now its devnet, but we will have to use a shorter number in devnet so that e2e doesn't spin until a really large block number

config/spec/mainnet.go
calbera: these are the Deneb1 values

config/spec/mainnet.go
```diff
@@ -48,16 +49,16 @@ const (
 	MainnetMaxValidatorsPerWithdrawalsSweep = 31
 
 	// MainnetMaxEffectiveBalance is the max stake of 10 million BERA at genesis.
-	MainnetMaxEffectiveBalance = 10_000_000 * 1e9
+	MainnetMaxEffectiveBalance = 10_000_000 * math.GweiPerWei
```
rezbera, line 52: Is the comment above wrong? Comment implies the value is 10million BERA, but the value is 10million gwei

calbera: maybe fix in another PR renaming "GweiPerWei" to "WeiPerGwei"

rezbera: https://github.com/berachain/beacon-kit/pull/2433

testing/e2e/e2e_inflation_test.go
```diff
@@ -0,0 +1,104 @@
...
+		)
+		s.Require().NoError(err)
+		s.Require().Zero(balance.Cmp(expectedBalance))
+	}
+
+	// If the addresses are different, enforce that the balance of the EVM inflation address
+	// prior to the hardfork is the same as it is now.
+	priorEVMInflationAddress := chainspec.EVMInflationAddress(constants.GenesisSlot)
+	postEVMInflationAddress := chainspec.EVMInflationAddress(math.Slot(deneb1ForkSlot))
+	if priorEVMInflationAddress != postEVMInflationAddress {
```
rezbera, line 87: this `if priorEVMInflationAddress != postEVMInflationAddress {` statement feels odd given we know they are not the same address in the test setup. I could see it being useful if you wanted to change the devspec in the future for some edge case testing but that should be a separate test

chain/spec.go
```diff
@@ -424,12 +427,28 @@ func (s spec) ValidatorSetCap() uint64 {
...
-	return s.Data.EVMInflationAddress
+func (s spec) EVMInflationAddress(slot math.Slot) common.ExecutionAddress {
+	fv := s.ActiveForkVersionForSlot(slot)
+	switch fv {
+	case version.Deneb1:
+		return s.Data.EVMInflationAddressDeneb1
+	case version.Deneb:
+		return s.Data.EVMInflationAddressGenesis
+	default:
+		panic(fmt.Sprintf("EVMInflationAddress not supported for this fork version: %d", fv))
```
shotes, line 438: I don't love the idea of panicing on an invalid fork version - for example, if a slot is used that lines up with `version.Electra`.

Currently the usages of this function `EVMInflationAddress` and `EVMInflationPerBlock` use the slot directly from the state, so I think it is safe. Generally though I'd prefer to return an error that can be handled instead of panicing. Don't think this is a blocker, just something to think about.

For example, what if there is a request from the node-api that requires processing a block in such a way that the slot lines up with `version.Electra`?

config/spec/devnet.go
```diff
@@ -26,16 +26,27 @@ import (
...
+	devnetEVMInflationPerBlock = 10e9
 
-	// MaxStakeAmount is the maximum amount of native EVM balance (in units
+	// devnetMaxStakeAmount is the maximum amount of native EVM balance (in units
 	// of Gwei) that can be staked.
-	DevnetMaxStakeAmount = 4000e9
+	devnetMaxStakeAmount = 4000e9
+
+	// devnetDeneb1ForkEpoch is the epoch at which the Deneb1 fork occurs.
+	devnetDeneb1ForkEpoch = 1
```
shotes, line 41: Love this - I know it's specifically for testing the deneb1 values, but it also gives confidence in fork logic (and finding bugs in regression, like the signature domain).

abi87 (approved):
LGTM

PR #2426, opened 2025-01-27
feat(cli): Tool to display all formats of public key

calbera:
### Expected usage

```shell
$ beacond --home $HOME_DIR deposit validator-keys

Comet Address:
D7D956D7D454FD2110FB4B99F9404847D410F335

Comet Pubkey (Base64):
Cgeg5/iYKUd4CemZqzT9NqcNX5uFhnvkp0ATgYQ2PSdurtCEe8wimjfbQgJ/DA1GGFujqPJclWM9g9MmqOTkHEDGrfrB1lJVNC+aNR6sYX3dBKmz/mbNPY6+I4RhpNcP

Eth/Beacon Pubkey (Compressed 48-byte Hex):
0xaa07a0e7f89829477809e999ab34fd36a70d5f9b85867be4a740138184363d276eaed0847bcc229a37db42027f0c0d46
```


cli/commands/deposit/keys.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/cometbft/cometbft/crypto/bls12381"
+	"github.com/spf13/cobra"
+)
+
+// GetValidatorKeysCmd returns a command that returns the validator public key in different formats
+// for the given private key files.
+//
+//nolint:lll // reads better if long description is one line.
+func GetValidatorKeysCmd() *cobra.Command {
+	cmd := &cobra.Command{
```
abi87, line 39: @calbera we do have requests from backend to have a tool that helps mapping "Comet-like" pub keys to "Eth-like" pub keys.
Should we rework this to accept the either and return the other?

calbera: this was more as a tool for validators that @codingwithmanny found to be useful. I think for backend they should use the underlying library and not rely on CLI

codingwithmanny:
can you leave review if “Expected Usage” looks good to u

abi87 (changes requested):
Just noting here we want to merge this after all the hard fork code

fridrik01 (approved):
lgtm

rezbera (approved):
Can't vouch for the outputs if they correct or not - ideally have some unit tests, but not in the core loop and solves a need

abi87 (approved):

PR #2430, opened 2025-01-28
chore(types): Set version

calbera:
There are a few objects which:

- we build passing a Version
- have a Version method, which currently returns Deneb

We are now introducing Deneb1, so these objects should:
- either return the right version in Version
- or have no Version method anymore

This PR attempt to do the first alternative (fix Version)

Open issues:

- we do not serialize version for backward compatibility reaons, which means that the simple checks were we serialize and deserialize an object fails due to version not being set.
  - My idea for this is that the client which does deserialization should know what is the right version and set it once the deserialization succeed. Not sure client can guarantee this.

### DIFF
--- a/consensus-types/types/block.go
+++ b/consensus-types/types/block.go
@@ -33,6 +33,9 @@ import (
 // BeaconBlock represents a block in the beacon chain during
 // the Deneb fork.
 type BeaconBlock struct {
+	// version is the version of the beacon block.
+	version uint32
+
 	// Slot represents the position of the block in the chain.
 	Slot math.Slot `json:"slot"`
 	// ProposerIndex is the index of the validator who proposed the block.
@@ -41,14 +44,15 @@ type BeaconBlock struct {
 	ParentRoot common.Root `json:"parent_root"`
 	// StateRoot is the hash of the state at the block.
 	StateRoot common.Root `json:"state_root"`
-	// Body is the body of the BeaconBlock, containing the block's
-	// operations.
+	// Body is the body of the BeaconBlock, containing the block's operations.
 	Body *BeaconBlockBody `json:"body"`
 }
 
 // Empty creates an empty beacon block.
-func (*BeaconBlock) Empty() *BeaconBlock {
-	return &BeaconBlock{}
+func (*BeaconBlock) Empty(forkVersion uint32) *BeaconBlock {
+	return &BeaconBlock{
+		version: forkVersion,
+	}
 }
 
 // NewBeaconBlockWithVersion assembles a new beacon block from the given.
@@ -61,6 +65,7 @@ func NewBeaconBlockWithVersion(
 	switch forkVersion {
 	case version.Deneb, version.Deneb1:
 		return &BeaconBlock{
+			version:       forkVersion,
 			Slot:          slot,
 			ProposerIndex: proposerIndex,
 			ParentRoot:    parentBlockRoot,
@@ -147,7 +152,7 @@ func (b *BeaconBlock) GetStateRoot() common.Root {
 
 // Version identifies the version of the BeaconBlock.
 func (b *BeaconBlock) Version() uint32 {
-	return version.Deneb
+	return b.version
 }
 
 // SetStateRoot sets the state root of the BeaconBlock.
--- a/consensus-types/types/block.go
+++ b/consensus-types/types/block.go
@@ -49,9 +49,10 @@ type BeaconBlock struct {
 }
 
 // Empty creates an empty beacon block.
-func (*BeaconBlock) Empty(forkVersion uint32) *BeaconBlock {
+func (*BeaconBlock) Empty() *BeaconBlock {
 	return &BeaconBlock{
-		version: forkVersion,
+		// By default, we return a Deneb block to maintain compatibility.
+		version: version.Deneb,
 	}
 }
 
--- a/consensus-types/types/block.go
+++ b/consensus-types/types/block.go
@@ -33,9 +33,6 @@ import (
 // BeaconBlock represents a block in the beacon chain during
 // the Deneb fork.
 type BeaconBlock struct {
-	// version is the version of the beacon block.
-	version uint32
-
 	// Slot represents the position of the block in the chain.
 	Slot math.Slot `json:"slot"`
 	// ProposerIndex is the index of the validator who proposed the block.
@@ -46,12 +43,16 @@ type BeaconBlock struct {
 	StateRoot common.Root `json:"state_root"`
 	// Body is the body of the BeaconBlock, containing the block's operations.
 	Body *BeaconBlockBody `json:"body"`
+
+	// version is the version of the beacon block.
+	// version must be not serialized
+	version uint32
 }
 
 // Empty creates an empty beacon block.
 func (*BeaconBlock) Empty() *BeaconBlock {
 	return &BeaconBlock{
-		// By default, we set the version to Deneb to maintain compatibility.
+		// By default, we set the version to Deneb to maintain backward-compatibility.
 		version: version.Deneb,
 	}
 }
@@ -66,12 +67,12 @@ func NewBeaconBlockWithVersion(
 	switch forkVersion {
 	case version.Deneb, version.Deneb1:
 		return &BeaconBlock{
-			version:       forkVersion,
 			Slot:          slot,
 			ProposerIndex: proposerIndex,
 			ParentRoot:    parentBlockRoot,
 			StateRoot:     common.Root{},
 			Body:          &BeaconBlockBody{},
+			version:       forkVersion,
 		}, nil
 	default:
 		// we return block here to appease nilaway
--- a/consensus-types/types/block.go
+++ b/consensus-types/types/block.go
@@ -47,7 +47,7 @@ type BeaconBlock struct {
 	// BbVersion is the BbVersion of the beacon block.
 	// BbVersion must be not serialized but it's exported
 	// to allow unit tests using reflect on beacon block.
-	BbVersion uint32
+	BbVersion uint32 `json:"-"`
 }
 
 // Empty creates an empty beacon block.
--- a/consensus-types/types/block.go
+++ b/consensus-types/types/block.go
@@ -51,7 +51,7 @@ type BeaconBlock struct {
 // Empty creates an empty beacon block.
 func (*BeaconBlock) Empty() *BeaconBlock {
 	return &BeaconBlock{
-		// By default, we return a Deneb block to maintain compatibility.
+		// By default, we set the version to Deneb to maintain compatibility.
 		version: version.Deneb,
 	}
 }
--- a/consensus-types/types/block.go
+++ b/consensus-types/types/block.go
@@ -44,16 +44,17 @@ type BeaconBlock struct {
 	// Body is the body of the BeaconBlock, containing the block's operations.
 	Body *BeaconBlockBody `json:"body"`
 
-	// version is the version of the beacon block.
-	// version must be not serialized
-	version uint32
+	// BbVersion is the BbVersion of the beacon block.
+	// BbVersion must be not serialized but it's exported
+	// to allow unit tests using reflect on beacon block.
+	BbVersion uint32
 }
 
 // Empty creates an empty beacon block.
 func (*BeaconBlock) Empty() *BeaconBlock {
 	return &BeaconBlock{
 		// By default, we set the version to Deneb to maintain backward-compatibility.
-		version: version.Deneb,
+		BbVersion: version.Deneb,
 	}
 }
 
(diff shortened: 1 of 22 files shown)


consensus-types/types/block.go
```diff
@@ -41,14 +41,20 @@ type BeaconBlock struct {
...
+	// version must be not serialized
+	version uint32
 }
 
 // Empty creates an empty beacon block.
 func (*BeaconBlock) Empty() *BeaconBlock {
-	return &BeaconBlock{}
+	return &BeaconBlock{
+		// By default, we set the version to Deneb to maintain backward-compatibility.
+		version: version.Deneb,
```
abi87, line 56: TODO: this could probably we replaced with EmptyWithVersion. To be checked

consensus-types/types/payload.go
```diff
@@ -454,13 +457,15 @@ func (p *ExecutionPayload) UnmarshalJSON(input []byte) error {
...
+func (p *ExecutionPayload) Empty(forkVersion uint32) *ExecutionPayload {
+	return &ExecutionPayload{
+		version: forkVersion,
+	}
 }
 
 // Version returns the version of the ExecutionPayload.
 func (p *ExecutionPayload) Version() uint32 {
-	return version.Deneb
+	return p.version
```
abi87, line 468: this change requires, similar to others, that:
- we either only use only ctors that duly set version
- we always call SetVersion after we use naked structs

consensus-types/types/payload_header.go
```diff
@@ -34,14 +34,6 @@ import (
 
 // ExecutionPayloadHeader is the execution header payload of Deneb.
 type ExecutionPayloadHeader struct {
-	// TODO: Enable once
-	// https://github.com/karalabe/ssz/pull/9/files# is merged.
```
abi87, line 38: not sure we should delete this. We should look into this PR

consensus-types/types/signed_beacon_block.go
```diff
@@ -56,7 +56,14 @@ func NewSignedBeaconBlockFromSSZ(
 	block := &SignedBeaconBlock{}
 	switch forkVersion {
 	case version.Deneb, version.Deneb1:
-		return block, block.UnmarshalSSZ(bz)
+		if err := block.UnmarshalSSZ(bz); err != nil {
+			return block, err
+		}
+
+		// duly setup fork version in every relevant block member
+		block.Message.Body.ExecutionPayload.EpVersion = forkVersion
```
abi87, line 64: without this, even a simple `make start` `make start-reth` fails

consensus-types/types/block.go
```diff
@@ -41,14 +41,21 @@ type BeaconBlock struct {
...
+	// BbVersion is the BbVersion of the beacon block.
+	// BbVersion must be not serialized but it's exported
+	// to allow unit tests using reflect on beacon block.
+	BbVersion uint32
 }
 
 // Empty creates an empty beacon block.
 func (*BeaconBlock) Empty() *BeaconBlock {
-	return &BeaconBlock{}
+	return &BeaconBlock{
```
abi87, line 55: I don't want to modify the signature here to pass the `forkVersion` although I think it should be safe

consensus-types/types/payload_test.go
```diff
@@ -232,13 +242,9 @@ func TestExecutionPayload_ToHeader(t *testing.T) {
...
 	require.Equal(t, payload.GetExcessBlobGas(), header.GetExcessBlobGas())
+	require.Equal(t, payload.Version(), header.Version())
 
-	// TODO: FIX LATER
-	// htrHeader, err := header.HashTreeRoot()
-	// require.NoError(t, err)
-	// htrPayload, err := payload.HashTreeRoot()
-	// require.NoError(t, err)
-	// require.Equal(t, htrPayload, htrHeader)
+	require.Equal(t, payload.HashTreeRoot(), header.HashTreeRoot())
```
abi87, line 247: this just work.

consensus-types/types/payload_test.go
```diff
@@ -232,13 +242,9 @@ func TestExecutionPayload_ToHeader(t *testing.T) {
 	require.Equal(t, payload.GetBlockHash(), header.GetBlockHash())
 	require.Equal(t, payload.GetBlobGasUsed(), header.GetBlobGasUsed())
 	require.Equal(t, payload.GetExcessBlobGas(), header.GetExcessBlobGas())
+	require.Equal(t, payload.Version(), header.Version())
```
abi87, line 245: added the version check, since I explicitly set the payload header version from the payload one

state-transition/core/state_processor_staking_test.go
```diff
@@ -336,15 +316,10 @@ func TestTransitionCreateValidator(t *testing.T) {
 		t,
 		st,
 		&types.BeaconBlockBody{
-			ExecutionPayload: &types.ExecutionPayload{
-				Timestamp:    blk1.Body.ExecutionPayload.Timestamp + 1,
```
abi87, line 340: we used blk1 here but this is wrong. The timestamp is hardly verified here, since we mocked the interaction with the EVM, but use blk.Timestamp is the correct thing to do (we increase of 1 second the parent payload timestamp)

state-transition/core/state_processor_staking_test.go
```diff
@@ -720,22 +682,17 @@ func TestTransitionHittingValidatorsCap_ExtraSmall(t *testing.T) {
...
+	blk := moveToEndOfEpoch(t, blk1, cs, sp, st, ctx, depRoot)
 
 	// finally the block turning epoch
-	blk := buildNextBlock(
+	blk = buildNextBlock(
 		t,
 		st,
 		&types.BeaconBlockBody{
-			ExecutionPayload: &types.ExecutionPayload{
-				Timestamp:    blk1.Body.ExecutionPayload.Timestamp + 1,
```
abi87, line 731: same here, we should use parent block Timestamp ( but again having mocked the EVM interaction we don't really validate the timestamp)

state-transition/core/state_processor_genesis.go
```diff
@@ -65,6 +65,7 @@ func (sp *StateProcessor[_]) InitializePreminedBeaconStateFromEth1(
 		Eth1Data: &ctypes.Eth1Data{},
 		ExecutionPayload: &ctypes.ExecutionPayload{
 			ExtraData: make([]byte, ctypes.ExtraDataSize),
+			EpVersion: constants.GenesisVersion,
```
abi87, line 68: I believe this does not really have any practical effect, but the version was not set right

abi87 (approved):
lgtm, but I coded most of it, so maybe I need a second approval

consensus-types/types/payload_header.go
```diff
@@ -78,28 +70,39 @@ type ExecutionPayloadHeader struct {
 	BlobGasUsed math.U64 `json:"blobGasUsed"`
 	// ExcessBlobGas is the amount of excess blob gas in the block.
 	ExcessBlobGas math.U64 `json:"excessBlobGas"`
+
+	// EphVersion is the fork EphVersion of the execution payload header.
+	// EphVersion must be not serialized but it's exported
+	// to allow unit tests using reflect on execution payload header.
+	// TODO: Enable once
+	// https://github.com/karalabe/ssz/pull/9/files# is merged.
```
fridrik01, line 78: I am not sure what this comment means

consensus-types/types/block.go
```diff
@@ -41,14 +41,21 @@ type BeaconBlock struct {
...
 	StateRoot common.Root `json:"state_root"`
-	// Body is the body of the BeaconBlock, containing the block's
-	// operations.
+	// Body is the body of the BeaconBlock, containing the block's operations.
 	Body *BeaconBlockBody `json:"body"`
+
+	// BbVersion is the BbVersion of the beacon block.
+	// BbVersion must be not serialized but it's exported
+	// to allow unit tests using reflect on beacon block.
+	BbVersion uint32
```
rezbera, line 50: Do we want this to be explicitly excluded from serializing, i.e. `json:"-"`

consensus-types/types/payload_header.go
```diff
@@ -78,28 +70,39 @@ type ExecutionPayloadHeader struct {
...
 	return h, h.UnmarshalSSZ(bz)
 }
 
 // NewFromJSON returns a new ExecutionPayloadHeader from the given JSON bytes.
 func (h *ExecutionPayloadHeader) NewFromJSON(
-	bz []byte, _ uint32,
+	bz []byte, forkVersion uint32,
 ) (*ExecutionPayloadHeader, error) {
 	h = h.Empty()
+	h.EphVersion = forkVersion
```
fridrik01, line 105: should we not set the `EphVersion` after we call `json.Unmarshal` to be more explicit on the intent? Or just remove version from the function signature and set it from outside

consensus-types/types/payload_header.go
abi87: Me neither, I just want to keep it there until we clean this up

state-transition/core/state_processor_staking_test.go
```diff
@@ -1249,15 +1162,10 @@ func TestValidatorNotWithdrawable(t *testing.T) {
...
-				Transactions: [][]byte{},
-				Withdrawals: []*engineprimitives.Withdrawal{
-					st.EVMInflationWithdrawal(constants.GenesisSlot + 1),
-				},
-				BaseFeePerGas: math.NewU256(0),
-			},
+			ExecutionPayload: testPayload(
+				10,
+				st.EVMInflationWithdrawal(constants.GenesisSlot+1),
+			),
```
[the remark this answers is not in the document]
abi87: I agree

execution/client/ethclient/engine.go
```diff
@@ -42,7 +42,12 @@ func (s *Client) NewPayload(
 	versionedHashes []common.ExecutionHash,
 	parentBlockRoot *common.Root,
 ) (*engineprimitives.PayloadStatusV1, error) {
-	if payload.Version() < version.Deneb {
+	// This is wrong. We should do bytes comparison of versions
+	// Setting this to version.Deneb1 but we should fix version
+	// comparison and drop this check here. The EL does not know
+	// anything about CL versions, so payload should not have a
+	// CL version (or at least not a check here).
+	if payload.Version() < version.Deneb1 {
```
calbera, line 50: this binary will fail backwards compat because if you're syncing from genesis, payload.Version() will be version.Deneb.

consensus-types/types/payload_header.go
abi87: I believe that was fine since version is not json marshalled. Still changed along your suggestions

execution/client/ethclient/engine.go
abi87: But it will start failing as soon as we start using the new version right?

calbera: version.Deneb1 > version.Deneb so we'll be fine

calbera: but agree with comment, this version check needs to be fully reworked eventually

rezbera: version.Deneb1 (26)  is greater than  version.Deneb (4). Deneb1 is a bit of an ugly catchall but this should work in a sync.

abi87: Just to cross check:

- The check was `err if payload.Version() < version.Deneb`.
- Since `version.Deneb1 > version.Deneb` the check `payload.Version() < version.Deneb` would start to fail as soon as payload gets created with version `version.Deneb1`
- Changing the check to `payload.Version() < version.Deneb1` should be fine

Actually not, it should be `payload.Version() <= version.Deneb1`, right @rezbera ?

calbera: the check should remain as is `if payload.Version() < version.Deneb` because its the error case i.e. the given version is unsupported. We don't support versions before Deneb, but we support everything after. So in the inverse `if payload.Version() >= version.Deneb` which is true if `payload.Version() == version.Deneb1`, we will continue as expected and NOT fail.

abi87: I am wrong too, we should err if version is not Deneb or Deneb1

execution/client/ethclient/engine.go
```diff
@@ -42,13 +42,18 @@ func (s *Client) NewPayload(
...
 	parentBlockRoot *common.Root,
 ) (*engineprimitives.PayloadStatusV1, error) {
-	if payload.Version() < version.Deneb {
-		return nil, ErrInvalidVersion
+	// This is wrong. We should do bytes comparison of versions
+	// Setting this to version.Deneb1 but we should fix version
+	// comparison and drop this check here. The EL does not know
+	// anything about CL versions, so payload should not have a
+	// CL version (or at least not a check here).
+	if payload.Version() == version.Deneb ||
```
rezbera, line 50: much better

execution/client/ethclient/engine.go
abi87: @calbera I understand your point. Still I think:

- Using ordering over versions as uint32 is wrong (it should be fine using ordering over they bytes representations)
- We do have Electra version but its fork is set to far future, so I am fine erring here as it would should we as wrongly setting electra somewhere

fridrik01 (approved):

PR #2431, opened 2025-01-28
chore: fix time validation in acceptOperatorChange in DepositContract…

mdqst:
I noticed the time validation in `acceptOperatorChange` might not be accurate.
The current condition uses `>`, which could allow the transaction to go through even if a full day hasn't passed. I've updated it to use `>=` to ensure the required one-day delay is strictly enforced.

This change makes the logic more precise and prevents potential edge cases.  

Here's the updated code:  
```solidity
if (queuedTimestamp + ONE_DAY >= uint96(block.timestamp)) {
    revert NotEnoughTime();
}
```  

In Bera we Trust!


calbera:
thanks for the PR, but this change is not bug fix so closing for now.

PR #2433, opened 2025-01-28
chore(types): Remove abstraction for Gwei

rezbera:
Remove unnecessary duplicate type we already have available in geth


shotes (approved):
LGTM

PR #2437, opened 2025-01-29
chore(types): Introduce canonical sorting of common.Version

calbera:
## TLDR
Making usage of version more consistent with the Eth2.0 spec, where fork `Version` is of type `Bytes4` (https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#custom-types).

## Changes
- **Prevents misusing versions as `uint32` and wrongly comparing them numerically**
  - All usage of `forkVersion`s as type `uint32` has been removed.
- Strongly enforces using `common.Version` as THE type to use when referring to fork versions
  - Adds comparable utility functions in `version` package to allow comparing common.Version 
- Necessary refactoring to make this change possible, all peripheral. NOTE: no functional changes to behavior have been made in this PR. For example, If we choose to decide to ignore beacon versions in our EngineAPI client, we should do that in another PR.


cli/utils/parser/errors.go
```diff
@@ -44,11 +44,6 @@ var (
 		"invalid signature length",
 	)
 
-	// ErrInvalidVersionLength is returned when the deposit version is invalid.
-	ErrInvalidVersionLength = errors.New(
```
calbera, line 48: unused

execution/client/ethclient/engine.go
calbera, line 1: this file currently contains most of the usage of `Version`. Choosing to keep the version check here to allow the possibility of constraining beacon version to a specific engine api version. So in our case, the mapping looks something like.

Beacon Version         >>         EngineAPI Version
-----------------------------------------------
Phase0, ..., Capella   >>  Err with unsupported
Deneb, Deneb1, ...     >>         V3
(Potentially) Some future beacon version >> V4

calbera: so this way if Pre-deneb support was needed can strictly enforce which engine api V1/V2 to use for each beacon version

execution/client/ethclient/engine_test.go
calbera, line 1: tests in this file provides example of the current beacon versions we support (and don't support) for calling (the appropriate version of) the engine api

rezbera (commented):
Any reason not to do it like this so we don't do the Bytes conversion on every call?

```
var (
	Phase0Version    = bytes.FromUint32(phase0)
	AltairVersion    = bytes.FromUint32(altair)
	BellatrixVersion = bytes.FromUint32(bellatrix)
	CapellaVersion   = bytes.FromUint32(capella)
	DenebVersion     = bytes.FromUint32(deneb)
	Deneb1Version    = bytes.FromUint32(deneb1)
	ElectraVersion   = bytes.FromUint32(electra)
)

// Phase0 returns phase0 as a common.Version.
func Phase0() common.Version {
	return Phase0Version
}
```

calbera:
> ```go
> var (
> 	Phase0Version    = bytes.FromUint32(phase0)
> 	AltairVersion    = bytes.FromUint32(altair)
> 	BellatrixVersion = bytes.FromUint32(bellatrix)
> 	CapellaVersion   = bytes.FromUint32(capella)
> 	DenebVersion     = bytes.FromUint32(deneb)
> 	Deneb1Version    = bytes.FromUint32(deneb1)
> 	ElectraVersion   = bytes.FromUint32(electra)
> )
> ```

Want to prevent using global variables (which could technically get modified in runtime).

primitives/bytes/b4_test.go
```diff
@@ -386,3 +275,147 @@ func TestBytes4HashTreeRoot(t *testing.T) {
...
+			b:        bytes.B4{0x01, 0x02, 0x03, 0x04},
+			expected: true,
+		},
+		{
+			name:     "fork versions",
+			a:        version.Deneb(),
+			b:        version.Deneb1(),
+			expected: true,
+		},
+	}
```
rezbera, line 355:
```go
		{
			name:     "lexographically sorted bytes",
			a:        version.Deneb1(),
			b:        version.Electra(),
			expected: true,
		},
```

The following test fails due to byte by byte comparison. That doesn't seem right as we would always consider Electra greater than Deneb

primitives/bytes/b4.go
```diff
@@ -108,3 +108,34 @@ func FromUint32(v uint32) B4 {
...
+// IsGreaterThanOrEqualTo returns true if each byte of h is greater than
+// or equal to the corresponding byte of b.
+func (h B4) IsGreaterThanOrEqualTo(b B4) bool {
+	for i := range B4Size {
+		if h[i] < b[i] {
+			return false
+		}
+	}
+	return true
+}
```
rezbera, line 141: Comparing each byte seems incorrect see https://github.com/berachain/beacon-kit/pull/2437/files#r1936378692

We should short circuit if we see a greater byte at the start

consensus-types/types/block.go
```diff
@@ -63,10 +63,10 @@ func NewBeaconBlockWithVersion(
 	slot math.Slot,
 	proposerIndex math.ValidatorIndex,
 	parentBlockRoot common.Root,
-	forkVersion uint32,
+	forkVersion common.Version,
 ) (*BeaconBlock, error) {
 	switch forkVersion {
-	case version.Deneb, version.Deneb1:
+	case version.Deneb(), version.Deneb1():
```
abi87, line 69: nit: not sure we should do this.
I think this works for use because:
- the underlying type of `common.Version` is  a byte slice
- we only use comparison not ordering

Still maybe we should introduce an explitic version.Equal helper?
It may be my C++ mind speaking here, this is probably fine

primitives/bytes/b4_test.go
calbera: fixed in commit 563093b15cbc6ce3d31dc9a8b1bc9a744faa664d

primitives/bytes/b4.go
calbera: fixed in commit 563093b15cbc6ce3d31dc9a8b1bc9a744faa664d

consensus-types/types/block.go
calbera: .Equals (among other comparable funcs) implemented [here](https://github.com/berachain/beacon-kit/blob/5ec0a3ded8a773388704e7afd247b6c19f2228a9/primitives/bytes/b4.go#L131)

consensus-types/types/genesis_test.go
```diff
@@ -34,10 +34,10 @@ import (
 
 func TestDefaultGenesis(t *testing.T) {
 	g := types.DefaultGenesis()
-	if g.ForkVersion != bytes.FromUint32(version.Deneb) {
+	if !g.ForkVersion.Equals(version.Deneb()) {
```
abi87, line 37: so the equals here is the bytes.Equal, which is fine.
I wonder if we should dress it with a version.Equal?

calbera: I explored keeping the comparable functions as utility functions in package `version`. In the end the readability of code felt better with putting these on the type itself with receiver.

engine-primitives/engine-primitives/attributes.go
```diff
@@ -112,7 +112,8 @@ func (p *PayloadAttributes) Validate() error {
 		return ErrEmptyPrevRandao
 	}
 
-	if p.Withdrawals == nil && p.version >= version.Capella {
+	// For any fork version Capella onwards, withdrawals are required.
+	if p.version.IsGreaterThanOrEqualTo(version.Capella()) && p.Withdrawals == nil {
```
abi87, line 116: this could be an `!version.isBefore(version.Capella())`

primitives/version/versions.go
```diff
@@ -20,23 +20,64 @@
...
-	// Deneb is the first version of Deneb, used for genesis of Berachain mainnet.
-	Deneb uint32 = 4
-	// Deneb1 is the first hardfork of Deneb on Berachain mainnet (TBD if used).
-	// There may also be Deneb2, Deneb3, etc. hardforks.
-	Deneb1 uint32 = 260
-	// Electra is the first version of Electra on Berachain mainnet.
-	Electra uint32 = 5
-	// Electra1 is the first hardfork of Electra on Berachain mainnet (TBD if used).
-	// There may also be Electra2, Electra3, etc. hardforks.
-	Electra1 uint32 = 261
```
abi87, line 41: the reason we dropped `Electra1` is that we don't need to show anymore how to get the next?

primitives/version/versions.go
```diff
@@ -20,23 +20,64 @@
...
+	// bellatrix is the second hardfork of the Beacon Chain.
+	bellatrix uint32 = 2
+	// capella is the third hardfork of the Beacon Chain.
+	capella uint32 = 3
+	// deneb is the first version of Deneb, used for genesis of Berachain mainnet.
+	deneb uint32 = 4
+	// deneb1 is the first hardfork of Deneb on Berachain mainnet. LittleEndian of {4, 1, 0, 0}.
+	deneb1 uint32 = 260
+	// electra is the first version of Electra on Berachain mainnet.
+	electra uint32 = 5
```
abi87, line 42: nit: probably excessive but we could:
- explicitly define these as `var <fork_name> []byte = <their byte representation>`
- copy the byte slices in the getter to avoid any client modifying them

It would make clearer that ordering of version follows the byte representation, not the numerical one

storage/beacondb/eth1.go
```diff
@@ -31,7 +34,7 @@ func (kv *KVStore) GetLatestExecutionPayloadHeader() (
 	if err != nil {
 		return nil, err
 	}
-	kv.latestExecutionPayloadCodec.SetActiveForkVersion(forkVersion)
+	kv.latestExecutionPayloadCodec.SetActiveForkVersion(bytes.FromUint32(forkVersion))
```
abi87, line 37: this is so bad, but indeed it requires a coordinate update to change the way we serialize data

execution/client/ethclient/engine_test.go
```diff
@@ -49,6 +51,85 @@ func TestGetPayloadV3NeverReturnsEmptyPayload(t *testing.T) {
...
+	c := ethclient.New(&stubRPCClient{t: t})
+	ctx := context.Background()
+
+	payload := &ctypes.ExecutionPayload{EpVersion: version.Capella()}
+	versionedHashes := []common.ExecutionHash{}
+	var parentBlockRoot *common.Root
+
+	status, err := c.NewPayload(ctx, payload, versionedHashes, parentBlockRoot)
+	require.ErrorIs(t, err, ethclient.ErrInvalidVersion)
+	require.Nil(t, status)
```
abi87, line 78: do we need `require.Nil(t, status)`? Why isn't the error enough

abi87: Same for the similar checks below

primitives/bytes/b4.go
```diff
@@ -108,3 +108,48 @@ func FromUint32(v uint32) B4 {
...
+// IsLessThanOrEqualTo returns true if h is less than or equal to b. This compares bytes from
+// most significant to least significant in little-endian order.
+func (h B4) IsLessThanOrEqualTo(b B4) bool {
+	for i := range B4Size {
+		if h[i] <= b[i] {
+			return true
+		}
+	}
+	return false
+}
```
abi87, line 144: Can these be version methods rather than B4?
I think it would be clearer

primitives/bytes/b4.go
```diff
@@ -108,3 +108,48 @@ func FromUint32(v uint32) B4 {
...
+
+// IsGreaterThanOrEqualTo returns true if h is greater than or equal to b.
+// This compares bytes from most significant to least significant in little-endian order.
+func (h B4) IsGreaterThanOrEqualTo(b B4) bool {
```
abi87, line 148: if we have `Equals` and `IsLessThanOrEqualTo` (even just `isLess`)\,w e don't need `IsGreaterThanOrEqualTo`

primitives/bytes/b4_test.go
```diff
@@ -386,3 +275,159 @@ func TestBytes4HashTreeRoot(t *testing.T) {
 		})
 	}
 }
+
+func TestBytes4Equals(t *testing.T) {
```
abi87, line 279: I think this is more than we need.
There could be a single UT in `version` package where we list the versions in a slice and we verify that each element is strictly before the next one

execution/client/ethclient/engine.go
```diff
@@ -36,28 +36,22 @@ import (
...
-	// Setting this to version.Deneb1 but we should fix version
-	// comparison and drop this check here. The EL does not know
-	// anything about CL versions, so payload should not have a
-	// CL version (or at least not a check here).
-	if payload.Version() == version.Deneb ||
-		payload.Version() == version.Deneb1 {
-		return s.NewPayloadV3(
-			ctx, payload, versionedHashes, parentBlockRoot,
-		)
+	// newPayloadV3 is used for beacon versions Deneb and onwards.
```
abi87, line 46: I think @rezbera idea is to move the fork versions check out if this.
Rationale being that EL does not really know about versions so even ethclient engine should not?

I like the idea but it does not have to be implemented in this PR

abi87 (changes requested):
few nits

consensus-types/types/genesis_test.go
abi87: But it's more than we need and we suggest that we have a canonical way to order/compare bytes, which we don't have.
We do only have versions ordering

execution/client/ethclient/engine.go
calbera: this is actually preserving the existing behavior so I'd rather keep the version enforcement in this PR. If we choose to remove it completely, can do in a separate PR

rezbera (changes requested):
IsLessThanOrEqualTo impl still incorrect as discussed - will review after fixes

primitives/version/versions.go
calbera: Don't mind keeping this in or out tbh

primitives/version/versions.go
calbera: keeping them as vars but private fine with me but then we have global vars, indifferent either way

execution/client/ethclient/engine_test.go
calbera: you're right

storage/beacondb/eth1.go
calbera: Yeah annoying, but erred on the side of keeping state consistent. So have to deal with a manual conversion before storage :/

consensus-types/types/genesis_test.go
calbera: done, enforced in sorting test

engine-primitives/engine-primitives/attributes.go
calbera: done ^

primitives/version/versions.go
calbera: done, kept as vars

execution/client/ethclient/engine_test.go
calbera: done

primitives/bytes/b4_test.go
calbera: done

primitives/constants/misc.go
```diff
@@ -40,7 +39,4 @@ const (
 const (
 	// FirstDepositIndex represents the index of the first deposit in the system, set at genesis.
 	FirstDepositIndex uint64 = 0
-
-	// GenesisVersion represents the fork version of the genesis block.
-	GenesisVersion = version.Deneb
```
calbera, line 45: replaced with usage `version.Genesis()`

primitives/version/versions.go
calbera: Kept in

primitives/version/comparable.go
```diff
@@ -0,0 +1,69 @@
...
+//		if version.Equals(a, b) {
+//			return 0
+//		}
+//      // a is after b
+//		return 1
+//	}
+
+// IsBefore returns true if a is before b. This compares bytes from most significant
+// to least significant in little-endian order.
+func IsBefore(a, b common.Version) bool {
```
shotes, line 51: Maybe I missed the discussion on this. Is there are reason this is `IsBefore(a, b)` instead of `a.IsBefore(b)`? I feel like the latter is way more readable. Is there an interface that this is implementing?

calbera: Agree on the readability, however would require adding these functions on the B4 type which opinionates sorting for _all_ B4 types. We wanted to enforce that this sort is strictly used for our fork versions.

primitives/version/comparable.go
calbera, line 1: Noting that these functions could be implemented by `bytes.Compare(a[:], b[:])`, which does yield the same results in our unit tests. The underlying impl looks something like [this](https://github.com/golang/go/blob/37f27fbecd422da9fefb8ae1cc601bc5b4fec44b/src/internal/bytealg/compare_generic.go#L55) at a high level, although uses the assembly version based on arch. 

Personally prefer keeping the implementation as is because we define how we want it to be sorted and its explicitly mentioned/documented here.

rezbera: I would lean towards `bytes.Compare()` if we have 1000% confidence that the comparison does not depend on machine arch endianness. Signs suggest it does not change based on machine arch

primitives/version/comparable.go
abi87: I don't mind either version but I think we could methods on the type alias without changing the base type.
So we could in principle define the methods over Version only.

See the following snippet,which does not compile
https://go.dev/play/p/OSrYK3p11Ue

primitives/version/comparable_test.go
```diff
@@ -0,0 +1,255 @@
...
+			a:        version.Phase0(),
+			b:        version.Altair(),
+			expected: true,
+		},
+		{
+			name:     "beacon version before minor",
+			a:        version.Deneb(),
+			b:        version.Deneb1(),
+			expected: true,
+		},
```
rezbera, line 79: Can we add test for Deneb1 is before Electra (test that failed last time)

rezbera (approved):
LGTM. wrt using byte.Compare(a,b) - it's a two way door given the consistent package interface so not fussed.

abi87 (commented):
nit but overall looks great to me

primitives/version/comparable_test.go
calbera: that test is below on line 81

engine-primitives/engine-primitives/attributes.go
```diff
@@ -112,7 +112,8 @@ func (p *PayloadAttributes) Validate() error {
 		return ErrEmptyPrevRandao
 	}
 
-	if p.Withdrawals == nil && p.version >= version.Capella {
+	// For any fork version after Bellatrix (Capella onwards), withdrawals are required.
+	if version.IsAfter(p.version, version.Bellatrix()) && p.Withdrawals == nil {
```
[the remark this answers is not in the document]
calbera: done

abi87 (approved):
looks great to me

Issue #2405, opened 2025-01-24
`ErrAcceptedPayloadStatus` special case handling


shotes:
Currently the `ACCEPTED` payload status (raised as `ErrAcceptedPayloadStatus` in beacon-kit) from `newPayload()` engine API calls is treated the same as a `VALID` payload status. In both cases, the payload is accepted by beacon-kit consensus as a valid payload.

However, the `ACCEPTED` payload status does not guarantee that the payload is valid. Here are the engine API spec requirements for returning `ACCEPTED`:
```
{status: ACCEPTED, latestValidHash: null, validationError: null} if the following conditions are met:
- all transactions have non-zero length
- the blockHash of the payload is valid
- the payload doesn't extend the canonical chain
- the payload hasn't been fully validated
- ancestors of a payload are known and comprise a well-formed chain.
```

Payloads that receive `ACCEPTED` are explicitly *not fully validated*, and we should not treat them as valid payloads until they are fully validated.

abi87:
@shotes is this still relevant? Looks closed to me

shotes:
https://github.com/berachain/beacon-kit/pull/2439 along with proposed modifications to engine api interactions in https://github.com/berachain/beacon-kit/pull/2508 lead me to believe that this is no longer a concern.

PR #2439, opened 2025-01-30
fix(engineAPI): error on ACCEPTED payload status

shotes:
This PR fixes the following:
- don't ignore ACCEPTED payload status
- remove ACCEPTED payload status handling for `forkchoiceUpdated` calls, since that is not a possible return status.


execution/engine/engine.go
```diff
@@ -89,12 +89,8 @@ func (ee *Engine) NotifyForkchoiceUpdate(
 	switch {
 	// We do not bubble the error up, since we want to handle it
 	// in the same way as the other cases.
-	case errors.IsAny(
-		err,
-		engineerrors.ErrAcceptedPayloadStatus,
-		engineerrors.ErrSyncingPayloadStatus,
```
abi87, line 95: The only reason I could imagine for putting these to together is if some EL could return them at the same time?

execution/engine/metrics.go
```diff
@@ -198,14 +198,14 @@ func (em *engineMetrics) markForkchoiceUpdateValid(
 	)
 }
 
-// markForkchoiceUpdateAcceptedSyncing increments
+// markForkchoiceUpdateSyncing increments
 // the counter for accepted syncing forkchoice updates.
-func (em *engineMetrics) markForkchoiceUpdateAcceptedSyncing(
+func (em *engineMetrics) markForkchoiceUpdateSyncing(
```
abi87, line 203: <3

abi87 (commented):
LGTM, will cross check accepted state in other EL

calbera (commented):
metrics for newpayload ACCEPTED vs SYNCING and then lgtm

execution/engine/engine.go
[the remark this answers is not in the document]
shotes: added in commit https://github.com/berachain/beacon-kit/pull/2439/commits/d063be4ba9d914f30975153ee1cdfa0a2f8ac8b4

execution/engine/engine.go
abi87: All good

abi87 (approved):
LGTM

shotes:
Related to https://github.com/berachain/beacon-kit/issues/2405

PR #2440, opened 2025-01-30
chore(types): Cleanup unused types/funcs and expose Service Registry

rezbera:
This cleans up unused types and funcs, and exposes the Service Registry which allows us to fetch the various services such as Comet or Blockchain Service from the ABCI, which will support the future testing framework


cli/commands/server/start.go
```diff
@@ -61,19 +61,6 @@ type StartCmdOptions[
 	AddFlags func(cmd *cobra.Command)
 }
 
-// StartCmd runs the service passed in, either stand-alone or in-process with
-// CometBFT.
-func StartCmd[
-	T interface {
```
rezbera, line 67: unused

cmd/beacond/types.go
```diff
@@ -97,46 +91,15 @@ type (
 /* -------------------------------------------------------------------------- */
 
 type (
-	// BeaconBlock type aliases.
-	ConsensusBlock = consruntimetypes.ConsensusBlock
-
```
rezbera, line 102: unused

config/config.go
```diff
@@ -86,16 +86,6 @@ func (c Config) Template() string {
 	return template.TomlTemplate
 }
 
-// MustReadConfigFromAppOpts reads the configuration options from the given
-// application options.
-func MustReadConfigFromAppOpts(opts AppOptions) *Config {
```
rezbera, line 91: unused

node-core/components/signer.go
```diff
@@ -58,6 +61,17 @@ func ProvideBlsSigner(in BlsSignerInput) (crypto.BLSSigner, error) {
...
 			privValStateFile = filepath.Join(homeDir, privValStateFile)
 		}
+
+		// Check key file existence here as the error in NewBLSSigner is vague.
+		if _, err := os.Stat(privValKeyFile); os.IsNotExist(err) {
+			return nil, fmt.Errorf("key file does not exist at path: %s", privValKeyFile)
+		}
+
+		// Check state file existence as the error in NewBLSSigner is vague.
+		if _, err := os.Stat(privValStateFile); os.IsNotExist(err) {
```
calbera, line 71: It isnt absolutely necessary to have the state file. See the CLI commands which use this ProvideSigner function to determine the BLS signer. They should all be able to run without state

execution/client/client.go
```diff
@@ -118,6 +118,8 @@ func (s *EngineClient) Start(
 			if err := s.verifyChainIDAndConnection(ctx); err != nil {
 				if errors.Is(err, ErrMismatchedEth1ChainID) {
 					s.logger.Error(err.Error())
+					// This is an unrecoverable error and should return
+					return err
```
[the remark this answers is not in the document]
rezbera: All the services will be Stopped if this Errors and the node will shut down

https://github.com/berachain/beacon-kit/blob/94b0e9981065c1719c69631e50c76951447bafa9/node-core/node/node.go#L71

node-core/components/signer.go
calbera: Basically all uses of ProvideSigner should allow not having state
- when node starts up
- CLI deposit commands
- etc

execution/client/client.go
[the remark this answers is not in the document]
calbera: got it, but this itself is triggered on node start [here](https://github.com/berachain/beacon-kit/blob/66bc722b629a387da7e602ba886843279c5c7374/node-core/node/node.go#L71), so the node will fail before even beginning processing. Which seems to be the intended behavior.

node-core/components/signer.go
calbera: EDIT: After investigating, the `LoadFilePV` that we use currently does indeed require having the state file so this check is safe to keep as is.

execution/client/client.go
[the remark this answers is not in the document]
abi87: So:
- currently we keep looping on this check BUT
- you're saying EL cannot fix this error without restart, so CL should err??

I think it could be fine like it is:
- if EL is on the wrong chain, you kill it, fix its config, restart and CL will pick up from there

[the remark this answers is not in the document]
rezbera: That a reasonable counter example @abi87 . Will remove the return.

cli/commands/server/start.go
abi87: This is a cli command, how do we check whether it is not used by any tooling? I understand it is not used inside beaconKit

node-core/components/signer.go
```diff
@@ -58,6 +61,17 @@ func ProvideBlsSigner(in BlsSignerInput) (crypto.BLSSigner, error) {
 		if !filepath.IsAbs(privValStateFile) {
 			privValStateFile = filepath.Join(homeDir, privValStateFile)
 		}
+
+		// Check key file existence here as the error in NewBLSSigner is vague.
```
abi87, line 65: q: we're saying we want to avoid the `os.Exit(1)` which `LoadFilePV` would cause right?

abi87 (commented):
I only have a doubt around `StartCmd`, other than that LGTM

cli/commands/server/start.go
abi87: Actually there is a `StartCmdWithOptions` so we are not really using anything as `StartCmd` is implemented in terms of `StartCmdWithOptions`

node-core/components/signer.go
rezbera: Correct - it does os.Exit(1) several times but this catches the most common reason

fridrik01 (approved):
lgtm

PR #2442, opened 2025-01-30
fix(flags): Fix Payload Builder Flags

rezbera:
This fix was required for the testing work as the flags are not used and do not have the correct mapstructure. This brings it in line with the yaml file format.

### DIFF
--- a/cli/flags/flags.go
+++ b/cli/flags/flags.go
@@ -31,10 +31,10 @@ const (
 	BeaconKitAcceptTos = beaconKitRoot + "accept-tos"
 
 	// Builder Config.
-	builderRoot              = beaconKitRoot + "payload-builder."
-	SuggestedFeeRecipient    = builderRoot + "suggested-fee-recipient"
-	LocalBuilderEnabled      = builderRoot + "local-builder-enabled"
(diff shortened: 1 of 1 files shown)


cli/flags/flags.go
```diff
@@ -31,10 +31,10 @@ const (
...
 
 	// Builder Config.
-	builderRoot              = beaconKitRoot + "payload-builder."
-	SuggestedFeeRecipient    = builderRoot + "suggested-fee-recipient"
-	LocalBuilderEnabled      = builderRoot + "local-builder-enabled"
-	LocalBuildPayloadTimeout = builderRoot + "local-build-payload-timeout"
+	builderRoot           = beaconKitRoot + "payload-builder."
+	SuggestedFeeRecipient = builderRoot + "suggested-fee-recipient"
+	BuilderEnabled        = builderRoot + "enabled"
+	BuildPayloadTimeout   = builderRoot + "payload-timeout"
```
rezbera, line 37: These flags were not used before. Hence the incorrect `local-builder-` went unnoticed.

abi87 (approved):

cli/flags/flags.go
rezbera: no need to drop them - i will use them for testing

PR #2443, opened 2025-01-31
refactor(deposits): Modularize Deposit Command for Easier Testing in Runtime

rezbera:
This PR contributes towards the [Godzilla PR](https://github.com/berachain/beacon-kit/pull/2411) by modularisiing the deposits function so that it can be realistically called in an active runtime. This is proven through the unit test added.

### DIFF
--- a/cli/commands/genesis/deposit.go
+++ b/cli/commands/genesis/deposit.go
@@ -28,7 +28,6 @@ import (
 	"github.com/berachain/beacon-kit/chain"
 	"github.com/berachain/beacon-kit/cli/context"
 	"github.com/berachain/beacon-kit/cli/utils/parser"
-	"github.com/berachain/beacon-kit/config"
 	"github.com/berachain/beacon-kit/consensus-types/types"
 	"github.com/berachain/beacon-kit/errors"
 	"github.com/berachain/beacon-kit/node-core/components"
@@ -68,16 +67,28 @@ func AddGenesisDepositCmd(cs chain.Spec) *cobra.Command {
 			cometConfig := context.GetConfigFromCmd(cmd)
 			appOpts := context.GetViperFromCmd(cmd)
 			outputDocument, _ := cmd.Flags().GetString(flags.FlagOutputDocument)
-			return AddGenesisDeposit(cs, cometConfig, appOpts, depositAmount, withdrawalAddress, outputDocument)
+
+			// Get the BLS signer.
+			blsSigner, err := components.ProvideBlsSigner(
+				components.BlsSignerInput{
+					AppOpts: appOpts,
+				},
+			)
+			if err != nil {
+				return err
+			}
+
+			return AddGenesisDeposit(cs, cometConfig, blsSigner, depositAmount, withdrawalAddress, outputDocument)
 		},
 	}
 	return cmd
 }
 
+// AddGenesisDeposit is the modularized version of AddGenesisDepositCmd that can be properly tested from within the runtime
 func AddGenesisDeposit(
 	cs chain.Spec,
 	cometConfig *cmtcfg.Config,
-	appOpts config.AppOptions,
+	blsSigner crypto.BLSSigner,
 	depositAmount math.Gwei,
 	withdrawalAddress common.ExecutionAddress,
 	outputDocument string,
@@ -90,16 +101,6 @@ func AddGenesisDeposit(
 			err,
 			"failed to initialize commands validator files",
 		)
-	}
-
-	// Get the BLS signer.
-	blsSigner, err := components.ProvideBlsSigner(
-		components.BlsSignerInput{
-			AppOpts: appOpts,
-		},
-	)
-	if err != nil {
-		return err
 	}
 
 	// All deposits are signed with the genesis version.
--- a/cli/commands/genesis/deposit.go
+++ b/cli/commands/genesis/deposit.go
@@ -77,14 +77,13 @@ func AddGenesisDepositCmd(cs chain.Spec) *cobra.Command {
 			if err != nil {
 				return err
 			}
-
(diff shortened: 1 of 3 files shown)


cli/commands/genesis/deposit.go
```diff
@@ -55,88 +56,97 @@ func AddGenesisDepositCmd(cs chain.Spec) *cobra.Command {
...
 			blsSigner, err := components.ProvideBlsSigner(
 				components.BlsSignerInput{
-					AppOpts: context.GetViperFromCmd(cmd),
+					AppOpts: appOpts,
 				},
 			)
 			if err != nil {
 				return err
 			}
+			return AddGenesisDeposit(cs, cometConfig, blsSigner, depositAmount, withdrawalAddress, outputDocument)
```
rezbera, line 80: The goal is to extract all the user provided fields so that we don't need Viper embedded in our core logic loop as that's not realistically testable

testing/utils/homedir.go
```diff
@@ -0,0 +1,52 @@
...
+	bytes := make([]byte, rndSuffixLen)
+	_, err := rand.Read(bytes)
+	require.NoError(t, err)
+
+	rndSuffix := hex.EncodeToString(bytes)
+
+	tmpFolder := filepath.Join(os.TempDir(), "/injected-consensus", rndSuffix)
+	require.NoError(t, os.MkdirAll(tmpFolder, 0750))
+	return tmpFolder
+}
```
fridrik01, line 46: You can replace this with https://pkg.go.dev/testing#T.TempDir or https://pkg.go.dev/os#MkdirTemp which takes care of making the folder unique

cli/commands/genesis/deposit.go
```diff
@@ -55,88 +56,102 @@ func AddGenesisDepositCmd(cs chain.Spec) *cobra.Command {
 		Long:  `Adds a validator to the genesis file with the necessary credentials. The arguments are expected in the order of the deposit amount and withdrawal address.`,
 		Args:  cobra.ExactArgs(2), //nolint:mnd // The number of arguments.
 		RunE: func(cmd *cobra.Command, args []string) error {
-			config := context.GetConfigFromCmd(cmd)
-
-			_, valPubKey, err := genutil.InitializeNodeValidatorFiles(
-				config, crypto.CometBLSType,
-			)
+			// Get the deposit amount.
+			var depositAmount math.Gwei
```
abi87, line 60: nit: you don't need `var depositAmount math.Gwei`

testing/utils/homedir.go
rezbera: great catch - will remove this file

cli/commands/genesis/deposit.go
rezbera: fixed

[Remove explicit depositAmount var](https://github.com/berachain/beacon-kit/pull/2443/commits/ffa5b07099ce1f4de54ca0c255110aab49a9b25a)

rezbera: I wonder if there's a lint rule for this

cli/commands/genesis/deposit.go
```diff
@@ -55,88 +56,102 @@ func AddGenesisDepositCmd(cs chain.Spec) *cobra.Command {
...
+		Pubkey:      depositMsg.Pubkey,
+		Amount:      depositMsg.Amount,
+		Signature:   signature,
+		Credentials: depositMsg.Credentials,
+	}
 
-			if err = writeDepositToFile(outputDocument, &deposit); err != nil {
-				return errors.Wrap(err, "failed to write signed gen tx")
-			}
+	//#nosec:G703 // Ignore errors on this line.
```
abi87, line 140: it looks this is not needed

cli/commands/genesis/deposit_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/cometbft/cometbft/crypto/bls12381"
+	"github.com/cometbft/cometbft/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGenesisDeposit(t *testing.T) {
+	homeDir := t.TempDir()
+	t.Log("Home folder:", homeDir)
+
+	chainSpec, err := spec.DevnetChainSpec()
```
abi87, line 42: nit: I don't think we have an agreement on this, but I would suggest to at least consider using mainnet specs here.
I reckon it hardly makes a difference for this test

cli/commands/genesis/deposit_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	blsSigner := signer.BLSSigner{PrivValidator: types.NewMockPVWithKeyType(bls12381.KeyType)}
+
+	err = genesis.AddGenesisDeposit(chainSpec, cometConfig, blsSigner, depositAmount, withdrawalAdress, outputDocument)
+	require.NoError(t, err)
+
+	require.FileExists(t, path.Join(homeDir, "nodekey.json"))
+	require.FileExists(t, path.Join(homeDir, "data", "priv_validator_state.json"))
+	require.FileExists(t, path.Join(homeDir, "config", "priv_validator_key.json"))
+	require.DirExists(t, path.Join(homeDir, "config", "premined-deposits"))
```
abi87, line 64: question: should we assert the content of these files?

cli/commands/genesis/deposit.go
abi87: AFAIK we err if we shadow errors, which is why we are often pushed to use `var`

abi87 (commented):
nits and a question around test assertion.
Other than that, LGTM

cli/commands/genesis/deposit_test.go
abi87: good enough for me

abi87 (approved):

PR #2445, opened 2025-01-31
chain-spec: Add validation for withdrawal sweep limits

VolodymyrBg:
This PR adds important validation checks for the MaxValidatorsPerWithdrawalsSweep configuration parameters in the chain specification. The changes ensure that:

1. Neither pre-upgrade nor post-upgrade withdrawal sweep limits can be zero
2. The post-upgrade withdrawal sweep limit cannot be larger than the pre-upgrade limit, maintaining backwards compatibility
3. The withdrawal sweep limits cannot exceed the validator registry limit

These validations prevent potential misconfiguration that could lead to invalid states or unexpected behavior in the validator withdrawal process.

Changes:
- Added validation checks in chainSpec.validate() method
- Added appropriate error messages for each validation case
- Ensures withdrawal sweep limits are properly bounded

Testing:
The existing test suite covers the validation functionality. No additional tests were needed as the changes are enforced through the existing validation mechanism.


abi87:
This is the nth attempt over the same kind of change. I really don't see what is this for

PR #2450, opened 2025-01-31
fix: remove redundant *LoggerConfig type from function signature

0xwitty:
I noticed that the function `DefaultComponents()` is declared to return three types: `Node`, `*Logger`, and `*LoggerConfig`. However, the implementation only returns `Node` and `*Logger`.

The `*LoggerConfig` type is unnecessary and should be removed from the function signature to avoid confusion and keep the code consistent.


abi87:
Hello @0xwitty thanks for this. We have plans for a larger scale cleanup of the logger (see https://github.com/berachain/beacon-kit/pull/2460 for an example).
Closing this and deferring the config cleanup for a PR to come

PR #2451, opened 2025-01-31
chore(state-transition): flipped transaction context attributes logic

abi87:
most of the attributes in `transaction.Context` have negated logic which does not feel natural to me.
This PR flips that, so that we would have `VerifyPayload`  instead of `SkipPayloadVerification`  etc


primitives/transition/context.go
```diff
@@ -35,15 +35,15 @@ type Context struct {
...
-	// SkipPayloadVerification indicates whether to skip calling NewPayload
-	// on the execution client. This can be done when the node is not
+	// VerifyPayload indicates whether to call NewPayload on the
+	// execution client. This can be done when the node is not
 	// syncing, and the payload is already known to the execution client.
-	SkipPayloadVerification bool
-	// SkipValidateRandao indicates whether to skip validating the Randao mix.
-	SkipValidateRandao bool
-	// SkipValidateResult indicates whether to validate the result of
+	VerifyPayload bool
```
shotes, line 41: I know this isn't directly the topic of this PR, but I wanted to bring it up - if we are only ever setting these values during construction and always using the getter, should these all be non-exported values?

state-transition/core/types.go
```diff
@@ -41,16 +41,15 @@ type Context interface {
...
-	GetSkipPayloadVerification() bool
-	// GetSkipValidateRandao returns whether to skip validating the RANDAO
-	// reveal.
-	GetSkipValidateRandao() bool
-	// GetSkipValidateResult returns whether to validate the result of the state
-	// transition.
-	GetSkipValidateResult() bool
+	// GetVerifyPayload returns whether to verify the payload
+	// if it already exists on the execution client.
+	GetVerifyPayload() bool
```
shotes, line 46: While we're at it, can we rename to something that is not `Get...()`? Like `ShouldVerifyPayload()` or `DoVerifyPayload()` or just `VerifyPayload()`.

shotes (commented):
nits, otherwise LGTM

state-transition/core/types.go
abi87: I cannot do that, it would conflict with the attributes name

primitives/transition/context.go
abi87: I would be down to get rid of the getters, but to do that we have to drop the interfaces. Let me have a look

beacon/blockchain/finalize_block.go
```diff
@@ -177,13 +177,13 @@ func (s *Service) executeStateTransition(
 			// of validators in their process proposal call and thus
 			// the "verification aspect" of this NewPayload call is
 			// actually irrelevant at this point.
-			SkipPayloadVerification: false,
+			VerifyPayload: true,
```
rezbera, line 180: Is the comment above accurate?

primitives/transition/context.go
abi87: @shotes  I have:

- Get rid of Context generic from state processor
- Drop getters and use Exported values only.

How does this look to you?

beacon/blockchain/finalize_block.go
abi87: I believe so. It's also showing that we are being inefficient on purpose, which I really do not understand.
Do you see anything wrong?

rezbera: You're right. My mistake

state-transition/core/types.go
abi87: Dropped the getter

primitives/transition/context.go
```diff
@@ -28,71 +28,25 @@ import (
 
 // Context is the context for the state transition.
 type Context struct {
-	context.Context
+	ConsensusCtx context.Context
```
rezbera, line 31: Would be good to explicitly comment that this is the context passed from CometBFT (and is currently implemented as `context.TODO()`

abi87: @rezbera done, LMK if the comments are clear enough

state-transition/core/state_processor.go
```diff
@@ -37,7 +37,7 @@ import (
 
 // StateProcessor is a basic Processor, which takes care of the
 // main state transition for the beacon chain.
-type StateProcessor[ContextT Context] struct {
+type StateProcessor struct {
```
calbera, line 40: If we're fixing this I love the removal of generics. But I would prefer using the `transition.Context` as an interface so that the fields of the struct are not mutable in the stateProcessor. Thoughts?

primitives/transition/context.go
calbera: strongly in favor of this. StateProcessor should not be able to modify, only read from imo

state-transition/core/state_processor.go
abi87: cc @berachain/core

rezbera (approved):
This is net improvement as is - lgtm

nidhi-singh02 (approved):
lgtm!

primitives/transition/context.go
fridrik01: oh I didn't know that, so it basically wont ever cancel

state-transition/core/state_processor.go
fridrik01: +1 for un-exporting the struct fields initialized via `NewXXX` method and having simple read only getters.

fridrik01 (approved):
lgtm, prefer the unexported and read-only getters

state-transition/core/state_processor.go
abi87: Added encapsulation, LMK how do you like this

primitives/transition/context.go
```diff
@@ -28,71 +28,116 @@ import (
...
+		meterGas: false,
+
+		// by default we don't have optimistic engine
+		// as it basically mute some checks
+		optimisticEngine: false,
+
+		// by default we keep all verification
+		verifyPayload: true,
+		verifyRandao:  true,
+		verifyResult:  true,
```
abi87, line 82: I rather set all validators to true by default. I think the safest is to explicitly switch them off

state-transition/core/types.go
```diff
@@ -33,32 +33,6 @@ import (
 	"github.com/karalabe/ssz"
 )
 
-// Context defines an interface for managing state transition context.
-type Context interface {
```
calbera, line 37: I think this Context interface should still be used so that only the read-only methods are usable in this package. Currently the Setters can still be called from within state-transition

testing/state-transition/state-transition.go
```diff
@@ -134,12 +134,16 @@ func SetupTestState(t *testing.T, cs chain.Spec) (
...
-		ProposerAddress:         dummyProposerAddr,
-	}
+	ctx := transition.NewTransitionCtx(
+		context.Background(),
+		0, // time
+		dummyProposerAddr,
+	).
+		WithVerifyPayload(false).
+		WithVerifyRandao(false).
+		WithVerifyResult(false).
```
abi87, line 144: technically `WithVerifyRandao` should be true cause SkiValidateRandao was set to the default value of false.
However the intention of the tests was to skip these checks at all and focus on the state transition, so it's ok to not verify randao (and this PR makes it more explicit that it currently is)

state-transition/core/types.go
abi87: Added a `transition.ReadOnlyContext` interface

primitives/transition/context.go
```diff
@@ -26,73 +26,130 @@ import (
 	"github.com/berachain/beacon-kit/primitives/math"
 )
 
+// ReadOnlyContext defines an interface for managing state transition context.
+type ReadOnlyContext interface {
```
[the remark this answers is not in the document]
abi87: Is this really necessary when we have only packages and not modules? I think a single interface reduced code duplication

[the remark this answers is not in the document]
calbera: not absolutely necessary. I generally like following this pattern and is used throughout the repo https://medium.com/@mbinjamil/using-interfaces-in-go-the-right-way-99384bc69d39

primitives/transition/context.go
```diff
@@ -28,71 +28,116 @@ import (
...
+}
+
+func (c *Context) WithVerifyRandao(verifyRandao bool) *Context {
+	c.verifyRandao = verifyRandao
+	return c
+}
+
+func (c *Context) WithVerifyResult(verifyResult bool) *Context {
+	c.verifyResult = verifyResult
+	return c
```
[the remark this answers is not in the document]
fridrik01: In this case we may use opts pattern instead as otherwise we could update `Context` after initialization. Not a big deal though if we use readonly interface for accessing fields

fridrik01 (approved):

primitives/transition/context.go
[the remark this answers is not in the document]
calbera: From https://go.dev/wiki/CodeReviewComments: 
> Go interfaces generally belong in the package that uses values of the interface type, not the package that implements those values... Do not define interfaces before they are used: without a realistic example of usage, it is too difficult to see whether an interface is even necessary, let alone what methods it ought to contain.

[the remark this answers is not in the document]
abi87: So I dropped the interface from `transition.Context` file and ended up having a single one in `state-processor`.
Basically the only use of this interface is using this as parameter in Transition method, so I think this is fine

calbera (approved):
LGTM

PR #2455, opened 2025-02-01
chore(tests): Add t.Parallel() to Tests and CI checks

rezbera:
This should be up CI time. 

Contains https://github.com/berachain/beacon-kit/pull/2461

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -38,7 +38,7 @@ linters-settings:
     # Ignore missing calls to `t.Parallel()` in subtests. Top-level tests are
     # still required to have `t.Parallel`, but subtests are allowed to skip it.
     # Default: false
-    ignore-missing-subtests: false
+    ignore-missing-subtests: true
 
   exhaustive:
     # Program elements to check for exhaustiveness.
@@ -301,7 +301,7 @@ linters:
     - testableexamples # checks if examples are testable (have an expected output)
     - testifylint # checks usage of github.com/stretchr/testify
     - testpackage # makes you use a separate _test package
-    - tparallel # detects inappropriate usage of t.Parallel() method in your Go test codes
+#    - tparallel # detects inappropriate usage of t.Parallel() method in your Go test codes. Replaced with paralleltest
     - unconvert # removes unnecessary type conversions
     - unparam # reports unused function parameters
     - usestdlibvars # detects the possibility to use variables/constants from the Go standard library
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -31,6 +31,15 @@ linters-settings:
     # Default: false
     check-type-assertions: true
 
+  paralleltest:
+    # Ignore missing calls to `t.Parallel()` and only report incorrect uses of it.
+    # Default: false
+    ignore-missing: false
+    # Ignore missing calls to `t.Parallel()` in subtests. Top-level tests are
+    # still required to have `t.Parallel`, but subtests are allowed to skip it.
+    # Default: false
+    ignore-missing-subtests: false
+
   exhaustive:
     # Program elements to check for exhaustiveness.
     # Default: [ switch ]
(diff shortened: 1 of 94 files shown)


cli/commands/genesis/storage_test.go
```diff
@@ -36,14 +36,17 @@ import (
...
 	t.Run("command should be available and have correct use", func(t *testing.T) {
+		t.Parallel()
 		chainSpec, err := spec.DevnetChainSpec()
 		require.NoError(t, err)
 		cmd := genesis.SetDepositStorageCmd(chainSpec)
 		require.Equal(t, "set-deposit-storage [eth/genesis/file.json]", cmd.Use)
 	})
 
 	t.Run("should set deposit storage correctly", func(t *testing.T) {
+		t.Parallel()
```
abi87, line 49: question: are we guaranteed that tmp dirs are different for UTs running in parallel?

consensus-types/types/body_test.go
```diff
@@ -187,6 +198,8 @@ func TestBeaconBlockBody_UnusedAttesterSlashingsEnforcement(t *testing.T) {
 
 // Ensure that the Attestations field cannot be unmarshaled with data in it,
 // enforcing that it's unused.
+//
+//nolint:paralleltest // data race in ssz lib
```
abi87, line 202: what is this for?

da/store/store_test.go
```diff
@@ -43,16 +42,14 @@ func setSlot(scs datypes.BlobSidecars, slot math.Slot) {
 }
 
 func TestStore_PersistRace(t *testing.T) {
+	t.Parallel()
 	// This test case needs to be run with the '-race' flag
-	tmpFilePath := "/tmp/store_test"
+	tmpFilePath := t.TempDir()
```
abi87, line 47: good

engine-primitives/engine-primitives/withdrawals_test.go
```diff
@@ -31,7 +31,9 @@ import (
 )
 
 func TestWithdrawals(t *testing.T) {
+	t.Parallel()
```
abi87, line 34: why do we need `t.Parallel` in both scopes?

.golangci.yaml
```diff
@@ -333,7 +342,7 @@ linters:
     - maintidx # measures the maintainability index of each function
     - misspell # [useless] finds commonly misspelled English words in comments
     #- nlreturn # [too strict and mostly code is not more readable] checks for a new line before return and branch statements to increase code clarity
-    #- paralleltest # [too many false positives] detects missing usage of t.Parallel() method in your Go test
```
abi87, line 336: IIUC we have only one false positive. I wonder why is this comment so off. Maybe has paralleltest tool improved?

engine-primitives/engine-primitives/withdrawals_test.go
rezbera: The `t.Parallel()` will ensure it runs in parallel with other unit tests outside of this scope. 
```go
func TestWithdrawals(t *testing.T) {
	t.Parallel()
```

The `t.Parallel()` below will allow the subtests to be run in parallel
```go
	t.Run("SizeSSZ", func(t *testing.T) {
		t.Parallel()
```

cli/commands/genesis/storage_test.go
rezbera: If we `t.TempDir()` across tests (unlike this one), then all of them are guaranteed to be unique as the dir name is based on the name:
```go
// Name returns the name of the running (sub-) test or benchmark.
//
// The name will include the name of the test along with the names of
// any nested sub-tests. If two sibling sub-tests have the same name,
// Name will append a suffix to guarantee the returned name is unique.
func (c *common) Name() string {
	return c.name
}
```

Fixed this test in https://github.com/berachain/beacon-kit/commit/bbfe9297f8376d4c7c3ccb2e30a4cc9412be1cb6

abi87 (approved):
LGTM

PR #2456, opened 2025-02-01
chore(api): Dropped node api context generic

abi87:
Cleaned up the NodeAPIs context.
Apologies to @fridrik01 for stealing this


node-api/handlers/debug/state.go
```diff
@@ -46,8 +47,8 @@ func (h *Handler[ContextT]) GetState(c ContextT) (any, error) {
 	}
 	return beacontypes.StateResponse{
 		// TODO: The version should be retrieved based on the slot
-		Version:             "deneb", // stubbed
```
abi87, line 49: to avoid this, we should introduce:
- a `latest version`
- a version.String method

cc @calbera

node-api/engines/echo/types.go
```diff
@@ -1,27 +0,0 @@
...
-// EXPRESS OR IMPLIED, INCLUDING (WITHOUT LIMITATION) WARRANTIES OF
-// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, NON-INFRINGEMENT, AND
-// TITLE.
-
-package echo
-
-import "github.com/labstack/echo/v4"
-
-type (
-	Context = echo.Context
```
abi87, line 26: moved to handler package

node-api/handlers/handlers.go
```diff
@@ -23,57 +23,56 @@ package handlers
 import (
 	"github.com/berachain/beacon-kit/errors"
 	"github.com/berachain/beacon-kit/log"
+	"github.com/labstack/echo/v4"
 )
 
+type Context = echo.Context
```
abi87, line 29: moved here from echo package to avoid import cycle

node-api/handlers/node/placeholders.go
```diff
@@ -47,15 +49,15 @@ func (h *Handler[ContextT]) Syncing(ContextT) (any, error) {
...
+func (h *Handler) Version(handlers.Context) (any, error) {
 	type VersionResponse struct {
 		Data struct {
 			Version string `json:"version"`
 		} `json:"data"`
 	}
 
 	response := VersionResponse{}
-	response.Data.Version = "1.0.0"
+	response.Data.Version = "1.1.0"
```
abi87, line 60: updated here just to signal that we should fix this

node-api/handlers/debug/state.go
calbera: makes sense. `latest` version might not be necessary since we want to provide the version at which the incoming request is queried for. Would probably just require returning `chainSpec.ActiveForkVersionForSlot(requestedSlot).String()`

rezbera (approved):
LGTM - Much needed

PR #2463, opened 2025-02-03
feat(validator): Implement validator statistics endpoint

avorylli:
Added ValidatorStats method and ValidatorStatsData type to provide basic validator statistics for monitoring purposes.

### DIFF
--- a/node-api/handlers/beacon/backend.go
+++ b/node-api/handlers/beacon/backend.go
@@ -85,4 +85,9 @@ type ValidatorBackend interface {
 		slot math.Slot,
 		ids []string,
 	) ([]*types.ValidatorBalanceData, error)
+	// ValidatorStats returns aggregated statistics for validators in the given epoch
+	ValidatorStats(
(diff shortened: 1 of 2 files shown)


calbera:
@avorylli which beacon api endpoint is this implementing? Also the route is not specified in the corresponding routes.go file

avorylli:
> @avorylli which beacon api endpoint is this implementing? Also the route is not specified in the corresponding routes.go file

@calbera 

Ah, you're right - I forgot to specify which endpoint.

I was thinking of adding this as a custom endpoint like `/eth/v1/beacon/custom/validator_stats`, but maybe there's a better way to fit this into an existing beacon API endpoint? Let me know which way you'd prefer to go with this.

calbera:
I would prefer closing this PR if it doesn't implement an endpoint in the [beacon spec](https://ethereum.github.io/beacon-APIs/). Unless there is a specific need for an endpoint let's keep this out for now. Thanks!

PR #2467, opened 2025-02-03
fix(node-api): Implement validators endpoint filtering properly

shotes:
Tested these two endpoints  -
https://ethereum.github.io/beacon-APIs/#/Beacon/getStateValidators
https://ethereum.github.io/beacon-APIs/#/Beacon/postStateValidators

- Refactored code for reusability as part of this change-set.
- Increased go-eth2 client version.

Testing other validator endpoints, will open separate PR as don't want to increase diff.


node-api/handlers/beacon/backend.go
```diff
@@ -76,10 +76,10 @@ type ValidatorBackend interface {
 	ValidatorByID(
 		slot math.Slot, id string,
 	) (*types.ValidatorData, error)
-	ValidatorsByIDs(
+	FilteredValidators(
 		slot math.Slot,
 		ids []string,
-		statuses []string,
+		statusees []string,
```
calbera, line 82: nit: typo

calbera (commented):
Changes look mostly good. Can we write 2 E2E tests querying validators endpoint. 1 with empty `IDs` & another with empty `Statuses` filters and show both calls return all validators?

node-api/backend/utils/validator.go
```diff
@@ -41,3 +45,51 @@ func ValidatorIndexByID(st *statedb.StateDB, keyOrIndex string) (math.U64, error
 	}
 	return st.ValidatorIndexByPubkey(key)
 }
+
+// GetValidatorStatus returns the current validator status based on its set
+// Epoch values.
+func GetValidatorStatus(epoch math.Epoch, validator *types.Validator) (string, error) {
```
calbera, line 51: i understand this is from Prysm, but I feel either we should have a sanity check UT for this function here OR put these functions on the Validator type and UTs there

node-api/handlers/beacon/validators.go
```diff
@@ -60,34 +53,24 @@ func (h *Handler) GetStateValidators(c handlers.Context) (any, error) {
...
+func (h *Handler) PostStateValidators(c handlers.Context) (any, error) {
+	req, err := utils.BindAndValidate[beacontypes.PostStateValidatorsRequest](
+		c, h.Logger(),
 	)
 	if err != nil {
 		return nil, err
 	}
-	return beacontypes.ValidatorResponse{
-		ExecutionOptimistic: false, // stubbed
-		Finalized:           false, // stubbed
```
calbera, line 88: can we introduce the logic for correctly returning these 2 fields as well? 
- I believed `Finalized` should always return true since the queried for slot (even "head") is guaranteed to be finalized by cometbft. Any raw slot # ahead of "head" should error out as its in the future anyways.
- Correspondingly `ExecutionOptimistic` I believe should always be false since we don't return data for any blocks that are not already Finalized.

nidhi-singh02:
> Changes look mostly good. Can we write 2 E2E tests querying validators endpoint. 1 with empty `IDs` & another with empty `Statuses` filters and show both calls return all validators?

Yes, introduced E2E tests with various combinations - both positive and negative scenarios.

node-api/backend/utils/validator.go
nidhi-singh02: Added `Status` function on Validator type and added unit tests.

node-api/handlers/beacon/validators.go
nidhi-singh02: yes. Changed `Finalized` to true and `ExecutionOptimistic` was false earlier.

consensus-types/types/validator.go
```diff
@@ -317,3 +318,50 @@ func (v Validator) GetWithdrawableEpoch() math.Epoch {
 func (v Validator) GetWithdrawalCredentials() WithdrawalCredentials {
 	return v.WithdrawalCredentials
 }
+
+// Status returns the current validator status based on its set epoch values.
+func (v *Validator) Status(currentEpoch math.Epoch) (string, error) {
```
calbera, line 323: nice!

abi87: Consider restructuring the logic as follows:
```
	switch {
	case currentEpoch < v.GetActivationEligibilityEpoch():
		return constants.ValidatorStatusPendingInitialized, nil
	case currentEpoch < v.GetActivationEpoch():
		return constants.ValidatorStatusPendingQueued, nil
	case currentEpoch < v.GetExitEpoch():
		if v.IsSlashed() {
			return constants.ValidatorStatusActiveSlashed, nil
		}
		return constants.ValidatorStatusActiveOngoing, nil
	case currentEpoch < v.GetExitEpoch():
		if v.IsSlashed() {
			return constants.ValidatorStatusExitedSlashed, nil
		}
		return constants.ValidatorStatusExitedUnslashed, nil
	default:
		if v.GetEffectiveBalance() != math.Gwei(0) {
			return constants.ValidatorStatusWithdrawalPossible, nil
		}
		return constants.ValidatorStatusWithdrawalDone, nil
	}
```

abi87: Also note that we don't slash currently so we could skip the `IsSlashed` checks and possibly err if this happens (to make sure we won't forget updating the APIs if/when we introduce slashing)

node-api/backend/validator.go
```diff
@@ -21,11 +21,75 @@
...
+
+	validators, err := st.GetValidators()
+	if err != nil {
+		return nil, err
+	}
+
+	// Filter on validator indexes and statuses.
+	validatorData := make([]*beacontypes.ValidatorData, 0, len(validators))
+	for _, validator := range validators {
+		// Skip the validator if we are filtering by indicies and this validator is not included.
```
abi87, line 59: type maybe? `indicies of this validator`

node-api/backend/validator.go
```diff
@@ -21,11 +21,75 @@
...
 )
 
+// FilteredValidators will grab all of the validators from the state at the
+// given slot. It will then filter them by the provided ids and statuses.
+func (b Backend) FilteredValidators(
+	slot math.Slot, ids []string, statuses []string,
+) ([]*beacontypes.ValidatorData, error) {
+	st, _, err := b.stateFromSlot(slot)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
abi87: Another bit, should we provide a better error if the slot requested is larger than current slot know by this block?

node-api/backend/validator.go
```diff
@@ -21,11 +21,75 @@
...
+	}
+
+	// Convert requested ids (can be validator index or pubkey) to validator index only.
+	validatorIndicies := make([]uint64, 0, len(ids))
+	for _, id := range ids {
+		validatorIndex, vErr := utils.ValidatorIndexByID(st, id)
```
abi87, line 44: I wonder if we should treat pubKey and index cases differently? This seems wasterful
If we have pubKeys, we don't need to do:
- pubKey -> index
- validator -> pubKey -> index -> check if the index is in the required list

We can load the validators directly by pubKey and then retrieve their index to be serialized in the response

node-api/engines/echo/validator.go
```diff
@@ -153,15 +153,15 @@ func ValidateRoot(value string) bool {
...
-		"pending_initialized": true,
-		"pending_queued":      true,
-		"active_ongoing":      true,
-		"active_exiting":      true,
-		"active_slashed":      true,
-		"exited_unslashed":    true,
-		"exited_slashed":      true,
-		"withdrawal_possible": true,
-		"withdrawal_done":     true,
+		utils.ActiveExiting:      true,
```
abi87, line 156: nit: is utils the right package to list validators statuses?

node-api/handlers/beacon/types/response.go
```diff
@@ -30,15 +30,26 @@ type GenericResponse struct {
 	Data                any  `json:"data"`
 }
 
-type ValidatorResponse struct {
+type Response struct {
```
abi87, line 33: This `Response` type is identical to `GenericResponse` above. Why do we need both?

node-api/handlers/beacon/validators.go
```diff
@@ -23,71 +23,47 @@ package beacon
...
-		return nil, err
-	}
-	// TODO: remove this once status filter is implemented.
-	if len(req.Statuses) > 0 {
-		return nil, types.ErrNotImplemented
-	}
-	slot, err := utils.SlotFromStateID(req.StateID, h.backend)
+// getStateValidators is a helper function to provide implementation
+// consistency between GetStateValidators and PostStateValidators, since they
+// are intended to behave the same way.
```
abi87, line 31: <3

node-api/handlers/beacon/header.go
```diff
@@ -41,18 +41,14 @@ func (h *Handler) GetBlockHeaders(c handlers.Context) (any, error) {
...
-		ExecutionOptimistic: false, // stubbed
-		Finalized:           false, // stubbed
-		Data: &beacontypes.BlockHeaderResponse{
-			Root:      header.GetBodyRoot(),
-			Canonical: true,
-			Header: &beacontypes.SignedBeaconBlockHeader{
-				Message:   beacontypes.BeaconBlockHeaderFromConsensus(header),
-				Signature: "", // TODO: implement
-			},
+	return beacontypes.NewResponse(&beacontypes.BlockHeaderResponse{
```
abi87, line 44: so here we just wrapped the `beacontypes.BlockHeaderResponse` in the `NewResponse` here

node-api/handlers/beacon/header.go
```diff
@@ -70,16 +66,12 @@ func (h *Handler) GetBlockHeaderByID(c handlers.Context) (any, error) {
...
```
abi87, line 69: so here we just wrapped the `beacontypes.BlockHeaderResponse` in the `NewResponse` here. No introducing a new TODO here

testing/e2e/e2e_staking_test.go
```diff
@@ -115,12 +115,14 @@ func (s *BeaconKitE2ESuite) TestDepositRobustness() {
...
-	s.Require().NotNil(apiClient.BeaconKitNodeClient)
-	response, err := apiClient.BeaconKitNodeClient.Validators(s.Ctx(), &api.ValidatorsOpts{
+	response, err := apiClient.Validators(s.Ctx(), &api.ValidatorsOpts{
 		State:   "genesis",
 		Indices: []phase0.ValidatorIndex{0, 1, 2, 3, 4},
 	})
 	s.Require().NoError(err)
+	s.Require().NotNil(response, "Validators response should not be nil")
+	s.Require().NotNil(response.Data, "Validators data should not be nil")
+
```
abi87, line 125: this is the same check we are doing in line 127 right? One can be removed

testing/e2e/suite/types/beacon_client.go
```diff
@@ -84,6 +91,14 @@ type BeaconKitNodeClient interface {
 	// Other beacon-kit node-api methods here...
 }
 
+// CustomBeaconClient is a custom implementation of the BeaconKitNodeClient interface
```
abi87, line 94: Can you clarify what is this for?
Should we instead consider returning Eth only forks and not providing "internal" ones like deneb1?
I would be open to that if it simplifies the whole thing

testing/e2e/e2e_api_test.go
```diff
@@ -48,3 +57,191 @@ func (s *BeaconKitE2ESuite) TestBeaconAPIStartup() {
...
+		&beaconapi.ValidatorsOpts{
+			State:   utils.StateIDHead,
+			Indices: indices,
+		},
+	)
+	s.Require().NoError(err)
+	s.Require().NotNil(validatorsResp)
+
+	validatorData := validatorsResp.Data
+	s.Require().NotNil(validatorData, "Validator data should not be nil")
```
abi87, line 77: these two checks are the same. We should drop one of them

testing/e2e/e2e_api_test.go
```diff
@@ -48,3 +57,191 @@ func (s *BeaconKitE2ESuite) TestBeaconAPIStartup() {
...
+	s.Require().
+		Len(validatorData, len(indices), "Number of validator responses should match number of requested indices")
+
+	for _, validator := range validatorData {
```
abi87, line 81: question: if these are genesis validators, should we compare the response against them and make sure they match?
All these checks are pretty loose in my opinion and the test as is does not show completeness (is the API returning all validators that it should)?

abi87 (dismissed):
I think this PR needs some code cleanup and more testing above all.
Also we may reconsider the implementation of `FilteredValidators` for efficiency

consensus-types/types/validator.go
calbera: @shotes I think you originally brought this function in from Prysm. Are one of these 2 the original sources? Either way we should probably include the source in the function doc comment.
- https://github.com/prysmaticlabs/prysm/blob/develop/beacon-chain/rpc/prysm/v1alpha1/validator/status.go#L406
- https://github.com/prysmaticlabs/prysm/blob/develop/beacon-chain/rpc/prysm/v1alpha1/beacon/validators.go#L797

Also I am indifferent to the restructuring as each case needs to be handled individually; okay keeping slashing case in now as it will become useful eventually.

node-api/handlers/utils/constants.go
```diff
@@ -34,3 +34,15 @@ const (
 	Head math.Slot = iota
 	Genesis
 )
+
+const (
+	ActiveExiting      = "active_exiting"
```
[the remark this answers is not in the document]
calbera: Yes I think this can be removed.

node-api/handlers/beacon/types/response.go
calbera: Would agree. I think we can just return GenericResponse from `NewResponse(data any)` and get rid of the `ValidatorResponse` struct

node-api/engines/echo/validator.go
calbera: let's use the one from primitives/constants/validator.go here

node-api/backend/validator.go
calbera: Correct me if I'm wrong, but I think the only typo is `indicies` should be spelled `indices`. The comment is saying to skip this validator if 1) we filter by indices AND 2) this validator is not included.

testing/e2e/e2e_api_test.go
calbera: I would say checking both explicitly is fine

testing/e2e/e2e_api_test.go
calbera: This specific test is only querying for 1 validator with index 0. It does not need a for loop.

testing/e2e/e2e_staking_test.go
calbera: this agreed is a duplicate check. Would remove one of them

node-api/backend/validator.go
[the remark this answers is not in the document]
nidhi-singh02: This is the error message when slot in request is higher `"failed to get state for slot 100: cannot query with height in the future; please provide a valid height: invalid height"`, feels fine to me. Pls suggest what error message you think might be better.

node-api/handlers/utils/constants.go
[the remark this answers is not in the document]
nidhi-singh02: My thought process/rationale was not to couple primitives package constant with node-api. Nw will remove from node-api.

testing/e2e/suite/types/beacon_client.go
nidhi-singh02: go-eth2-client is not compatible with our code. While calling the Validator function, I got error `failed to parse consensus version
no consensus version header and failed to parse response
unrecognised data version "deneb1"`. 
The validator endpoint calls `GetState` which returns "deneb1" as Version. I don't have exact context I see this got changed in this PR https://github.com/berachain/beacon-kit/pull/2456/files#diff-47bd111bd1dd529d5ab49d090583734214f3eebff67a47e6b06e71c74758d8a5

If we can return the standard forks and then remove the need to override the go-eth2-client function, would be great. CC @calbera

node-api/backend/validator.go
nidhi-singh02: My understanding is same as Cal's. @shotes Could you pls confirm.

node-api/handlers/beacon/types/response.go
nidhi-singh02: My bad, addressed.

node-api/handlers/beacon/header.go
nidhi-singh02: right, the TODO for Signature was already there, so kept it as in.

testing/e2e/e2e_staking_test.go
nidhi-singh02: done, removed one

testing/e2e/e2e_api_test.go
nidhi-singh02: In one, we are checking the response and in another we are checking Data inside the response.

testing/e2e/e2e_api_test.go
nidhi-singh02: @abi87 yes, If you see test `TestValidatorsEmptyIndices` we are verifying the validators that we got from the API response and what we passed i.e 5 and it matches.

Yeah, removed the loop in above case.

testing/e2e/suite/types/beacon_client.go
shotes: I think I'd be ok with returning the top level fork - i.e. the eth compatible ones only. So if we are on "deneb1" fork, the node-api endpoint returns "deneb"

consensus-types/types/validator.go
shotes: The original source is at https://github.com/prysmaticlabs/prysm/blob/0229a2055e6349655a471b2427f349e40c275cee/beacon-chain/rpc/eth/helpers/validator_status.go#L31

Also the restructuring makes the assumption that a validator that does not pass any of the checks defaults to `ValidatorStatusWithdrawalDone`, when it should instead be erroring because the validator status should not be possible.

testing/e2e/suite/types/beacon_client.go
shotes: I'll add that if we are returning data over the node-api that breaks compatibility with off the shelf ETH beacon API clients, then we should provide a compatible client (by forking).

So we should either modify the node-api to return compatible versions, or we should fork go-eth2-client such that it works with our versions. I think that may be work for a separate PR though

calbera: Am not opposed to returning the "top level" fork in the beacon API. i.e. {Deneb, Deneb1, Deneb2, etc all return "deneb"}. 

@nidhi-singh02 I think we should implement this IF it means no forking is required for these functions. 

IF it still means we must fork to handle our responses, I'm okay with returning and keeping as is for now. Most users must call the beacon API via raw HTTP clients right now (i.e. our proofs/ endpoint is using a custom golang client built around the endpoint)

node-api/backend/validator.go
nidhi-singh02: Thanks Shotes, made changes. I have renamed the variable name to `validatorIndices`.

consensus-types/types/validator.go
nidhi-singh02: Thanks Shotes. I have added a comment with the source link. I prefer keeping the function as is so it is identical to prsym.

node-api/backend/validator.go
nidhi-singh02: I have removed the logic where we are getting indexes before getting all the validators. Tried to handle it differently in `FilteredValidators` by first iterating over public key.

testing/e2e/suite/types/beacon_client.go
nidhi-singh02: So what happens if we query the validators only by state and have empty Indices and empty status, then the endpoint calls `GetState` which returns the data as `BeaconState`. I played around with the node api endpoint `GetState` and changed the version to "Deneb". Since our BeaconState is not identical to the Eth2 consensus spec we keep on getting errors like first I got `genesis_time is missing`, I added it and then got error for another field.

So for time being, I am leaving the Custom beacon client for validator endpoints, when we make BeaconState identical across in later PR, we can try to remove this and use the method for go-eth2-client directly.

nidhi-singh02:
> I think this PR needs some code cleanup and more testing above all. Also we may reconsider the implementation of `FilteredValidators` for efficiency

I had tested all the different combinations for these endpoints directly via swagger, tested it again with the current changes after the cleanup.

Addressed all review comments, only piece is pending which is UT for `FilteredValidators` where I need help.

consensus-types/types/validator.go
abi87: Agree that the `switch` implementation does assume some ordering of epochs while the `if`s implementation does not. Hence the latter is more robust. Let's keep it as is

node-api/backend/validator.go
```diff
@@ -21,80 +21,180 @@
...
+		if b.matchesPubkey(validator, id) || b.matchesIndex(index, id) {
+			return true
+		}
+	}
+	return false
+}
+
+func (b Backend) matchesPubkey(validator *types.Validator, id string) bool {
+	var pubkey crypto.BLSPubkey
+	if err := pubkey.UnmarshalText([]byte(id)); err != nil {
```
abi87, line 110: Also here I think we want to turn strings into PubKeys only once before passing them to `processValidator`. Otherwise we risk to Unmarshal over and over the same pubkeys

node-api/backend/validator.go
```diff
@@ -21,80 +21,180 @@
...
+	}
+	return validatorData, nil
+}
+
+func (b Backend) processValidator(
+	st *statedb.StateDB,
+	validator *types.Validator,
+	slot math.Slot,
+	ids []string,
+	statuses []string,
```
abi87, line 70: I may be wrong but one change I see wrt to previous implementation is that with this schema we allow to filter for both ids and statuses at the same time?
Not sure if this is a problem (I guess not), just highlighting a diff in the implementation.
cc @berachain/core

testing/e2e/e2e_api_test.go
abi87: You're right, I got confused here

abi87 (dismissed):
Much more readable and efficient code it seems to me!
I left some notes around:
- turning ids and status in their right types early (before `processValidator`) so to avoid marshalling them over and over
- made matches free functions wherever possible

Looking forward to merge this!

consensus-types/types/validator.go
```diff
@@ -317,3 +318,52 @@ func (v Validator) GetWithdrawableEpoch() math.Epoch {
...
+	}
+
+	// Status: withdrawal
+	if withdrawableEpoch <= currentEpoch {
+		if v.GetEffectiveBalance() != math.Gwei(0) {
+			return constants.ValidatorStatusWithdrawalPossible, nil
+		}
+		return constants.ValidatorStatusWithdrawalDone, nil
+	}
+	return "", errors.New("invalid validator status")
```
abi87, line 368: @berachain/core not sure I am happy with this choice here.
This case should only happen if the beacon state is messed up and not if above is possible (e.g. for a validator exitHeight < activationHeight).
However if this ever happens, this error would stop querying any validator, because of the error.
I know prysm does the same but should we instead log an error and let the other validators be returned?

node-api/backend/validator.go
```diff
@@ -21,80 +21,180 @@
...
+			return nil, errors.Wrapf(errInProcess, "failed to process validator")
+		}
+		if data != nil {
+			validatorData = append(validatorData, data)
+		}
+	}
+	return validatorData, nil
+}
+
+func (b Backend) processValidator(
```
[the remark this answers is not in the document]
nidhi-singh02: Good idea, calculated epoch and passed to `processValidator` which in turn made `processValidator` free.

node-api/backend/validator.go
```diff
@@ -21,80 +21,180 @@
...
+
+func (b Backend) matchesIndex(index math.U64, id string) bool {
+	return strconv.FormatUint(index.Unwrap(), 10) == id
+}
+
+func (b Backend) matchesStatusFilter(status string, statuses []string) bool {
+	return len(statuses) == 0 || slices.Contains(statuses, status)
+}
+
+func (b Backend) buildValidatorData(
```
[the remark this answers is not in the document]
nidhi-singh02: Addressed this and other places where functions can be made free.

node-api/backend/validator.go
```diff
@@ -21,80 +21,180 @@
...
+	var pubkey crypto.BLSPubkey
+	if err := pubkey.UnmarshalText([]byte(id)); err != nil {
+		return false
+	}
+	validatorPubkey := validator.GetPubkey()
+	return bytes.Equal(validatorPubkey[:], pubkey[:])
+}
+
+func (b Backend) matchesIndex(index math.U64, id string) bool {
+	return strconv.FormatUint(index.Unwrap(), 10) == id
```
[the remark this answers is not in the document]
nidhi-singh02: addressed, parsed ID as numeric and doing it before `processValidator`.

node-api/backend/validator.go
nidhi-singh02: addressed, parsed pubKey as crypto.BLSPubkey and doing it before processValidator.

node-api/backend/validator.go
nidhi-singh02: Yeah, this is as per the beacon API spec https://ethereum.github.io/beacon-APIs/#/Beacon/postStateValidators. Quoting `The lists of validator IDs and statuses to filter on`.

node-api/backend/validator.go
```diff
@@ -21,80 +21,180 @@
...
+		return nil, errors.Wrapf(err, "failed to get validators")
+	}
+
+	validatorData := make([]*beacontypes.ValidatorData, 0, len(validators))
+	for _, validator := range validators {
+		data, errInProcess := b.processValidator(st, validator, slot, ids, statuses)
+		if errInProcess != nil {
+			return nil, errors.Wrapf(errInProcess, "failed to process validator")
+		}
+		if data != nil {
```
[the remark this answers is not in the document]
nidhi-singh02: Personally for me above case is not so readable as compared to the if conditions. I am okay with whatever the majority says.
Since this pattern is spread across the code, it would be better to handle it in a separate PR.

consensus-types/types/validator.go
calbera: I actually thinks returning error here is good. If we see this error returned from the API, it’s a strong sign that the beacon state registry of validators needs a fixing

node-api/backend/validator.go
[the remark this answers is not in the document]
calbera: Would agree with explicitly defining an error and handling that when the data is empty. It generally should be OK to assume the response is not nil when err is nil

node-api/backend/validator.go
calbera: Lgtm

node-api/backend/validator.go
```diff
@@ -21,80 +21,224 @@
...
+	epoch math.Epoch,
+	statuses []string,
+) (*beacontypes.ValidatorData, error) {
+	status, err := validator.Status(epoch)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get validator status")
+	}
+
+	if !matchesStatusFilter(status, statuses) {
+		//nolint:nilnil // no data to return
```
calbera, line 171: Would agree that rather than returning nil, nil, we should return a defined error like ErrNoValidatorData and then explicitly catch this error case in the calling function.

Will remove the nolinting of nil,nil as well

consensus-types/types/validator.go
abi87: I agree, but we would not see which validator is broken right? The API will return an error and no clear indications of where is the error.
Maybe we should at least decorate the final error with some info about the faulty validators

node-api/backend/validator.go
```diff
@@ -21,80 +21,224 @@
...
+		f.numericIDs = append(f.numericIDs, index)
+		return
+	}
+
+	// Try parsing as pubkey
+	var pubkey crypto.BLSPubkey
+	if err := pubkey.UnmarshalText([]byte(id)); err == nil {
+		f.pubkeys = append(f.pubkeys, pubkey)
+	}
+	// Silently skip invalid IDs
```
abi87, line 68: q: is this correct? Down the line we threat an empty filter as `show all validators`. Should we treat a list of malformed IDs into a catchall?
Consider erring here.
cc @calbera

node-api/backend/validator.go
```diff
@@ -21,80 +21,224 @@
...
+	validators, err := st.GetValidators()
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get validators")
+	}
+
+	// Parse all IDs and pubkeys once at the start
+	filters := parseValidatorIDs(ids)
+	epoch := b.cs.SlotToEpoch(slot)
+
+	return filterAndBuildValidatorData(st, validators, filters, epoch, statuses)
```
abi87, line 90: q: why we do sanitize ids but not statuses? Shouldn't we check that statuses are among the only ones listed?

consensus-types/types/validator.go
calbera: Yeah can patch this error message with validator index and/or pubkey

node-api/backend/validator.go
calbera: Am indifferent here, really depends on what the spec enforces

node-api/backend/validator.go
calbera: I believe the echo engine handles this

abi87: Nice

consensus-types/types/validator.go
```diff
@@ -365,5 +365,5 @@ func (v *Validator) Status(currentEpoch math.Epoch) (string, error) {
 		}
 		return constants.ValidatorStatusWithdrawalDone, nil
 	}
-	return "", errors.New("invalid validator status")
+	return "", errors.Wrapf(errors.New("invalid validator status"), "validator pubkey: %s, index: %d,", v.Pubkey[:], index)
```
abi87, line 368: you don't need to do this here. You can return the error and wrap it in the called

node-api/backend/validator.go
nidhi-singh02: makes sense, returned an error and then handled it in `filterAndBuildValidatorData` so that when no status matches, error is not returned (as per spec).

consensus-types/types/validator.go
calbera, line 368:
```suggestion

	return "", errors.New("invalid validator status")
```

node-api/backend/validator.go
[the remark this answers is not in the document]
calbera: so @nidhi-singh02, last thing is I think we can now remove the `if data != nil` check.

node-api/backend/validator.go
```diff
@@ -21,80 +21,230 @@
...
+		return nil, errors.Wrapf(err, "failed to get validator status for validator pubkey %s and index %d", validator.GetPubkey(), index)
+	}
+
+	if !matchesStatusFilter(status, statuses) {
+		return nil, ErrStatusFilterMismatch
+	}
+
+	balance, err := st.GetBalance(index)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get validator balance")
```
calbera, line 183: nit: add the val index/pubkey here

node-api/backend/validator.go
```diff
@@ -21,80 +21,230 @@
...
 		if err != nil {
-			return nil, err
+			return nil, errors.Wrapf(err, "failed to get validator index by id %s", id)
 		}
 		var balance math.U64
 		// TODO: same issue as above, shouldn't error on not found.
 		balance, err = st.GetBalance(index)
 		if err != nil {
-			return nil, err
+			return nil, errors.Wrapf(err, "failed to get validator balance")
```
calbera, line 247: nit: add the val index/pubkey here

calbera (approved):
couple nits then LGTM

testing/e2e/e2e_staking_test.go
```diff
@@ -115,14 +115,15 @@ func (s *BeaconKitE2ESuite) TestDepositRobustness() {
 
 	// Grab genesis validators to get withdrawal creds.
 	s.Require().NoError(apiClient.Connect(s.Ctx()))
-	s.Require().NotNil(apiClient.BeaconKitNodeClient)
-	response, err := apiClient.BeaconKitNodeClient.Validators(s.Ctx(), &api.ValidatorsOpts{
+	response, err := apiClient.Validators(s.Ctx(), &api.ValidatorsOpts{
 		State:   "genesis",
 		Indices: []phase0.ValidatorIndex{0, 1, 2, 3, 4},
```
shotes, line 120: Can this be changed to empty `Indices` so that we know we are querying for all validators, instead of the hardcoded indices.

shotes (commented):
nits and questions. Really liking how this is turning out and all of the additional test cases. Also like the change of the `NewResponse(data any) GenericResponse` return value in the node-api. Fits well.

node-api/backend/validator.go
nidhi-singh02: done

node-api/backend/validator.go
nidhi-singh02: done, added index at one more place.

testing/e2e/e2e_staking_test.go
nidhi-singh02: Wanted to keep a test for querying multiple specific indices. We already have a test with empty Indices `TestValidatorsEmptyIndices` in that case we get all the validators.

testing/e2e/e2e_api_test.go
```diff
@@ -48,3 +57,184 @@ func (s *BeaconKitE2ESuite) TestBeaconAPIStartup() {
...
+// TestValidatorsEmptyStatuses tests that querying validators with empty statuses returns all validators.
+func (s *BeaconKitE2ESuite) TestValidatorsEmptyStatuses() {
+	client := s.initBeaconTest()
+
+	// Query validators with empty statuses
+	validatorsResp, err := client.Validators(
+		s.Ctx(),
+		&beaconapi.ValidatorsOpts{
+			State:           utils.StateIDHead,
+			ValidatorStates: []apiv1.ValidatorState{}, // empty statuses
```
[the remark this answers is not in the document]
nidhi-singh02: In terms of results, both `TestValidatorsEmptyStatuses` and `TestValidatorsEmptyIndices` returns all validators, there's no difference as such. There's a comment before the function specifying same.

As it feels redundant, I have merged both the tests into one `TestValidatorsEmptyIndicesAndStatuses` and added comments for clarity.

node-api/backend/validator.go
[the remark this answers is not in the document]
nidhi-singh02: done [here](https://github.com/berachain/beacon-kit/pull/2467/commits/eb07e3f488ff80ecdcead91cccfd0f70f54a99b2)

consensus-types/types/validator.go
nidhi-singh02: Introduced typed error [here](https://github.com/berachain/beacon-kit/pull/2467/commits/eb07e3f488ff80ecdcead91cccfd0f70f54a99b2)

abi87 (approved):

PR #2468, opened 2025-02-03
refactor(init-cli): Refactor the Init CLI command to be callable from runtime

rezbera:


cli/commands/initialize/initialize.go
```diff
@@ -125,7 +124,7 @@ func InitCmd(mm interface {
 			if config.RootDir == "" {
 				config.RootDir = clientCtx.HomeDir
 			}
-
+			config.Moniker = args[0]
```
rezbera, line 127: Reordered setting the moniker

cli/commands/initialize/initialize.go
```diff
@@ -92,10 +92,9 @@ func displayInfo(dst io.Writer, info printInfo) error {
 	return err
 }
 
-//nolint:funlen,gocognit,mnd // based on cosmossdk implementation
+// InitCmd is based on cosmossdk implementation.
 func InitCmd(mm interface {
 	DefaultGenesis() map[string]json.RawMessage
-	ValidateGenesis(genesisData map[string]json.RawMessage) error
```
rezbera, line 98: We run ValidateGenesis in `initChain` and hence we don't use it here (i.e. this was unused)

PR #2471, opened 2025-02-04
chore: improve test coverage stability with atomic mode

mdqst:
noticed that running tests with `-race` could lead to issues in concurrent tests.
adding `-covermode=atomic` ensures more reliable coverage results.
this should prevent flaky behavior when measuring coverage in race conditions.


calbera:
Thanks! Closing in favor of https://github.com/berachain/beacon-kit/pull/2521

PR #2478, opened 2025-02-04
fix(cli): unlocking rollback

abi87:
Rollback currently crash cause of "todo, will implement later".  This PR fixes this


cli/commands/server/rollback.go
```diff
@@ -71,16 +71,16 @@ application.
...
 			}
-			// rollback the multistore
 
+			// rollback the multistore
 			if err = app.CommitMultiStore().RollbackToVersion(height); err != nil {
 				return fmt.Errorf("failed to rollback to version: %w", err)
 			}
 
 			logger.Info(
-				"Rolled back state to height %d and hash %X\n",
```
abi87, line 81: this will print gibberish

node-core/node/node.go
```diff
@@ -41,17 +42,18 @@ type node struct {
 	logger log.Logger
 	// registry is the node's service registry.
 	registry *service.Registry
-
-	// TODO: FIX, HACK TO MAKE CLI HAPPY FOR NOW.
-	// THIS SHOULD BE REMOVED EVENTUALLY.
-	types.Node
```
abi87, line 47: rollback command currently fails because this nil interface is called when command try and load CommitMultistore.
Instead we should use the commit multistore provided by consensus service

consensus/cometbft/service/abci.go
```diff
@@ -55,7 +55,7 @@ func (s *Service) PrepareProposal(
 func (s *Service) Info(context.Context,
 	*cmtabci.InfoRequest,
 ) (*cmtabci.InfoResponse, error) {
-	lastCommitID := s.sm.CommitMultiStore().LastCommitID()
+	lastCommitID := s.sm.GetCommitMultiStore().LastCommitID()
```
abi87, line 58: I introduced the `GetCommitMultiStore` to reduce the structs implementing `CommitMultistoreAccessor`

node-core/services/registry/registry.go
```diff
@@ -66,6 +74,31 @@ func NewRegistry(logger log.Logger, opts ...RegistryOption) *Registry {
...
+	}
+
+	sh, ok := svc.(CommitMultistoreAccessor)
+	if !ok {
+		logger.Warn("Rollback won't work",
+			"err",
+			fmt.Sprintf("service %s does not implement CommitMultistoreAccessor",
+				cservice.AppName,
+			),
+		)
```
abi87, line 98: same here

node-core/services/registry/registry.go
```diff
@@ -38,8 +41,10 @@ type Basic interface {
 	Name() string
 }
 
-type Dispatcher interface {
-	Start(ctx context.Context) error
```
abi87, line 42: This is the usual code duplication, probably due to code copying

node-core/services/registry/registry.go
```diff
@@ -38,8 +41,10 @@ type Basic interface {
...
+// CommitMultistoreAccessor allows access to the commit multistore
+// This is required by commands like Rollback.
+type CommitMultistoreAccessor interface {
```
abi87, line 46: This is a new interface that registry must support to allow Rollback

node-core/services/registry/registry.go
```diff
@@ -151,3 +156,22 @@ func (s *Registry) FetchService(service interface{}) error {
...
+	}
+
+	sh, ok := svc.(CommitMultistoreAccessor)
+	if !ok {
+		err := fmt.Errorf("service %s does not implement CommitMultistoreAccessor",
+			cservice.AppName,
+		)
+		panic(err)
+	}
+	return sh.CommitMultiStore()
```
abi87, line 176: we could do this in NewRegistry, but I believe this would cause tests like `TestRegistry_FetchService` and all the testing built on top of it more complex

consensus/cometbft/service/finalize_block.go
```diff
@@ -99,15 +99,15 @@ func (s *Service) finalizeBlockInternal(
 
 // workingHash gets the apphash that will be finalized in commit.
 // These writes will be persisted to the root multi-store
-// (s.sm.CommitMultiStore()) and flushed
-// to disk in the Commit phase. This means when the ABCI client requests
-// Commit(), the application state transitions will be flushed to disk and as a
-// result, but we already have
-// an application Merkle root.
+// (s.smGet.CommitMultiStore()) and flushed to disk  in the
```
rezbera, line 102: `(s.smGet.CommitMultiStore())`

nit: typo

consensus/cometbft/service/service.go
```diff
@@ -222,12 +222,13 @@ func (s *Service) Stop() error {
 
 // Name returns the name of the cometbft.
 func (s *Service) Name() string {
-	return appName
+	return AppName
```
rezbera, line 225: Feels odd to have a getter for this constant. Might be refactor for a separate PR if its too big a change.

rezbera: Nevermind - looks like it's necessary for SeviceRegistry

consensus/cometbft/service/state/state.go
```diff
@@ -87,8 +87,7 @@ func (sm *Manager) Close() error {
 	return sm.db.Close()
 }
 
-// CommitMultiStore returns the CommitMultiStore of the Manager.
-// TODO:REMOVE.
```
rezbera, line 91: Any idea on the logic for `TODO:Remove`?

node-core/services/registry/registry.go
```diff
@@ -151,3 +156,22 @@ func (s *Registry) FetchService(service interface{}) error {
...
+		err := fmt.Errorf("service %s not found", cservice.AppName)
+		panic(err)
+	}
+
+	sh, ok := svc.(CommitMultistoreAccessor)
+	if !ok {
+		err := fmt.Errorf("service %s does not implement CommitMultistoreAccessor",
+			cservice.AppName,
+		)
+		panic(err)
```
rezbera, line 174: Can we use `func (s *Registry) FetchService(service interface{}) error {` instead? Then you should be able to query it like [this](https://github.com/berachain/beacon-kit/pull/2411/files#diff-3e5193e02d69281deb8c8e0fc02e088d7ab21642d6083e61d71ac517bd3cefffR344)

rezbera: I'll sanity check above first to see if it works

node-core/services/registry/registry.go
rezbera, line 176: this works
```suggestion
	var cometService *cometbft.Service
	err := s.FetchService(&cometService)
	if err != nil {
		err = fmt.Errorf("failed to fetch cometbft service: %w", err)
		panic(err)
	}
	return cometService.CommitMultiStore()
```

node-core/node/node.go
```diff
@@ -41,17 +42,18 @@ type node struct {
...
 }
 
 // New returns a new node.
 func New[NodeT types.Node](
 	registry *service.Registry, logger log.Logger) NodeT {
+	node := &node{
```
rezbera, line 50: shadowing type declaration https://github.com/berachain/beacon-kit/blob/bd08c21db90ae9447625ad3e7f2e354a19c87cf0/node-core/node/node.go#L40 - consider renaming `node`

rezbera (commented):
Reviewed - consider the alternative impl for `registry.CommitMultiStore` and nits.

Otherwise lgtm

node-core/services/registry/registry.go
abi87: Nice!

rezbera (approved):
LGTM

node-core/node/node.go
abi87: Done

node-core/services/registry/registry.go
abi87: Done

node-core/services/registry/registry.go
```diff
@@ -151,3 +156,13 @@ func (s *Registry) FetchService(service interface{}) error {
 	}
 	return errUnknownService
 }
+
+func (s *Registry) CommitMultiStore() store.CommitMultiStore {
+	var cometService *cometbft.Service
```
calbera, line 161:
```suggestion
	var cometService cometbft.Service
```
try initializing cometService as the empty struct

rezbera: Not sure about empty struct but has to be double pointer iirc for fetch

calbera (approved):
tACK. LGTM

node-core/services/registry/registry.go
calbera: not sure its needed, works without issue for me

Issue #2408, opened 2025-01-24
beacond unable to connect to ethereumjs


nidhi-singh02:
`make start-ethereumjs` not working as expected : beacond unable to connect to ethereumjs.

CL keeps on showing logs sayings `Waiting for execution client to start.`

abi87:
Outdated due to Pectra1 fork, introducing bera-reth and bera-geth as only evm clients

PR #2479, opened 2025-02-05
Ask to delete EL data dir on make start-[client]

fridrik01:
I have now several times run `start-geth` without realizing it always deletes the EL data directory causing me to need to reset my local node. This PR addresses this so when any EL starts it checks if the eth data dir exists and asks if it should delete it or not (similar to how `make start` works).

Furthermore, this PR also addresses the case where on linux (Ubuntu under WSL) using Docker Desktop the data directories for `geth` and `ethereumjs` will be created with root privileges requiring manually deleting them with `sudo`. Now, we first try to delete with `rm` and if that fails we try with `sudo`.

NOTE: Not deleting the eth data directory each time uncovers an issue in `reth` where if we ever restart it, then beacond will error with `error=payload status is SYNCING` and ` err=no beacon block in abci request`.


scripts/build/testing.mk
```diff
@@ -183,14 +203,15 @@ start-nethermind: ## start an ephemeral `nethermind` node
 	--Init.ChainSpecPath ../$(NETHER_ETH_GENESIS_PATH)
 
 start-besu: ## start an ephemeral `besu` node
+	$(call ask_reset_dir_func, .tmp/besu)
 	docker run \
 	-p 30303:30303 \
 	-p 8545:8545 \
 	-p 8551:8551 \
-	-v $(PWD)/${HOMEDIR}:/${HOMEDIR} \
+	-v $(PWD)/.tmp:/.tmp \
```
fridrik01, line 211: besu was not properly mounting the correct directory, now stopping it and restarting works as expected

scripts/build/testing.mk
```diff
@@ -234,7 +255,7 @@ start-erigon: ## start an ephemeral `erigon` node
 	--datadir /.tmp/erigon
 
 start-ethereumjs:
-	rm -rf .tmp/ethereumjs
+	$(call ask_reset_dir_func, .tmp/ethereumjs)
```
fridrik01, line 258: Note that I was not able to get ethereumjs to work so that beacond was producing blocks

scripts/build/testing.mk
```diff
@@ -183,14 +203,15 @@ start-nethermind: ## start an ephemeral `nethermind` node
...
 	docker run \
 	-p 30303:30303 \
 	-p 8545:8545 \
 	-p 8551:8551 \
-	-v $(PWD)/${HOMEDIR}:/${HOMEDIR} \
+	-v $(PWD)/.tmp:/.tmp \
 	-v $(PWD)/${TESTAPP_FILES_DIR}:/${TESTAPP_FILES_DIR} \
 	hyperledger/besu:latest \
-	--data-path=.tmp/besu \
+	--data-path=/.tmp/besu \
```
fridrik01, line 214: Let me know if you prefer I create constants for this (and erigon/ethereumjs)

scripts/build/testing.mk
calbera: are you able to get it working on main?

rezbera: open issue https://github.com/berachain/beacon-kit/issues/2408

scripts/build/testing.mk
```diff
@@ -23,6 +23,26 @@ HTTP_URL = localhost:8551
 IPC_PREFIX = ipc://
 HTTP_PREFIX = http://
 
+# ask_reset_dir_func checks if the directory passed in exists, and if so asks the user whether it
+# should delete it. Note that on linux, docker may have created the directory with root
+# permissions, so we may need to ask the user to delete it with sudo
+define ask_reset_dir_func
+	@abs_path=$(abspath $(1)); \
```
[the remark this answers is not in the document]
fridrik01: yeah bash is bad enough, but doing it inside make .. dear lord

scripts/build/testing.mk
calbera: would agree to removing it in another PR

calbera (approved):
I like this change, I effectively do this by myself and having it automated is nice. utACK

nidhi-singh02 (approved):
lgtm! I have seen different data dir while testing different EL combinations. This is helpful.

rezbera (approved):
Tested and it works. Noting that when I tested with Geth, a CL restart was also required to recover.

PR #2480, opened 2025-02-05
Add Perimeter fuzzing suite

ljz3:


calbera:
Thanks @ljz3! We should add this to our contracts repo https://github.com/berachain/contracts instead!

PR #2482, opened 2025-02-06
chore: enhance deposit command messages

### LINKED ISSUE #2477: Improve help messages
1. The default help for beacond has this line as the first line of output:
"A basic beacon node, usable most standard networks."
maybe: "A beacon node, usable with most Ethereum execution clients"

2. `help deposit create-validator` says: Creates a validator deposit with the necessary credentials.  This suggests it transacts for you. The help further refers to a broadcast flag, which is not otherwise mentioned. Revise to clarify what this command now does, which is generate transaction parameters.

3. `deposit validate` should produce some kind of message to indicate that the parameters check out. Right now the only hint everything's fine is if the exit status $? is 0

4.  `deposit create-validator` returns the amount as a hex string.   If you then provide this hex string to `deposit validate` it is rejected. Either return decimal (preferred) or accept hex (misleading, because this is generally understood to represent bytes and not a hexademical encoding of a decimal number).

wonderfan:
use this change set to fix issue #2477


cli/commands/deposit/validate.go
```diff
@@ -101,7 +102,13 @@ func validateDepositMessage(chainSpec chain.Spec) func(
 			return err
 		}
 
-		return ValidateDeposit(chainSpec, pubkey, credentials, amount, genesisValidatorRoot, signature)
+		err = ValidateDeposit(chainSpec, pubkey, credentials, amount, genesisValidatorRoot, signature)
+		if err != nil {
+			return err
+		}
+
+		fmt.Println("Deposit message validated successfully")
```
calbera, line 110: can you use the cobra command's Print function?

wonderfan: Done

wonderfan:
@calbera Could you please take a look again?

calbera:
Hey thanks for the PR, but we'll be resolving this issue (among other fixes) in https://github.com/berachain/beacon-kit/pull/2565

PR #2483, opened 2025-02-06
chore(node-api): Validator balance endpoints and getStateValidator

nidhi-singh02:
Break-down of tasks

- [x] POST Validator balance breaking
- [x] Validator balance endpoint returns error when no record found
- [x] POST/GET Validator balance with public key failing
- [x] Address TODO in validator balance
- [x] https://ethereum.github.io/beacon-APIs/#/Beacon/getStateValidator fails
- [x] If validator not found in getStateValidator return response in certain format
- [x] Write E2E tests for endpoints that exist in go-eth2 client
- [x] Write E2E tests for endpoints not in go-eth2 client by calling the API endpoint directly

Endpoints - 
https://ethereum.github.io/beacon-APIs/#/Beacon/getStateValidator
https://ethereum.github.io/beacon-APIs/#/Beacon/getStateValidatorBalances
https://ethereum.github.io/beacon-APIs/#/Beacon/postStateValidatorBalances


consensus-types/types/validator.go
```diff
@@ -317,3 +318,50 @@ func (v Validator) GetWithdrawableEpoch() math.Epoch {
 func (v Validator) GetWithdrawalCredentials() WithdrawalCredentials {
 	return v.WithdrawalCredentials
 }
+
+// Status returns the current validator status based on its set epoch values.
+func (v *Validator) Status(currentEpoch math.Epoch) (string, error) {
+	activationEpoch := v.GetActivationEpoch()
```
abi87, line 324: Isn't the following equivalent and easier to understand:

```
	switch {
	case currentEpoch < v.GetActivationEligibilityEpoch():
		return constants.ValidatorStatusPendingInitialized, nil
	case currentEpoch < v.GetActivationEpoch():
		return constants.ValidatorStatusPendingQueued, nil
	case currentEpoch < v.GetExitEpoch():
		if v.IsSlashed() {
			return constants.ValidatorStatusActiveSlashed, nil
		}
		return constants.ValidatorStatusActiveOngoing, nil
	case currentEpoch < v.GetExitEpoch():
		if v.IsSlashed() {
			return constants.ValidatorStatusExitedSlashed, nil
		}
		return constants.ValidatorStatusExitedUnslashed, nil
	default:
		if v.GetEffectiveBalance() != math.Gwei(0) {
			return constants.ValidatorStatusWithdrawalPossible, nil
		}
		return constants.ValidatorStatusWithdrawalDone, nil
	}
```

abi87: Note that BeaconKit must guarantee that `ActivationElegibilityEpoch < ActivationEpoch <= ExitEpoch < WithdrawalEpoch`

abi87: Also we don't currently have slashing so we could even consider returning error if a validator is ever slashed and further simplifying the switch

nidhi-singh02: Good suggestion, will change to switch case. Above method `Status()` is being called only in case of node api validator endpoints. This is essentially to fetch the status of a validator.  I believe if we don't have slashing, the status would never be set to slashing so that piece of code won't really be executed in any case. Should we remove the slashing condition altogether ?

nidhi-singh02: These comments are part of the open PR - https://github.com/berachain/beacon-kit/pull/2467#discussion_r1949372351 , let's discuss there.  I have kept the switch as is I agree with Cal's and Shote's as it might be useful eventually and wanted to have the same logic as per prysm. Let's continue the convo in one place, marking this as resolved.

node-api/handlers/error.go
```diff
@@ -0,0 +1,45 @@
...
+// TO THE EXTENT PERMITTED BY APPLICABLE LAW, THE LICENSED WORK IS PROVIDED ON
+// AN “AS IS” BASIS. LICENSOR HEREBY DISCLAIMS ALL WARRANTIES AND CONDITIONS,
+// EXPRESS OR IMPLIED, INCLUDING (WITHOUT LIMITATION) WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, NON-INFRINGEMENT, AND
+// TITLE.
+
+package handlers
+
+// HTTPError represents an HTTP error response.
+type HTTPError struct {
```
abi87, line 24: we may have this already in our codebase? cc @rezbera

node-api/handlers/beacon/types/request.go
```diff
@@ -67,7 +67,7 @@ type GetValidatorBalancesRequest struct {
 
 type PostValidatorBalancesRequest struct {
 	types.StateIDRequest
-	IDs []string `validate:"dive,validator_id"`
+	IDs []string `json:"-" validate:"dive,validator_id"`
```
abi87, line 70: what is `dive`?

nidhi-singh02: dive is a validation tag in go-playground/validator package- it is used to dive in array/map and validate each element inside it according to the validation rules.
In this case, go into slice `IDs` and apply `validator_id` validation rule at each string in the slice. Refer [dive](https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Dive)

node-api/backend/validator.go
```diff
@@ -200,11 +204,21 @@ func (b Backend) ValidatorByID(
...
+	default:
 		return nil, errors.Wrapf(err, "failed to get validator index by id %s", id)
 	}
 	validator, err := st.ValidatorByIndex(index)
-	if err != nil {
+	switch {
+	case err == nil:
+		// continue processing
+	case errors.Is(err, collections.ErrNotFound):
+		return nil, ErrValidatorNotFound
```
abi87, line 220: I think you don't need to special case here?
Index may not exist but if it exists the validator must exist.
So here you can just
```
if err != nil {
return err
}
```

node-api/backend/validator.go
```diff
@@ -228,21 +242,49 @@ func (b Backend) ValidatorByID(
 func (b Backend) ValidatorBalancesByIDs(
 	slot math.Slot, ids []string,
 ) ([]*beacontypes.ValidatorBalanceData, error) {
-	var index math.U64
 	st, _, err := b.stateFromSlot(slot)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to get state from slot %d", slot)
 	}
+
+	// If no IDs provided, return all validator balances
```
abi87, line 250: what are the specs here?
We return every validator balance or only active ones?
Note that a validator entry in create in beaconState the moment a deposit is successfully processed, even if it has 10K and it cannot be activated yet.

node-api/handlers/beacon/validators.go
```diff
@@ -82,9 +88,15 @@ func (h *Handler) GetStateValidator(c handlers.Context) (any, error) {
...
+		if errors.Is(err, backend.ErrValidatorNotFound) {
+			return &handlers.HTTPError{
+				Code:    http.StatusNotFound,
+				Message: "Validator not found",
+			}, nil
+		}
 		return nil, err
 	}
-	return validator, nil
+	return beacontypes.NewResponse(validator), nil
```
abi87, line 99: nit: switch?

node-api/backend/validator.go
nidhi-singh02: `Balances will be returned for all indices or public key that match known validators.If an index or public key does not match any known validator, no balance will be returned but this will not cause an error.` from [reference](https://ethereum.github.io/beacon-APIs/#/Beacon/getStateValidatorBalances).
Here status is not a filter, so all validator balance matching the filter.

node-api/handlers/beacon/validators.go
nidhi-singh02: done

node-api/backend/validator.go
nidhi-singh02: We want to show the specific message `Validator not found`, so done like this. Another way is like you said, not returning the defined error of `ErrValidatorNotFound` and directly returning err and in caller `GetStateValidator ` using `collections.ErrNotFound` in switch case.
Wanted to use our defined error in handler hence doing like this. Let me know your thoughts.

abi87: Oh, I see, I misunderstood your intent. All good

node-api/handlers/beacon/validators.go
```diff
@@ -95,7 +108,15 @@ func (h *Handler) GetStateValidatorBalances(c handlers.Context) (any, error) {
 		return nil, err
 	}
 	slot, err := utils.SlotFromStateID(req.StateID, h.backend)
+
+	// Check for error message "slot not found at state root"
 	if err != nil {
+		if strings.Contains(err.Error(), "slot not found at state root") {
```
calbera, line 114: I know its not strictly in the scope of this PR but can we also make this error a strongly typed `ErrNoSlotForStateRoot` or so? And then this check can be done with `errors.Is(err, ErrNoSlotForStateRoot)`

node-api/handlers/error.go
calbera: this looks fine to me, but wondering if we should make this consistent across all beacon API endpoints

testing/e2e/e2e_api_test.go
```diff
@@ -21,14 +21,76 @@
...
+// Get overrides http.Client.Get to use baseURL
+func (c *BeaconHTTPClient) Get(path string) (*http.Response, error) {
+	return c.Client.Get(c.baseURL + path)
+}
+
+// pubkey is the public key of the validator.
+// Made it global so that I can use it in another test.
+var pubkey string
+
+const localHost = "localhost"
```
calbera, line 69: nit: move to top of file

testing/e2e/e2e_api_test.go
```diff
@@ -21,14 +21,76 @@
...
+}
+
+// Get overrides http.Client.Get to use baseURL
+func (c *BeaconHTTPClient) Get(path string) (*http.Response, error) {
+	return c.Client.Get(c.baseURL + path)
+}
+
+// pubkey is the public key of the validator.
+// Made it global so that I can use it in another test.
+var pubkey string
```
calbera, line 67: I don't understand why this needs to be reused in test. Would prefer not having global vars.

calbera: Also nit: for code comments, let's try not to use "I"; "we" is sometimes okay but code comments generally shouldn't be written in the [first-person](https://www.grammarly.com/blog/grammar/first-second-and-third-person/)

calbera (commented):
some minor nits, but makes sense overall

node-api/handlers/beacon/validators.go
```diff
@@ -109,12 +130,20 @@ func (h *Handler) GetStateValidatorBalances(c handlers.Context) (any, error) {
 }
 
 func (h *Handler) PostStateValidatorBalances(c handlers.Context) (any, error) {
-	req, err := utils.BindAndValidate[beacontypes.PostValidatorBalancesRequest](
-		c, h.Logger(),
-	)
-	if err != nil {
-		return nil, err
+	var ids []string
+	if err := c.Bind(&ids); err != nil {
```
calbera, line 134: q: why can't we do BindAndValidate here anymore? I may be missing smth

testing/e2e/e2e_api_test.go
nidhi-singh02: The idea was to query the validator by the public key. Removed the global vars, re-written the test in another way.

Thanks, never do that. This auto comment :D

node-api/handlers/beacon/validators.go
nidhi-singh02: addressed.

node-api/handlers/beacon/validators.go
nidhi-singh02: No, we can't use BindAndValidate - the API was breaking earlier due to this. `PostValidatorBalancesRequest` has IDs []string with a json:"-" tag, which means it won't be bound from JSON. This is why you're manually binding the IDs from the request body.

node-api/handlers/beacon/types/request.go
```diff
@@ -57,7 +57,7 @@ type PostStateValidatorsRequest struct {
 
 type GetStateValidatorRequest struct {
 	types.StateIDRequest
-	ValidatorID string `query:"validator_id" validate:"required,validator_id"`
+	ValidatorID string `param:"validator_id" validate:"required,validator_id"`
```
abi87, line 60: @nidhi-singh02, question: why we replaced query instead of adding params?

testing/e2e/e2e_api_test.go
```diff
@@ -21,14 +21,72 @@
...
+type StateValidatorResponse struct {
+	Data struct {
+		Index     string `json:"index"`
+		Balance   string `json:"balance"`
+		Status    string `json:"status"`
+		Validator struct {
+			Pubkey string `json:"pubkey"`
+		} `json:"validator"`
+	} `json:"data"`
+}
```
abi87, line 53: @nidhi-singh02, question: why do we have to redefine it here?

testing/e2e/e2e_api_test.go
```diff
@@ -21,14 +21,72 @@
...
+
+// Get overrides http.Client.Get to use baseURL
+func (c *BeaconHTTPClient) Get(path string) (*http.Response, error) {
+	return c.Client.Get(c.baseURL + path)
+}
+
+// initHTTPBeaconTest initializes the http client for the beacon node api.
+// It gets the public ports from the consensus client and creates a http client with the baseURL.
+// This is needed where we want to test the node-api directly as
+// some of the methods are not present in go-eth2-client library.
```
abi87, line 70: why we are adding extra methods, outside of `go-eth2-client`?

node-api/handlers/beacon/types/request.go
abi87: This is to adhere to API definition

testing/e2e/e2e_api_test.go
abi87: Limitations of the `go-eth2-client` client

node-api/handlers/utils/id.go
```diff
@@ -49,7 +51,11 @@ func SlotFromStateID[StorageBackendT interface {
 	if err != nil {
 		return 0, err
 	}
-	return storage.GetSlotByStateRoot(root)
+	slot, err := storage.GetSlotByStateRoot(root)
+	if err != nil {
+		return 0, ErrNoSlotForStateRoot
```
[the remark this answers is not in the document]
nidhi-singh02: Yes, will clean up in a separate PR.

testing/e2e/e2e_api_test.go
nidhi-singh02: Addressed, removed it. Using `ValidatorData` from node-api package.

calbera:
TODO: fix usage of `validator_id` param
```go
	Method:  http.MethodGet,
	Path:    "/eth/v1/beacon/states/:state_id/validators/:validator_id",
	Handler: h.GetStateValidator,
```

abi87:
> TODO: fix usage of `validator_id` param
> 
> ```go
> 	Method:  http.MethodGet,
> 	Path:    "/eth/v1/beacon/states/:state_id/validators/:validator_id",
> 	Handler: h.GetStateValidator,
> ```

fixed in https://github.com/berachain/beacon-kit/pull/2539

node-api/handlers/beacon/validators.go
shotes: I wonder if there is a way to make BindAndValidate accept the unnamed json list from the request. If not, then this makes sense as it is.

testing/e2e/e2e_api_test.go
```diff
@@ -205,3 +253,271 @@ func (s *BeaconKitE2ESuite) TestValidatorsWithSpecificStatus() {
...
+	s.Require().NoError(err)
+	s.Require().NotNil(balancesResp)
+	s.Require().NotEmpty(balancesResp.Data)
+
+	// Verify balance data
+	for _, balance := range balancesResp.Data {
+		s.Require().NotNil(balance)
+		s.Require().True(balance > 0, "Validator balance should be positive")
+		// 4e12 Gwei = 4 * 10^12 Gwei = 4,000,000,000,000 Gwei = 4 million BERA
+		s.Require().True(balance <= 4e12, "Validator balance should not exceed 4 million BERA")
```
calbera, line 382: nit: 4e12 Gwei would be 4e3, or 4000, BERA

nidhi-singh02:
> #2539



> > TODO: fix usage of `validator_id` param
> > ```go
> > 	Method:  http.MethodGet,
> > 	Path:    "/eth/v1/beacon/states/:state_id/validators/:validator_id",
> > 	Handler: h.GetStateValidator,
> > ```
> 
> fixed in #2539

I remember discussing about `param` change here https://github.com/berachain/beacon-kit/pull/2483#discussion_r1969303285. This [particular API](https://ethereum.github.io/beacon-APIs/#/Beacon/getStateValidator) (as shown in the PR description) is part of this PR along with the e2e tests for this, I think there was some confusion.

> ValidatorID string `param:"validator_id" validate:"required,validator_id"

nidhi-singh02:
> > TODO: fix usage of `validator_id` param
> > ```go
> > 	Method:  http.MethodGet,
> > 	Path:    "/eth/v1/beacon/states/:state_id/validators/:validator_id",
> > 	Handler: h.GetStateValidator,
> > ```
> 
> fixed in #2539

I remember discussing about param change here https://github.com/berachain/beacon-kit/pull/2483#discussion_r1969303285. This [particular API](https://ethereum.github.io/beacon-APIs/#/Beacon/getStateValidator) (as shown in the PR description) is part of this PR along with the e2e tests for this, I think there was some confusion.

testing/e2e/e2e_api_test.go
nidhi-singh02: my bad, Thank you! Addressed.

node-api/handlers/beacon/validators.go
nidhi-singh02: One way is to create another version of BindAndValidate that convert unnamed JSON array to a struct which I feel would be unnecessary. As the current approach is straightforward. I am open to suggestions :)

calbera (approved):
LGTM! I like that we can e2e test this in the code very specifically

testing/e2e/e2e_api_test.go
```diff
@@ -205,3 +253,271 @@ func (s *BeaconKitE2ESuite) TestValidatorsWithSpecificStatus() {
...
+	s.Require().NoError(err)
+	s.Require().NotNil(balancesResp)
+
+	// Verify the response is not empty
+	s.Require().NotNil(balancesResp.Data)
+	s.Require().Len(balancesResp.Data, len(indices))
+	s.Require().NotEmpty(balancesResp.Data)
+
+	// Verify balance data
+	for index, balance := range balancesResp.Data {
```
abi87, line 298: nit: @nidhi-singh02 can we assert the exact lenght, (same as `len(indices)` I expect)

testing/e2e/e2e_api_test.go
```diff
@@ -205,3 +253,271 @@ func (s *BeaconKitE2ESuite) TestValidatorsWithSpecificStatus() {
...
+		State:   utils.StateIDHead,
+		Indices: indices,
+	})
+	s.Require().NoError(err)
+	s.Require().NotNil(balancesResp)
+
+	// Verify the response is not empty
+	s.Require().NotNil(balancesResp.Data)
+	s.Require().Len(balancesResp.Data, len(indices))
+	s.Require().NotEmpty(balancesResp.Data)
```
abi87, line 295: replace with `s.Require().Len(balancesResp.Data, len(indices))`?

testing/e2e/e2e_api_test.go
```diff
@@ -205,3 +253,271 @@ func (s *BeaconKitE2ESuite) TestValidatorsWithSpecificStatus() {
...
+			Indices: indices,
+		},
+	)
+
+	s.Require().NoError(err)
+	s.Require().NotNil(balancesResp)
+	s.Require().Len(balancesResp.Data, len(indices))
+
+	// Verify all requested indices are present
+	returnedIndices := make(map[phase0.ValidatorIndex]bool)
```
abi87, line 325: nit: consider `map[phase0.ValidatorIndex]struct{}` instead of `map[phase0.ValidatorIndex]bool`. The former is the idiomatic way to specify sets in go

testing/e2e/e2e_api_test.go
```diff
@@ -205,3 +253,271 @@ func (s *BeaconKitE2ESuite) TestValidatorsWithSpecificStatus() {
...
+	pubkey := validator.Validator.PublicKey
+
+	balancesResp, err := client.ValidatorBalances(s.Ctx(), &beaconapi.ValidatorBalancesOpts{
+		State:   utils.StateIDHead,
+		Indices: []phase0.ValidatorIndex{},
+		PubKeys: []phase0.BLSPubKey{pubkey},
+	})
+	s.Require().NoError(err)
+	s.Require().NotNil(balancesResp)
+	s.Require().NotEmpty(balancesResp.Data)
```
abi87, line 375: Consred asserting lenght is exactly 1

consensus-types/types/validator.go
```diff
@@ -364,6 +364,5 @@ func (v *Validator) Status(currentEpoch math.Epoch) (string, error) {
 		}
 		return constants.ValidatorStatusWithdrawalDone, nil
 	}
-
```
abi87, line 367: super nit: consider removing this (unnecessary diff)

abi87 (commented):
Minor nits around tests. LGTM once addressed

testing/e2e/e2e_api_test.go
nidhi-singh02: We already have `s.Require().Len(balancesResp.Data, len(indices))` at line no 294.

nidhi-singh02: You want me to remove NotNil and NotEmpty check ?

testing/e2e/e2e_api_test.go
nidhi-singh02: addressed

testing/e2e/e2e_api_test.go
abi87: I have missed line 294, which is fine. I think line 295 is redundant (and a weaker check) so we can drop it.

consensus-types/types/validator.go
nidhi-singh02: done!

testing/e2e/e2e_api_test.go
nidhi-singh02: done!

testing/e2e/e2e_api_test.go
nidhi-singh02: Make sense, addressed. removed from one more place.

nidhi-singh02:
> Minor nits around tests. LGTM once addressed

Addressed, Kindly check.

testing/e2e/e2e_api_test.go
```diff
@@ -266,7 +266,6 @@ func (s *BeaconKitE2ESuite) TestValidatorBalances() {
 
 	// Verify the response is not empty
 	s.Require().NotNil(balancesResp.Data)
-	s.Require().NotEmpty(balancesResp.Data)
```
abi87: I think this was necessary right?

abi87 (approved):

PR #2484, opened 2025-02-06
chore(networks): 80094 network configs

calbera:


rezbera (approved):
lgtm

shotes:
I think we should have a top level `network` directory that contains the configs that are expected to be used by a node runner, such as mainnet and any public testnets.

It feels weird to put mainnet configs in a testing directory.

nidhi-singh02 (approved):
lgtm

PR #2487, opened 2025-02-06
refactor(depinject): Replace Concrete CometBFT implementation with Interface

rezbera:
We are underutiltising the power of depinject. This PR abstracts everything we need from the CometBFT Service into an interface. This will allow me to implement a different version of the cometbft service which I can use for testing and unblocks https://github.com/berachain/beacon-kit/pull/2411

ComosSDK Depinject has issues if there are multiple identical interfaces, which may be the case for us in the future if we turn other concrete types into interfaces. The proposed solution for this future problem would be to implement a NOOP method on the interface to differentiate it from others.

### DIFF
--- a/consensus/cometbft/service/service.go
+++ b/consensus/cometbft/service/service.go
@@ -52,7 +52,7 @@ const (
 	appName           string = "beacond"
 )
 
-// ConsensusService defines everything we utilise externally from CometBFT
+// ConsensusService defines everything we utilise externally from CometBFT.
 type ConsensusService interface {
 	Start(ctx context.Context) error
 	Stop() error
--- a/consensus/cometbft/service/service.go
+++ b/consensus/cometbft/service/service.go
@@ -52,6 +52,18 @@ const (
 	appName           string = "beacond"
 )
 
+// ConsensusService defined everything we utilise externally from CometBFT
+type ConsensusService interface {
+	Start(ctx context.Context) error
+	Stop() error
+	Name() string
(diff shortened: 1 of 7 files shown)


consensus/cometbft/service/service.go
```diff
@@ -52,6 +52,18 @@ const (
 	appName           string = "beacond"
 )
 
+// ConsensusService defines everything we utilise externally from CometBFT.
+type ConsensusService interface {
```
fridrik01, line 56: did you consider declaring at the caller side, for example in node-core/types.go?

node-core/components/cometbft_service.go
```diff
@@ -41,7 +41,7 @@ func ProvideCometBFTService(
 	cmtCfg *cmtcfg.Config,
 	appOpts config.AppOptions,
 	telemetrySink *metrics.TelemetrySink,
-) *cometbft.Service {
+) cometbft.ConsensusService {
```
fridrik01, line 44: here we could still return the concrete type?

fridrik01 (approved):
had just minor comments, but otherwise lgtm.

consensus/cometbft/service/service.go
rezbera: Good call - i wasn't sure of a good home for this but that seems better being away from the concrete impl

node-core/components/cometbft_service.go
rezbera: good catch

PR #2485, opened 2025-02-06
fix(engine api): [1/3] error on engine API errors

shotes:
1. Enforces the synchronicity of the `FinalizeBlock`'s `engine_forkchoiceUpdated` call.
2. Always returns error on `engine_newPayload` and `engine_forkchoiceUpdated` failures. Therefore
3. Remove optimistic engine from `transaction.Context` and the engine.

With both of these, we ensure that the CL will fail when it goes out of sync with the EL.

In future PRs, we should attempt retries on these engine API calls to allow for easier recovery.

PR #2488, opened 2025-02-06
fix(engine api): [2/3] Add force head in FinalizeBlock

shotes:
~~This enforces the `engine_forkchoiceUpdated` is always the first Engine API call made by a node, regardless of where it is during sync, or if it is genesis, or if it is even a validator node.~~

~~This will always notify the Execution Layer Client of the head of the chain on startup.~~

Edit: After some deliberation with the team, this PR has been revised:

On node startup, the first call to the Engine API is `engine_newPayload`. This notifies the engine client of the payload so that it can store it in its block cache. 

Afterwards, `engine_forkchoiceUpdated` is called to make that payload the new head of the chain. This does the following:
- if not in sync, triggers the execution layer client to start syncing from its peers.
- if in sync already, sets the head of the chain in the EL client and we're ready to go.

NOTE: the variable `forceStartupSyncOnce` is shared between both force startup sync instances, so this should only ever run once in all scenarios.


beacon/blockchain/payload.go
shotes, line 1: More thought is required around line 47:
```go
	// TODO: Verify if the slot number is correct here, I believe in current
	// form it should be +1'd. Not a big deal until hardforks are in play though.
	if err = s.localBuilder.SendForceHeadFCU(ctx, st, slot+1); err != nil {
```

The slot number is used to determine the `ForkVersion` to send to the EL client. Why is the slot number incremented by 1 for this?

beacon/blockchain/finalize_block.go
```diff
@@ -43,6 +43,12 @@ func (s *Service) FinalizeBlock(
 		finalizeErr error
 	)
 
+	// Send an FCU to force the HEAD of the chain on the EL on startup.
+	s.forceStartupSyncOnce.Do(func() { finalizeErr = s.forceStartupHead(ctx, s.storageBackend.StateFromContext(ctx)) })
```
rezbera, line 47: This doesn't seem like it will error if the EL is 'SYNCING'. 

If it does return 'syncing' - that would be an unexpected state we should error on right?

shotes: This does not resolve any `SYNCING` error. With the previous PR https://github.com/berachain/beacon-kit/pull/2485, the CL and EL should never be out of sync with each other.

This simply enforces that the very first engine API call to the EL is a `engine_forkchoiceUpdated`. Previously, if you restarted both your CL and EL during bootstrap sync, the very first call would be a `engine_newPayload`. This could result in the EL responding at startup with [`forced head needed for startup`](https://github.com/ethereum/go-ethereum/blob/d11e9c0b517364ae3a372af48f0b34f754e1dbc0/eth/downloader/skeleton.go#L268), and would result in a `SYNCING` status return, which would immediately error with the new erroring changes.

rezbera: > With the previous PR https://github.com/berachain/beacon-kit/pull/2485, the CL and EL should never be out of sync with each other.

Correct. Which is why if we _do _ see SYNCING, then that's unexpected and we should error on

rezbera: Update: NVM we do return error on SYNCING

beacon/blockchain/process_proposal.go
```diff
@@ -219,10 +219,11 @@ func (s *Service) VerifyIncomingBlock(
...
 
 	// Force a sync of the startup head if we haven't done so already.
-	//
-	// TODO: This is a super hacky. It should be handled better elsewhere,
-	// ideally via some broader sync service.
-	s.forceStartupSyncOnce.Do(func() { s.forceStartupHead(ctx, preState) })
+	var fcuErr error
+	s.forceStartupSyncOnce.Do(func() { fcuErr = s.forceStartupSync(ctx, beaconBlk) })
+	if fcuErr != nil {
+		return fcuErr
```
shotes, line 225: Should we trigger a panic here? If `forceStartupSyncOnce` (which runs only once at startup) fails, then something is already wrong with our CL <-> EL comms or sync. This returns error, which will simply reject the proposal and then continue on, never to run `forceStartupSyncOnce` again. I think this should panic.

rezbera: My opinion is that forceStartupSyncOnce can be removed from process_proposal since it's not very realistic to occur - but doesn't need to change in this PR

calbera: Can we add a TODO in the code to address this later? might be worth completely removing from ProcessProposal now.

shotes: Added TODO

calbera (approved):
tACK

beacon/blockchain/payload.go
```diff
@@ -22,34 +22,61 @@ package blockchain
...
+	parentBeaconBlockRoot := beaconBlock.GetParentBlockRoot()
+	if err := s.executionEngine.VerifyAndNotifyNewPayload(
+		ctx, ctypes.BuildNewPayloadRequest(
+			executionPayload,
+			beaconBlock.GetBody().GetBlobKzgCommitments().ToVersionedHashes(),
+			&parentBeaconBlockRoot,
+		),
+	); err != nil {
+		// Don't return error here, because we want to send the forkchoice update regardless.
+		s.logger.Warn("failed to push new payload during force startup head", "error", err)
```
abi87, line 53: q: wonder if we should handle different errors differently? I would expect the only benign one is SYNCING? Maybe also ACCEPTED would be fine, given the next FCU, but that should never happen really.

beacon/blockchain/payload.go
```diff
@@ -22,34 +22,61 @@ package blockchain
...
-	// TODO: Verify if the slot number is correct here, I believe in current
-	// form it should be +1'd. Not a big deal until hardforks are in play though.
-	if err = s.localBuilder.SendForceHeadFCU(ctx, st, slot+1); err != nil {
-		s.logger.Error(
-			"failed to send force head FCU",
-			"error", err,
-		)
+	// Submit the forkchoice update to the EL client. This will ensure that it is either synced or
+	// starts up a sync.
+	var attrs *engineprimitives.PayloadAttributes
```
abi87, line 58: nit: we don't need this? This should be equivalent to set `PayloadAttributes` to nil

beacon/blockchain/payload.go
```diff
@@ -22,34 +22,61 @@ package blockchain
...
+			State: &engineprimitives.ForkchoiceStateV1{
+				HeadBlockHash:      executionPayload.GetBlockHash(),
+				SafeBlockHash:      executionPayload.GetParentHash(),
+				FinalizedBlockHash: executionPayload.GetParentHash(),
+			},
+			PayloadAttributes: attrs,
+			ForkVersion:       s.chainSpec.ActiveForkVersionForSlot(beaconBlock.GetSlot()),
+		},
+	)
+	if err != nil {
```
abi87, line 70: nit: handling `err ==nil` first reduces indentation.

abi87 (commented):
nits and a couple of questions around error handling

beacon/blockchain/payload.go
```diff
@@ -22,34 +22,61 @@ package blockchain
...
-		)
+	// Submit the forkchoice update to the EL client. This will ensure that it is either synced or
+	// starts up a sync.
+	var attrs *engineprimitives.PayloadAttributes
+	_, _, err := s.executionEngine.NotifyForkchoiceUpdate(
+		ctx, &ctypes.ForkchoiceUpdateRequest{
+			State: &engineprimitives.ForkchoiceStateV1{
+				HeadBlockHash:      executionPayload.GetBlockHash(),
+				SafeBlockHash:      executionPayload.GetParentHash(),
+				FinalizedBlockHash: executionPayload.GetParentHash(),
```
abi87, line 64: so here we must to this vs using `GetBlockHash` for each entry cause we use this call for ProcessProposal as well?
What if the first block we happen to verify is not valid?

beacon/blockchain/process_proposal.go
shotes: Undid any changes to the forceStartup functionality in ProcessProposal.

beacon/blockchain/payload.go
shotes: Added switches to both calls to better handle errors.

beacon/blockchain/payload.go
shotes: Good question - this is no longer a concern, as I have split the implementation between the ProcessProposal force startup and the FinalizeBlock force startup. This PR no longer touches the ProcessProposal version.

beacon/blockchain/payload.go
shotes: Added more explicit error handling.

beacon/blockchain/payload.go
```diff
@@ -52,6 +56,74 @@ func (s *Service) forceStartupHead(
...
+) error {
+	// NewPayload call first to load payload into EL client.
+	executionPayload := beaconBlock.GetBody().GetExecutionPayload()
+	parentBeaconBlockRoot := beaconBlock.GetParentBlockRoot()
+	payloadReq := ctypes.BuildNewPayloadRequest(
+		executionPayload,
+		beaconBlock.GetBody().GetBlobKzgCommitments().ToVersionedHashes(),
+		&parentBeaconBlockRoot,
+	)
+	if err := payloadReq.HasValidVersionedAndBlockHashes(); err != nil {
```
abi87, line 74: I was the main proposer of pulling `HasValidVersionedAndBlockHashes` out of executionEngine and just have `NotifyNewPayload`, but now it's annoying to have to call `HasValidVersionedAndBlockHashes` explicitly.
We should consider letting `BuildNewPayloadRequest` return an error and have `HasValidVersionedAndBlockHashes` called in it.

abi87 (approved):

PR #2490, opened 2025-02-07
feat(engine api): [3/3] Engine API retry

shotes:
This PR introduces retries to `engine_newPayload` and `engine_forkchoiceUpdated` calls.

This PR introduces a third party [go port](https://github.com/cenkalti/backoff) of [google-http-client ExponentialBackOff.java](https://github.com/googleapis/google-http-java-client/blob/da1aa993e90285ec18579f1553339b00e19b3ab5/google-http-client/src/main/java/com/google/api/client/util/ExponentialBackOff.java). I have read through all of the code that the third party package implements. Please verify that package as well.

All changes:
- [x] retry `engine_newPayload` and `engine_forkchoiceUpdated` calls:
  - [x] `rpc-retries` can be set to 0 for infinite retries and is used to determine the number of times to retry the engine api call
  - [x] `rpc-retry-interval` new config value that determine the starting backoff between engine api call retries.
  - [x] `rpc-max-retry-interval` new config value that determines the maximum backoff time between engine api call retries. This is needed because the backoff is set to increase exponentially.
- [x] Aggregate ACCEPTED and SYNCING metric function because they were being treated the exact same anyways. I only did this because the linter was complaining about the length of the NotifyNewPayload function.
- [x] During `forceSyncUponFinalize`, call `NotifyNewPayload` with `retryOnSyncingStatus` set to true. We don't want to error when there is a SYNCING status during this newPayload call.
- [x] Add todo to extend `retryOnSyncingStatus = false` to FinalizeBlock as well. This is not required, but will make it such that we could delete `forceSyncUponFinalize` completely.
- [x] Retry is context-cancellable - see https://github.com/berachain/beacon-kit/pull/2568


execution/engine/engine.go
```diff
@@ -33,6 +34,8 @@ import (
 	jsonrpc "github.com/berachain/beacon-kit/primitives/net/json-rpc"
 )
 
+const engineAPITimeout = time.Minute * 5
```
shotes, line 37: Let's discuss this timeout value.

execution/engine/retry.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/berachain/beacon-kit/errors"
+)
+
+const (
+	individualRetryTimeout = time.Minute * 1
+
+	// These values are based on the default values specified at
+	// https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md.
+	baseDelay  = 100 * time.Millisecond
```
shotes, line 36: also consider these const values. I used default backoffs from google grpc, but we should modify these to our own specification.

fridrik01 (commented):
@shotes I checked out this branch (to check whether it resolved some syncing issue I had locally) and I observed that now beacond hangs when it receives SIGTERM if its currently inside a retry. It requires to either wait up to 5 minutes or send it a SIGKILL to exit. This behaviour was not in deposit-sync-force-startup-fcu branch.

execution/engine/retry.go
```diff
@@ -0,0 +1,108 @@
+// SPDX-License-Identifier: BUSL-1.1
```
rezbera, line 1: This feels like a library that likely exists already - have we tried any existing libs which didn't meet our requirements?

shotes: Good point, shouldn't attempt to engineer a solution that may exist. Looking for alternatives.

shotes: Removed `retryWithTimeout`, using a go port uses the implementation from google HTTP client for java.

shotes:
> @shotes I checked out this branch (to check whether it resolved some syncing issue I had locally) and I observed that now beacond hangs when it receives SIGTERM if its currently inside a retry. It requires to either wait up to 5 minutes or send it a SIGKILL to exit. This behavior was not in `deposit-sync-force-startup-fcu` branch.
> 
> EDIT: I think the main reason is that the context passed to `NotifyForkchoiceUpdate` is not properly cancelled. Let me know if you want me to take a look at that, I can push a separate PR on top of this one, should have time this weekend.

I am seeing this issue as well, having to specifically kill the process each time I restart. Just pushed up changes that uses a different lib, still seeing the same issue. Investigating now - feel free to test and push any fixes.

fridrik01:
> > @shotes I checked out this branch (to check whether it resolved some syncing issue I had locally) and I observed that now beacond hangs when it receives SIGTERM if its currently inside a retry. It requires to either wait up to 5 minutes or send it a SIGKILL to exit. This behavior was not in `deposit-sync-force-startup-fcu` branch.
> > EDIT: I think the main reason is that the context passed to `NotifyForkchoiceUpdate` is not properly cancelled. Let me know if you want me to take a look at that, I can push a separate PR on top of this one, should have time this weekend.
> 
> @fridrik01 I am seeing this issue as well, having to specifically kill the process each time I restart. Just pushed up changes that uses a different lib, still seeing the same issue. Investigating now - feel free to test and push any fixes.

 @shotes I pushed https://github.com/berachain/beacon-kit/pull/2496 which addresses hanging on SIGTERM, however it unwraps an issue it seems in cometbft where it panics on a cancelled context. Need to figure out what is going on there, maybe cometbft does not support ctx cancellation

execution/engine/retry.go
fridrik01: In this case I would just have used what you had already written instead of adding another new upstream dependency for such a simple feature.

execution/engine/engine.go
```diff
@@ -74,129 +74,190 @@ func (ee *Engine) NotifyForkchoiceUpdate(
...
-	// Notify the execution engine of the forkchoice update.
-	payloadID, err := ee.ec.ForkchoiceUpdated(
-		ctx,
-		req.State,
-		req.PayloadAttributes,
-		req.ForkVersion,
-	)
+	// Configure backoff.
+	engineAPIBackoff := backoff.NewExponentialBackOff()
+	engineAPIBackoff.InitialInterval = ee.ec.GetRPCTimeout()
```
shotes, line 81: Now configured with RPC Timeout as the intial backoff interval.

execution/engine/engine.go
```diff
@@ -74,129 +74,190 @@ func (ee *Engine) NotifyForkchoiceUpdate(
...
+			// Retry on unknown errors, we'll log the error and retry.
+			// TODO: discriminate more of these errors:
+			//     RPC Timeout Errors
+			//     Connection Refused Errors
+			//     Erroneous Parsing Errors
+			return nil, innerErr
+		}
+	},
+		backoff.WithBackOff(engineAPIBackoff),
+		backoff.WithMaxTries(uint(ee.ec.GetRPCRetries())),
```
shotes, line 151: Instead of using an arbitrary overall max timeout, we use the max number of configurable RPC retries, combined with the backoff.

testing/networks/80069/app.toml
```diff
@@ -124,6 +124,9 @@ rpc-retries = "3"
 # RPC timeout for execution client requests.
 rpc-timeout = "900ms"
 
+# RPC retry interval for execution client requests.
+rpc-retry-interval = "100ms"
```
shotes, line 128: I added to all `app.toml`s. However, this will still work (and use the default 100ms set in beacon-kit) without this explicitly set in the app.toml, so we do not need to update configs of already-existing nodes.

execution/engine/engine.go
shotes: Now configured with RPC Retry Interval as the initial backoff interval. Default:
```
# RPC retry interval for execution client requests.
rpc-retry-interval = "100ms"
```

execution/engine/engine.go
```diff
@@ -74,129 +75,188 @@ func (ee *Engine) NotifyForkchoiceUpdate(
...
-	// Notify the execution engine of the forkchoice update.
-	payloadID, err := ee.ec.ForkchoiceUpdated(
-		ctx,
-		req.State,
-		req.PayloadAttributes,
-		req.ForkVersion,
-	)
+	// Configure backoff.
+	engineAPIBackoff := backoff.NewExponentialBackOff()
+	engineAPIBackoff.InitialInterval = ee.ec.GetRPCRetryInterval()
```
shotes, line 82: Now configured with RPC Retry Interval as the initial backoff interval. Default:
```
# RPC retry interval for execution client requests.
rpc-retry-interval = "100ms"
```

shotes:
For testing, make sure to update `rpc-retries` in your `app.toml` to 0.

Tested on mainnet node:
- During SYNC, kill EL node only. You should see the CL unable to connect to the EL, and it will keep retrying infinitely (with a maximum 10s backoff). Start EL node again, and the two should continue on like nothing happened.
- Wipe EL node state. On startup, the CL should send newPayload and FCU, then wait forever until EL catches up.

state-transition/core/state_processor_payload.go
```diff
@@ -169,7 +169,9 @@ func (sp *StateProcessor) validateStatefulPayload(
 		return err
 	}
 
-	if err = sp.executionEngine.NotifyNewPayload(ctx, payloadReq); err != nil {
+	// TODO: set retryOnSyncingStatus to false if we are in FinalizeBlock.
+	// Otherwise leave as true. This is ok to leave this way for now.
+	if err = sp.executionEngine.NotifyNewPayload(ctx, payloadReq, true); err != nil {
```
shotes, line 174: Leaving this as `true` right now leaves the behavior the same as you'd expect - during all state transition calls, it will always retry on all SYNCING/ACCEPTED statuses.

I believe this can be changed to false during FinalizeBlock, such that we can ignore SYNCING in newPayload because we'll continue on to the FCU, and that will wait for the node to be SYNCED up.

shotes:
For testing, make sure to update `rpc-retries` in your `app.toml` to 0.

Successfully tested on my local mainnet node:
- During SYNC, kill EL node only. You should see the CL unable to connect to the EL, and it will keep retrying infinitely (with a maximum 10s backoff). Start EL node again, and the two should continue on like nothing happened.
- Wipe EL node state. Restart both EL and CL. The CL should send newPayload (and allow the SYNCING status, only on startup) and FCU, then wait forever until EL catches up.

execution/engine/engine.go
```diff
@@ -74,129 +77,184 @@ func (ee *Engine) NotifyForkchoiceUpdate(
...
+		// Max duration so we don't ever hit it before maxRetries.
+		backoff.WithMaxElapsedTime(time.Duration(math.MaxInt64)),
+	)
 }
 
 // NotifyNewPayload notifies the execution client of the new payload.
 func (ee *Engine) NotifyNewPayload(
 	ctx context.Context,
 	req *ctypes.NewPayloadRequest,
+	retryOnSyncingStatus bool,
```
shotes, line 164: I'm not obsessed with this `retryOnSyncingStatus` flag usage. Let me know if this needs to change.

testing/networks/80069/app.toml
```diff
@@ -119,11 +119,18 @@ datadog-hostname = "my_beacond_node"
 rpc-dial-url = "http://localhost:8551"
 
 # Number of retries before shutting down consensus client.
-rpc-retries = "3"
```
abi87, line 122: @shotes IIUC, this config was never used, despite being read?

execution/engine/engine.go
```diff
@@ -74,129 +77,184 @@ func (ee *Engine) NotifyForkchoiceUpdate(
...
+			//     RPC Timeout Errors
+			//     Connection Refused Errors
+			//     Erroneous Parsing Errors
+			return nil, innerErr
+		}
+	},
+		backoff.WithBackOff(engineAPIBackoff),
+		backoff.WithMaxTries(maxRetries),
+		// Max duration so we don't ever hit it before maxRetries.
+		backoff.WithMaxElapsedTime(time.Duration(math.MaxInt64)),
```
abi87, line 156: is it safe to use `math.MaxInt64` as `time.Duration`? https://brandur.org/fragments/go-max-time-duration

execution/engine/engine.go
```diff
@@ -74,129 +77,184 @@ func (ee *Engine) NotifyForkchoiceUpdate(
...
 ) error {
-	// Log the new payload attempt.
-	ee.metrics.markNewPayloadCalled(
-		req.ExecutionPayload.GetBlockHash(),
-		req.ExecutionPayload.GetParentHash(),
-	)
+	// Configure backoff. This will retry maxRetries number of times.
+	// Specifying 0 maxRetries will retry infinitely. Between each retry, it
+	// will wait RPCRetryInterval amount of time. This backoff will increase
+	// exponentially until it reaches RPCMaxRetryInterval.
```
abi87, line 169: why do we have backoff on NewPayload too?
I understand it can return `syncing` if called into `forceSyncUponFinalize` but I think having syncing returned there is fine and does not need retrial?

state-transition/core/state_processor_payload.go
abi87: I would expect the opposite, so setting this to false for `ProcessProposal` and let it possibly err if the EL is not ready and then set it ` true` during `FinalizeBlock` to have that waiting on the EL to catch up.

testing/networks/80069/app.toml
shotes: Indeed, it was never used previously. It is now used in this PR though.

execution/engine/engine.go
shotes: It is safe. That blog post specifies a case in which they were doing exponential arithmetic on the duration that overflowed.

The `time.Duration(math.MaxInt64)` will never overflow because we never do any arithmetic on it.

shotes: However, instead I can just set the duration to 0, and it will not use the max elapsed time at all. This is a better alternative and is more explicitly what we are looking for.

state-transition/core/state_processor_payload.go
shotes: Then if there is an error in `NotifyNewPayload` during `ProcessProposal`, we will automatically vote `REJECT` on the block instead of abstaining. We should not vote `REJECT` unless we determine the block to be bad. And we should not vote `ACCEPT` unless we determine the block to be good. And until we can make the determination that the block is good or bad, we should retry, no?

execution/engine/engine.go
shotes: I'm not sure I understand the question about backoff on NewPayload. Are you asking why we retry failures on NewPayload?

shotes:
Did some devnet testing for some learning - If a single validator node in a network of many nodes pauses to simulate retries:
- pause 2 minutes in `ProcessProposal`: The node simply pauses. It will not proceed to `FinalizeBlock` even as the network continues on. It will not ACCEPT or REJECT a block. It will also not propose new blocks during this time. Comet probably prevotes nil underneath the hoods, but at the app layer, we never continue on. Once it finishes the 2 minute pause, it will continue again and catch up to the network.
- pause 2 minutes in `FinalizeBlock`: Same thing.


This was all done with some code modifications and `make start-devnet`. If you need details lmk.

shotes:
> Awesome testing so far. I would only suggest one change that is:
> 
> * default should be permanent errors and we should only retry on specific errors (like syncing, possibly API timeout), making sure we log the errors with warning so not to miss them.
>   Other than that this looks fine with me

Noted. I have added an `IsFatalError()` and `IsNonFatalError()` to distinguish what should be retried and what shouldn't be:
- Fatal: All JSON-RPC errors as well as connection refused errors return `backoff.Permanent(err)`. 
- Non Fatal: Currently only timeout errors are non-fatal and will be retried.

I have also made the default behavior (for unknown errors) to return `backoff.Permanent(err)`.

abi87 (approved):
LGTM

PR #2493, opened 2025-02-07
Merge after Testing - Updated Genesis Files for Geth

rezbera:


abi87:
Done in https://github.com/berachain/beacon-kit/pull/2753

PR #2496, opened 2025-02-08
fix(abci): Add cometbft service cancellable context 

fridrik01:
Currently, beacond hangs on SIGTERM if its inside a retry timeout. This is because the context used in the retry is not cancelled in our signal hander. 

To address this, this PR makes the context available to the `cometbft` service which it gets from the context passed in to `Start()`. Now, when signal handler catches a signal, the context cancellation will be checked by the ABCI++ endpoints properly. This causes us to properly handle signal cancellation in the retry timeout. 

Unfortunately, CometBFT does not yet support context cancellation, or context usage at all. It currently passes in `context.TODO()` into all of the ABCI methods, so the ctx we see in the ABCI function implementations is not connected to the ctx from our node. I also can't seem to find anything in cometBFT RFC or docs that indicate supporting this.

Since CometBFT does not support context cancellation, this means that there is no clean way out -- we MUST return error in `ProcessProposal` and `FinalizeBlock` (which causes the `CONSENSUS FAILURE!!!` panic). If we don't then our node will be attempting to participate in consensus with potentially corrupt, invalid, or missing data. I have tested this behavior by not returning an error from `FinalizeBlock` during context cancellation, and indeed my node still proceeded to call `Commit`.


consensus/cometbft/service/finalize_block.go
```diff
@@ -73,28 +80,40 @@ func (s *Service) finalizeBlockInternal(
...
-	valUpdates, err := iter.MapErr(
-		finalizeBlock,
-		convertValidatorUpdate[cmtabci.ValidatorUpdate],
-	)
-	if err != nil {
-		return nil, err
-	}
+	go func() {
+		finalizeBlock, err := s.Blockchain.FinalizeBlock(
+			s.finalizeBlockState.Context(),
```
shotes, line 88: I fear that this whole context thing is partially incomplete. In the current state of the code, we are successfully stopping the node. However, we are stopping the node because the `FinalizeBlock` ABCI method returns and the app closes. This does NOT mean that this call to `s.Blockchain.FinalizeBlock(s.finalizeBlockState.Context(), req)` is also respecting to the context cancellation. So, the engine-api retry loop is still going on in the background while the node is shutting down services.

In the end, the node still shuts down, which is an improvement. However, this could result in a weird situation where this `finalizeBlockInternal` function returns, deleting the `finalizeResultCh`, and then `s.Blockchain.FinalizeBlock()` before the node shuts down, resulting in a panic of "writing to a closed channel"

consensus/cometbft/service/finalize_block.go
```diff
@@ -73,28 +80,40 @@ func (s *Service) finalizeBlockInternal(
...
+			req,
+		)
+		finalizeResultCh <- finalizeResult{finalizeBlock, err}
+	}()
+
+	select {
+	// Node will panic on context cancel with "CONSENSUS FAILURE!!!" due to error.
+	// We expect this to happen and do not want to finalize any incomplete or invalid state.
+	// TODO: switch to ctx.Done() when CometBFT implements contexts.
+	case <-s.ctx.Done():
```
shotes, line 98: Must return error here in FinalizeBlock and ProcessProposal, resulting in CONSENSUS FAILURE.

consensus/cometbft/service/finalize_block.go
shotes: Could I do something stupid simple like set:
```go
if s.finalizeBlockState == nil {
	s.finalizeBlockState = s.resetState(s.ctx) // <-- changed from `s.resetState(ctx)`
}
```

shotes: Ah, I've come up with a simpler solution. Simply combine the cometBFT context with the original context. This can then be combined before hitting the cosmos layer, so we do not have to special case `ctx` vs `s.ctx`, or deal with handling the cosmos SDK contexts.

consensus/cometbft/service/prepare_proposal.go
```diff
@@ -34,6 +34,12 @@ func (s *Service) prepareProposal(
 	ctx context.Context,
 	req *cmtabci.PrepareProposalRequest,
 ) (*cmtabci.PrepareProposalResponse, error) {
+	// Check if ctx is still good. CometBFT does not check this.
+	if ctx.Err() != nil {
```
shotes, line 38: These checks are here such that CometBFT does not continue calling them once the context has been cancelled. We should immediately check for context, instead of getting all the way to the engine-api calls and then realizing the context is cancelled.

PR #2499, opened 2025-02-09
fix(deposits): Modify Deposit Verification Logic

rezbera:
This PR modifies the deposit logic such that we trust the deposits in the block in Finalize, which helps users avoid syncing issues. 
However, once at the tip of the chain, users may still encounter deposit issues with this approach but this is less likely due to the lower relative frequency of deposits at the tip of the chain vs historical deposits. 

how can we test this better?

### DIFF
--- a/beacon/blockchain/finalize_block.go
+++ b/beacon/blockchain/finalize_block.go
@@ -90,7 +90,21 @@ func (s *Service) FinalizeBlock(
 
 	// fetch and store the deposit for the block
 	blockNum := blk.GetBody().GetExecutionPayload().GetNumber()
-	s.depositFetcher(ctx, blockNum)
+
+	if isSyncing(req.Height, req.SyncingToHeight) {
+		s.logger.Info(
+			"Syncing in finalize_block, skipping deposit fetcher",
+			"height", req.Height, "syncing_to_height", req.SyncingToHeight,
+		)
+		// If we're syncing, we just use the consensus finalized block deposits as an optimization
+		// The DepositStore does not get included in AppHash calculation and hence introducing this will not result in AppHash.
+		if err = s.storageBackend.DepositStore().EnqueueDeposits(ctx, blk.GetBody().GetDeposits()); err != nil {
+			return nil, fmt.Errorf("failed to enqueue deposits: %w", err)
+		}
+	} else {
+		// If we're NOT syncing we want to run the deposit fetcher as we want to have our own view of the deposits which we can validate in ProcessProposal.
+		s.depositFetcher(ctx, blockNum)
+	}
 
 	// store the finalized block in the KVStore.
 	// TODO: Store full SignedBeaconBlock with all data in storage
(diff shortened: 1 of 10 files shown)


state-transition/core/state_processor_staking.go
```diff
@@ -50,17 +50,22 @@ func (sp *StateProcessor) processOperations(
...
+			st,
+			deposits,
+			blk.GetBody().GetEth1Data().DepositRoot,
+		); err != nil {
+			return err
+		}
 	}
 
 	for _, dep := range deposits {
+		// We want to update our local state regardless as `processDeposit` is idempotent
```
rezbera, line 68: TODO: is this actually idempotent?

beacon/validator/block_builder.go
```diff
@@ -381,6 +381,7 @@ func (s *Service) computeStateRoot(
 		WithVerifyPayload(false).
 		WithVerifyRandao(false).
 		WithVerifyResult(false).
+		WithVerifyDeposits(true).
```
calbera, line 384: may not be necessary here, could be false

state-transition/core/state_processor_staking.go
calbera: Not sure because if `st.SetEth1DepositIndex(eth1DepositIndex + 1)` gets called a different amount of times, even with the same value, I think that will lead to a different AppHash

calbera:
Necessary to sync the Deposit DB with whats in the block if we skip verification of deposits

rezbera:
> Necessary to sync the Deposit DB with whats in the block if we skip verification of deposits

You're referring to `EnqueueDeposits` right?

beacon/validator/block_builder.go
rezbera: Agreed it could be false, but want to minimize the change impact scope

rezbera:
> > Necessary to sync the Deposit DB with whats in the block if we skip verification of deposits
> 
> You're referring to `EnqueueDeposits` right @calbera ?

I'm not convinced we need the EnqueueDeposits if we're not verifying but added it as it's idempotent with minimal impact

beacon/validator/block_builder.go
rezbera: Set to false

state-transition/core/state_processor_staking.go
```diff
@@ -50,14 +50,25 @@ func (sp *StateProcessor) processOperations(
...
-		ctx.ConsensusCtx(),
-		st,
-		deposits,
-		blk.GetBody().GetEth1Data().DepositRoot,
-	); err != nil {
-		return err
+	// When we're finalizing a block, given that consensus majority has already agreed to the block deposits, we
+	// assume it is correct. However, in ProcessProposal, we do not make this assumption and verify the deposits.
+	if ctx.VerifyDeposits() {
+		// 	Instead we directly compare block deposits with our local store ones.
```
calbera, line 56:
```suggestion
		// Instead we directly compare block deposits with our local store ones.
```

beacon/blockchain/finalize_block.go
calbera, line 1: skip `s.depositFetcher` if bootstrapping (`req.SyncingToHeight < req.Height`)

PR #2503, opened 2025-02-11
fix(deposits): Remove verification of deposits in FinalizeBlock

rezbera:
This PR modifies the deposit logic such that we trust the deposits in the block in Finalize, which helps users avoid syncing issues.
However, once at the tip of the chain, users may still encounter deposit issues with this approach but this is less likely due to the lower relative frequency of deposits at the tip of the chain vs historical deposits.

### DIFF
--- a/beacon/blockchain/deposit.go
+++ b/beacon/blockchain/deposit.go
@@ -34,6 +34,7 @@ import (
 const defaultRetryInterval = 20 * time.Second
 
 // depositFetcher fetches the blocks at blockNum-s.eth1FollowDistance so that they can be included in blockNum+1.
+// It also writes them to the deposit store.
 func (s *Service) depositFetcher(
 	ctx context.Context,
 	blockNum math.U64,
--- a/beacon/blockchain/deposit.go
+++ b/beacon/blockchain/deposit.go
@@ -33,6 +33,7 @@ import (
 // defaultRetryInterval processes a deposit event.
 const defaultRetryInterval = 20 * time.Second
 
+// depositFetcher fetches the blocks at blockNum-s.eth1FollowDistance so that they can be included in blockNum+1.
 func (s *Service) depositFetcher(
 	ctx context.Context,
 	blockNum math.U64,
--- a/beacon/blockchain/finalize_block.go
+++ b/beacon/blockchain/finalize_block.go
@@ -98,7 +98,21 @@ func (s *Service) FinalizeBlock(
 
 	// fetch and store the deposit for the block
 	blockNum := blk.GetBody().GetExecutionPayload().GetNumber()
-	s.depositFetcher(ctx, blockNum)
+
+	if isSyncing(req.Height, req.SyncingToHeight) {
+		s.logger.Info(
+			"Syncing in finalize_block, skipping deposit fetcher",
+			"height", req.Height, "syncing_to_height", req.SyncingToHeight,
+		)
+		// If we're syncing, we just use the consensus finalized block deposits as an optimization
+		// The DepositStore does not get included in AppHash calculation and hence introducing this will not result in AppHash.
+		if err = s.storageBackend.DepositStore().EnqueueDeposits(ctx, blk.GetBody().GetDeposits()); err != nil {
+			return nil, fmt.Errorf("failed to enqueue deposits: %w", err)
+		}
+	} else {
+		// If we're NOT syncing we want to run the deposit fetcher as we want to have our own view of the deposits which we can validate in ProcessProposal.
+		s.depositFetcher(ctx, blockNum)
+	}
 
 	// store the finalized block in the KVStore.
 	// TODO: Store full SignedBeaconBlock with all data in storage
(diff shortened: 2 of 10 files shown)


beacon/blockchain/finalize_block.go
```diff
@@ -98,7 +99,22 @@ func (s *Service) FinalizeBlock(
...
+		)
+		// If we're syncing, we just use the consensus finalized block deposits as an optimization
+		// The DepositStore does not get included in AppHash calculation and hence introducing this will not result in AppHash.
+		if err = s.storageBackend.DepositStore().EnqueueDeposits(ctx, blk.GetBody().GetDeposits()); err != nil {
+			return nil, fmt.Errorf("failed to enqueue deposits: %w", err)
+		}
+	} else {
+		// If we're NOT syncing we want to run the deposit fetcher as we want to have our own view
+		// of the deposits which we can validate in ProcessProposal.
+		s.depositFetcher(ctx, blockNum)
```
calbera, line 116: super nit: can you just remove `blockNum` defn from line 100
```suggestion
		s.depositFetcher(ctx, blk.GetBody().GetExecutionPayload().GetNumber())
```

beacon/blockchain/deposit.go
```diff
@@ -33,6 +33,7 @@ import (
 // defaultRetryInterval processes a deposit event.
 const defaultRetryInterval = 20 * time.Second
 
+// depositFetcher fetches the blocks at blockNum-s.eth1FollowDistance so that they can be included in blockNum+1.
```
calbera, line 36: nit: can you just add here that this func ALSO writes these deposits to deposit store

calbera (approved):
tACK

beacon/validator/block_builder.go
```diff
@@ -380,6 +380,7 @@ func (s *Service) computeStateRoot(
 		WithVerifyPayload(false).
 		WithVerifyRandao(false).
 		WithVerifyResult(false).
+		WithVerifyDeposits(false).
```
abi87, line 383: q: I wonder if this should stay true? If we do not have all the deposits and we are not able to verify state, we really should not build anything.
I see other verifications are all false, but I would rather fail building that building blocks with wrong deposits.

beacon/blockchain/finalize_block.go
```diff
@@ -98,7 +99,22 @@ func (s *Service) FinalizeBlock(
...
 	blockNum := blk.GetBody().GetExecutionPayload().GetNumber()
-	s.depositFetcher(ctx, blockNum)
+
+	if isSyncing(req.Height, req.SyncingToHeight) {
+		s.logger.Info(
+			"Syncing in finalize_block, skipping deposit fetcher",
+			"height", req.Height, "syncing_to_height", req.SyncingToHeight,
+		)
+		// If we're syncing, we just use the consensus finalized block deposits as an optimization
+		// The DepositStore does not get included in AppHash calculation and hence introducing this will not result in AppHash.
```
abi87, line 109: note: maybe the long term fix is to include deposits in the state, to be able to roll them back as the rest of the beacon state

abi87: nit: maybe the comment can be cleaned up? Something like:
`If we're syncing we do not expect the EL state to be up to date and so we don't pull deposits from it. Instead we stored deposits contained in the finalized block`

abi87 (commented):
mostly nits and a question around `isSyncing`

beacon/blockchain/finalize_block.go
```diff
@@ -194,3 +211,8 @@ func (s *Service) executeStateTransition(
 		blk.GetBeaconBlock(),
 	)
 }
+
+// isSyncing returns true if we are catching up and replaying blocks, i.e. only running FinalizeBlock.
+func isSyncing(height, syncingToHeight int64) bool {
+	return height < syncingToHeight
```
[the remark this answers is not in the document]
rezbera: @abi87 

In the commet code I see this:
```go
	// If the node is syncing/replaying blocks - target height. If not, syncing_to == height.
	SyncingToHeight int64 `protobuf:"varint,9,opt,name=syncing_to_height,json=syncingToHeight,proto3" json:"syncing_to_height,omitempty"`
```
This comment implies the following:
- if `syncing_to == height` then we are NOT syncing. As such if `height < syncing_to`, then we ARE syncing

beacon/validator/block_builder.go
rezbera: I kept this true initially but @calbera thought we should set to false for consistency. I personally think true is fine as if the block builder EL is behind, it will be rejected by peers anyway and we see new height. I don't like the idea of using the deposits check purely for the side-effect of catching a behind EL.

rezbera:
Closing this PR for now as it may not be needed with as https://github.com/berachain/beacon-kit/pull/2488 should prevent beacon nodes from entering a state where there EL is behind. 

If their EL falls behind at later point, e.g. due to a dropped NewPayload request, then BeaconKit will catch that and fail.

PR #2504, opened 2025-02-12
do not reject PRs based on missing period in a comment

fridrik01:
I don't see the value of rejecting perfectly fine PRs just because they having missing periods within their comments.


rezbera (commented):
I'll leave this one to @calbera lmao. he may have an opinion here. TBH my linter just adds them locally do I'm not fussed by it.

calbera (approved):
lmao Im good with this change. My formatter adds periods too so overall indifferent.

PR #2505, opened 2025-02-12
Fix broken CI due to using newest go version which is not supported b…

fridrik01:
…y comet


calbera (approved):
okay with this and manually updating go versions as we do in our own go.mod

rezbera (approved):

PR #2508, opened 2025-02-12
fix(engine api): [1/3] error on engine API errors (revive #2485)

shotes:
Revives https://github.com/berachain/beacon-kit/pull/2485


rezbera (commented):
Performance test results from slack

looks pretty good! but we're at ~2.11s peak. Max Finalize times well below 500ms. Should be fine with stable block times
image

execution/engine/engine.go
```diff
@@ -92,17 +92,31 @@ func (ee *Engine) NotifyForkchoiceUpdate(
...
+			ee.logger.Warn(
+				"Received nil payload ID on VALID engine response",
+				"head_eth1_hash", req.State.HeadBlockHash,
+				"safe_eth1_hash", req.State.SafeBlockHash,
+				"finalized_eth1_hash", req.State.FinalizedBlockHash,
+			)
+			return nil, nil, ErrNilPayloadOnValidResponse
+		}
+
+		return payloadID, latestValidHash, nil
```
abi87, line 107: this processing is done only for `err ==nil` so moved it up to simplify flow

abi87 (approved):

PR #2516, opened 2025-02-15
chore(state-processor): Add UTs for payload timestamp verification

abi87:
Finally added UTs for payload timestamp verification.
We had introduced and reverted by mistake this verification. These UTs make sure at least a test fails in case the validation was dropped again.
The PR required cleaning up the state generation, which I believe now is much closer to what we do in prod code (if not clearer). This should explain the somewhat large number of lines


testing/state-transition/state-transition.go
```diff
@@ -58,23 +58,30 @@ type (
 	TestStateProcessorT          = core.StateProcessor
 )
 
-type testKVStoreService struct {
-	ctx sdk.Context
```
abi87, line 62: this was a first attempt at creating state but it is wrong.
We should use the context provided by the transactions and extract the state frorm there

testing/state-transition/state-transition.go
```diff
@@ -90,12 +97,12 @@ func initTestStores() (*beacondb.KVStore, *depositstore.KVStore, error) {
...
 	if err = cms.LoadLatestVersion(); err != nil {
-		return nil, nil, fmt.Errorf("failed to load latest version: %w", err)
+		return nil, nil, nil, fmt.Errorf("failed to load latest version: %w", err)
 	}
 
-	ctx := sdk.NewContext(cms, true, nopLog)
-	testStoreService := &testKVStoreService{ctx: ctx}
-	return beacondb.New(testStoreService),
+	testStoreService := &testKVStoreService{}
+	return cms,
```
abi87, line 104: we need to export the multistore in order to be able to create temporary cache states out of it, each for a single block.
This does not affect other tests

testing/state-transition/state-transition.go
```diff
@@ -129,21 +138,23 @@ func SetupTestState(t *testing.T, cs chain.Spec) (
...
+		DummyProposerAddr,
 	).
 		WithVerifyPayload(false).
 		WithVerifyRandao(false).
 		WithVerifyResult(false).
 		WithMeterGas(false).
 		WithOptimisticEngine(true)
 
-	return sp, beaconState, depositStore, ctx
+	return sp, beaconState, depositStore, ctx, cms, execEngine
```
abi87, line 159: we need to export the mocked index because we need to set expectations right when handling the payload

state-transition/core/state_processor_payload_test.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/berachain/beacon-kit/primitives/version"
+	statedb "github.com/berachain/beacon-kit/state-transition/core/state"
+	statetransition "github.com/berachain/beacon-kit/testing/state-transition"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+// TestPayloadTimestampVerification ensures that payload timestamp
+// is properly validated
```
abi87, line 47: This test works as follow:
- perform genesis operations
- have a table test where first block had payload timestamp set in various ways. Some of the pass, other fails verification.

state-transition/core/state_processor_payload_test.go
```diff
@@ -0,0 +1,163 @@
...
+			},
+			payloadTime: consensusBlkTime.Add(2 * time.Second),
+			expectedErr: payloadtime.ErrTooFarInTheFuture,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// these test cases should not be run in parallel
+			// since state processors is shared by them
```
abi87, line 126: can't really use t.Parallel here I think. Sorry @rezbera

state-transition/core/state_processor_payload_test.go
```diff
@@ -0,0 +1,163 @@
...
+	}{
+		{
+			name: "Payload timestamp < consensus timestamp",
+			setupMocksF: func() {
+				// we don't really verify payloads here, so just mock a positive result
+				mockEngine.EXPECT().VerifyAndNotifyNewPayload(mock.Anything, mock.Anything).Return(nil)
+			},
+			payloadTime: consensusBlkTime.Add(-10 * time.Second),
+			expectedErr: nil,
+		},
```
rezbera, line 94: This test case feels a bit weird as reading it implies that a proposer can use an old timestamp. 

In reality, the execution client for reject timestamp older than parent. Perhaps make it clear in a comment

state-transition/core/state_processor_payload_test.go
```diff
@@ -0,0 +1,163 @@
...
+		{
+			name: "Payload timestamp == consensus timestamp",
```
rezbera, line 96: IIRC even an `==` payload timestamp to parent would be rejected by EL

rezbera (commented):
Nice! - just 1 comment about historical timestamps which we could clarify better in a comment

state-transition/core/state_processor_payload_test.go
```diff
@@ -0,0 +1,163 @@
...
+	genPayloadHeader.Timestamp = math.U64(genesisTime.Unix())
+
+	_, err := sp.InitializePreminedBeaconStateFromEth1(
+		st, genDeposits, genPayloadHeader, genVersion,
+	)
+	require.NoError(t, err)
+	require.NoError(t, ds.EnqueueDeposits(ctx.ConsensusCtx(), genDeposits))
+
+	// write genesis changes to make them available for next blocks
+	ctx.ConsensusCtx().(sdk.Context).MultiStore().(storetypes.CacheMultiStore).Write()
```
rezbera, line 76: is this casting the only way? i would never have figured out you can reach the cache multistore from the `ctx`

state-transition/core/state_processor_payload_test.go
abi87: What I am trying to show is that payload timestamp can be smaller then consensus, but you are right it must be strictly increasing. Will clarify

state-transition/core/state_processor_payload_test.go
abi87: I disagree, cause consensus timestamp in this case is strictly larger than genesis time.
But I definitely need to clarify this

state-transition/core/state_processor_payload_test.go
abi87: We do something like this in consensus service, but we track the cache explicitly.
To avoid the casting here I would have to pass the cache from testSetup, along with many other return types.
I felt I should have not overcomplicated that, but if you think that is better I will do it

rezbera: > To avoid the casting here I would have to pass the cache from testSetup, along with many other return types.

as is fine 👍

state-transition/core/state_processor_payload_test.go
rezbera: You are right - I got confused and thought the test was comparing change against the parent block, but its actually a test of drift against the consensus timestamp. My mistake

state-transition/core/state_processor_payload_test.go
rezbera: As above - my misunderstanding

consensus-types/types/payload_requests.go
```diff
@@ -98,53 +83,78 @@ func (n *NewPayloadRequest) HasValidVersionedAndBlockHashes() error {
...
-			UncleHash:        gethprimitives.EmptyUncleHash,
-			Coinbase:         gethprimitives.ExecutionAddress(payload.GetFeeRecipient()),
-			Root:             gethprimitives.ExecutionHash(payload.GetStateRoot()),
-			TxHash:           gethprimitives.DeriveSha(gethprimitives.Transactions(txs), gethprimitives.NewStackTrie(nil)),
-			ReceiptHash:      gethprimitives.ExecutionHash(payload.GetReceiptsRoot()),
-			Bloom:            gethprimitives.LogsBloom(payload.GetLogsBloom()),
-			Difficulty:       big.NewInt(0),
-			Number:           new(big.Int).SetUint64(payload.GetNumber().Unwrap()),
-			GasLimit:         payload.GetGasLimit().Unwrap(),
-			GasUsed:          payload.GetGasUsed().Unwrap(),
```
abi87, line 127: Refactored out int MakeEthBlock to be able to reuse it in unit tests

state-transition/core/core_test.go
```diff
@@ -125,18 +146,6 @@ func generateTestPK(t *testing.T, rndSeed int) (bytes.B48, int) {
...
-	payload := &types.ExecutionPayload{
-		Timestamp:     timestamp,
-		ExtraData:     []byte("testing"),
-		Transactions:  [][]byte{},
-		Withdrawals:   withdrawals,
-		BaseFeePerGas: math.NewU256(0),
-		EpVersion:     version.Deneb1(),
-	}
-	return payload
-}
```
abi87, line 138: absorbed into `buildNextBlock`

state-transition/core/state_processor_payload_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+// TestPayloadTimestampVerification ensures that payload timestamp
+// is properly validated
+func TestPayloadTimestampVerification(t *testing.T) {
+	// Create state processor to test
+	cs := setupChain(t)
+	sp, st, ds, ctx, cms, mockEngine := statetransition.SetupTestState(t, cs)
+
+	// process genesis before any other block
+	genesisTime := time.Now().Truncate(time.Second)
```
shotes, line 54: Could we make this genesis processing into a separate function? Seems to be used in a few different test cases, and is probably useful for initializing genesis for UTs in general.

abi87: I know it's annoying but I do not find a meaningful way to simplify genesis creation. I would leave it as is and revisit as we add new tests

consensus-types/types/payload_requests.go
```diff
@@ -98,53 +83,78 @@ func (n *NewPayloadRequest) HasValidVersionedAndBlockHashes() error {
...
-			ParentBeaconRoot: (*gethprimitives.ExecutionHash)(n.ParentBeaconBlockRoot),
-		},
-	).WithBody(gethprimitives.Body{
-		Transactions: txs, Uncles: nil, Withdrawals: *(*gethprimitives.Withdrawals)(unsafe.Pointer(&wds)),
-	}); common.ExecutionHash(block.Hash()) != payload.GetBlockHash() {
+	if common.ExecutionHash(block.Hash()) != n.ExecutionPayload.GetBlockHash() {
 		return errors.Wrapf(engineprimitives.ErrPayloadBlockHashMismatch,
 			"%x, got %x",
-			payload.GetBlockHash(), block.Hash(),
+			block.Hash(), n.ExecutionPayload.GetBlockHash(),
```
rezbera, line 96: since we're flipping these, can we change msg to  "expected %x, got %x"

consensus-types/types/payload_requests.go
```diff
@@ -98,53 +83,78 @@ func (n *NewPayloadRequest) HasValidVersionedAndBlockHashes() error {
...
 		)
 	}
 	return nil
 }
 
+func MakeEthBlock(
```
rezbera, line 102: nice this is useful. Can we add a comment above as it's not immediately clear what the second return value is from the type without introspecting the code

consensus-types/types/payload_requests.go
abi87: Nice catch, done

consensus-types/types/payload_requests.go
abi87: Added comment, LMK if it's enough

testing/state-transition/state-transition.go
```diff
@@ -65,8 +65,13 @@ func (kvs *testKVStoreService) OpenKVStore(ctx context.Context) corestore.KVStor
...
 
-//nolint:gochecknoglobals // unexported and used only in tests
-var testStoreKey = storetypes.NewKVStoreKey("state-transition-tests")
+var (
+	//nolint:gochecknoglobals // unexported and used only in tests
+	testStoreKey = storetypes.NewKVStoreKey("state-transition-tests")
+
+	//nolint:gochecknoglobals // exported but used only in tests
+	DummyProposerAddr = []byte{0xff}
+)
```
rezbera, line 74: My pref for these is to have private variables with public Getters, not a blocker

rezbera (approved):
lgtm, 1 non-blocking nit

testing/state-transition/state-transition.go
abi87: The point of this is just to feed the mocked signature verification. If anyone modified `DummyProposerAddr` in a UT I expect that UT to fail but not others.
I am going to keep it as is, and revisit as soon as it is necessary (or we have another dev wanting the getter)

PR #2517, opened 2025-02-17
fix(errors): Ignore requests Txs + RPC Error Handlings Fixes

rezbera:
- For some reason, in the error case we were returning `req.Txs`. That is the only time it is used and from what I can tell - it's always empty. This minor change makes it more clear that we do not use req.Txs

- Our error handling from the RPC was also borked. We did not check status codes and immediately tried to unmarshal which mean that any unmarshalling errors masked the real issue which was the status code. This occurred most frequently for unauthorized errors. There was also bugs with the handling of Unauthorized Errors

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -76,6 +76,16 @@ func (c Config) GetEngine() *engineclient.Config {
 	return &c.Engine
 }
 
+// GetPayloadBuilder returns the block store configuration.
+func (c Config) GetPayloadBuilder() *builder.Config {
+	return &c.PayloadBuilder
+}
+
+// GetBlockStoreService returns the block store configuration.
+func (c Config) GetBlockStoreService() *blockstore.Config {
+	return &c.BlockStoreService
+}
+
 // GetLogger returns the logger configuration.
 func (c Config) GetLogger() *log.Config {
 	return &c.Logger
--- a/consensus/cometbft/service/encoding/encoding.go
+++ b/consensus/cometbft/service/encoding/encoding.go
@@ -39,7 +39,7 @@ func ExtractBlobsAndBlockFromRequest(
 	}
 
 	blk, err := UnmarshalBeaconBlockFromABCIRequest(
-		req,
+		req.GetTxs(),
 		beaconBlkIndex,
 		forkVersion,
 	)
@@ -48,7 +48,7 @@ func ExtractBlobsAndBlockFromRequest(
 	}
 
 	blobs, err := UnmarshalBlobSidecarsFromABCIRequest(
-		req,
+		req.GetTxs(),
 		blobSidecarsIndex,
 	)
 
@@ -58,16 +58,11 @@ func ExtractBlobsAndBlockFromRequest(
 // UnmarshalBeaconBlockFromABCIRequest extracts a beacon block from an ABCI
 // request.
 func UnmarshalBeaconBlockFromABCIRequest(
-	req ABCIRequest,
+	txs [][]byte,
 	bzIndex uint,
 	forkVersion common.Version,
 ) (*ctypes.SignedBeaconBlock, error) {
 	var signedBlk *ctypes.SignedBeaconBlock
-	if req == nil {
-		return signedBlk, ErrNilABCIRequest
-	}
-
-	txs := req.GetTxs()
 	lenTxs := uint(len(txs))
 
 	// Ensure there are transactions in the request and that the request is
@@ -91,15 +86,10 @@ func UnmarshalBeaconBlockFromABCIRequest(
 // UnmarshalBlobSidecarsFromABCIRequest extracts blob sidecars from an ABCI
 // request.
 func UnmarshalBlobSidecarsFromABCIRequest(
-	req ABCIRequest,
+	txs [][]byte,
 	bzIndex uint,
 ) (datypes.BlobSidecars, error) {
 	var sidecars datypes.BlobSidecars
-	if req == nil {
-		return sidecars, ErrNilABCIRequest
-	}
-
-	txs := req.GetTxs()
 	if len(txs) == 0 || bzIndex >= uint(len(txs)) {
 		return sidecars, ErrNoBlobSidecarInRequest
 	}
--- a/consensus/cometbft/service/prepare_proposal.go
+++ b/consensus/cometbft/service/prepare_proposal.go
@@ -77,7 +77,7 @@ func (s *Service) prepareProposal(
 			"time", req.Time,
 			"err", err,
 		)
-		return &cmtabci.PrepareProposalResponse{Txs: req.Txs}, nil
+		return &cmtabci.PrepareProposalResponse{Txs: [][]byte{}}, nil
 	}
 
 	return &cmtabci.PrepareProposalResponse{
--- a/execution/client/client.go
+++ b/execution/client/client.go
@@ -23,7 +23,6 @@ package client
 import (
 	"context"
 	"math/big"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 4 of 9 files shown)


execution/client/errors.go
```diff
@@ -67,15 +67,12 @@ func (s *EngineClient) handleRPCError(
...
+		return err
+	}
 	// Check for connection errors.
-	//
-	//nolint:errorlint // from prysm.
 	e, ok := err.(jsonrpc.Error)
 	if !ok {
-		if jsonrpc.IsUnauthorizedError(e) {
-			return http.ErrUnauthorized
-		}
```
rezbera, line 78: This code is borked. It attempt to use `e` as the input for `IsUnauthorizedError` which is nil given that the casting failed and hence `ok == false`. This meant that `IsUnauthorizedError` always returned false.

execution/client/ethclient/rpc/client.go
```diff
@@ -182,6 +184,16 @@ func (rpc *client) callRaw(
 		return nil, err
 	}
 
+	switch response.StatusCode {
+	case http.StatusOK:
+		// OK: just proceed (no return)
+	case http.StatusUnauthorized:
+		return nil, beaconhttp.ErrUnauthorized
```
rezbera, line 191: Rather than attempting to do string comparisons as before - which is extremely flakey as different ELs have slightly different strings, we just check the status

execution/client/errors.go
```diff
@@ -27,26 +27,13 @@ import (
...
 )
 
-// ErrUnauthenticatedConnection indicates that the connection is not
-// authenticated.
 const (
 	UnauthenticatedConnectionErrorStr = `could not verify execution chain ID as your
 	connection is not authenticated. If connecting to your execution client via HTTP, you
 	will need to set up JWT authentication...`
-
-	AuthErrMsg = "HTTP authentication to your execution client " +
```
rezbera, line 37: Unused

execution/client/errors.go
```diff
@@ -27,26 +27,13 @@ import (
...
-		"is set correctly, or use an IPC " +
-		"connection if on the same machine."
 )
 
 var (
-	// ErrNotStarted indicates that the execution client is not started.
-	ErrNotStarted = errors.New("engine client is not started")
-
-	// ErrFailedToRefreshJWT indicates that the JWT could not be refreshed.
-	ErrFailedToRefreshJWT = errors.New("failed to refresh auth token")
```
rezbera, line 49: Unusued

consensus/cometbft/service/prepare_proposal.go
```diff
@@ -77,7 +77,7 @@ func (s *Service) prepareProposal(
 			"time", req.Time,
 			"err", err,
 		)
-		return &cmtabci.PrepareProposalResponse{Txs: req.Txs}, nil
+		return &cmtabci.PrepareProposalResponse{Txs: [][]byte{}}, nil
```
rezbera, line 80: We should be explicit in the fail case here rather than what we were doing before

execution/client/client.go
```diff
@@ -167,7 +167,7 @@ func (s *EngineClient) verifyChainIDAndConnection(
 	// After the initial dial, check to make sure the chain ID is correct.
 	chainID, err = s.Client.ChainID(ctx)
 	if err != nil {
-		if strings.Contains(err.Error(), "401 Unauthorized") {
```
rezbera, line 170: string comparisons is bad

primitives/net/json-rpc/errors.go
```diff
@@ -88,21 +86,3 @@ var (
...
-func IsUnauthorizedError(err error) bool {
-	if err == nil {
-		return false
-	}
-
-	//nolint:errorlint // its'ok.
-	e, ok := err.(Error)
-	if !ok {
-		if strings.Contains(
-			e.Error(), "401 Unauthorized",
```
rezbera, line 102: string comparisons is bad

execution/client/errors.go
```diff
@@ -67,15 +54,14 @@ func (s *EngineClient) handleRPCError(
...
 	// Check for connection errors.
-	//
-	//nolint:errorlint // from prysm.
-	e, ok := err.(jsonrpc.Error)
-	if !ok {
-		if jsonrpc.IsUnauthorizedError(e) {
-			return http.ErrUnauthorized
-		}
+	var e jsonrpc.Error
+	ok := errors.As(err, &e)
```
rezbera, line 63: prysm uses errors.As now as well
https://github.com/prysmaticlabs/prysm/blob/3eec5a5cb6095591e25ff3c08fabd513eb465434/beacon-chain/execution/engine_client.go#L706

primitives/net/http/errors.go
```diff
@@ -44,10 +44,7 @@ type TimeoutError interface {
 // status.
 // Returns true if the error is a timeout error, false otherwise.
 func IsTimeoutError(e error) bool {
-	if e == nil {
-		return false
-	}
-	//nolint:errorlint // by design.
-	t, ok := e.(TimeoutError)
+	var t TimeoutError
```
rezbera, line 47: https://github.com/prysmaticlabs/prysm/blob/3eec5a5cb6095591e25ff3c08fabd513eb465434/beacon-chain/execution/engine_client.go#L768-L772

same as new prysm impl

primitives/net/json-rpc/errors.go
abi87: It seems again a copy of Prysm

execution/client/client.go
rezbera: Prysm uses the the [go-ethereum RPC client](https://github.com/ethereum/go-ethereum/blob/99bbbc0277e34fc3a31512a345ba20874ae98e18/rpc/http.go#L233-L245), whereas we have a custom RPC client. 

Honestly, it's not clear to me why they don't check against the [Geth HTTPError](https://github.com/ethereum/go-ethereum/blob/99bbbc0277e34fc3a31512a345ba20874ae98e18/rpc/http.go#L240). Might just not be a priority.

config/config.go
```diff
@@ -76,6 +76,16 @@ func (c Config) GetEngine() *engineclient.Config {
...
+// GetPayloadBuilder returns the block store configuration.
+func (c Config) GetPayloadBuilder() *builder.Config {
+	return &c.PayloadBuilder
+}
+
+// GetBlockStoreService returns the block store configuration.
+func (c Config) GetBlockStoreService() *blockstore.Config {
+	return &c.BlockStoreService
+}
+
```
rezbera, line 88: Added these for consistency with other Getters - not directly related to the goal of this PR but useful in a future one

consensus/cometbft/service/encoding/encoding.go
```diff
@@ -58,16 +58,11 @@ func ExtractBlobsAndBlockFromRequest(
 // UnmarshalBeaconBlockFromABCIRequest extracts a beacon block from an ABCI
 // request.
 func UnmarshalBeaconBlockFromABCIRequest(
-	req ABCIRequest,
+	txs [][]byte,
```
rezbera, line 61: This is also a tangential change not relevant to this PR. However it makes this function easier to test.
It also removes the need for duplicate nil checks on `req`

execution/client/client.go
abi87: My guess is we were busy introducing generics and event driven garbage

rezbera: > My guess is we were busy introducing generics and event driven garbage

was referring to prysm team skill issue not us 😄

consensus/cometbft/service/prepare_proposal.go
abi87: I think strictly speaking this is a different behaviour, but it does not matter in our case. IIUC:

- PrepareProposal passes to the app the transactions CometBFT knows about, that are available in CometBFT mempool (our CometBFT mempool is set to zero, so no txs)
- The app is free to reorder, replace and add any transaction it wants. CometBFT will use the transactions returned in  `PrepareProposalResponse` to prepare the block
- now my guess is that default behaviour for the app, in case of errors is to let CometBFT build a block with whatever txs it has (mempool txs are verified via CheckTx anyhow)

It does not currently matter for us, so I am not opposed to this change.
I would only ask @sbudella-gco to confirm my story

abi87 (approved):
LGTM pending confirmation that returning no txs makes no difference for us currently.

fridrik01 (approved):
lgtm

PR #2518, opened 2025-02-17
chore(state-transition): cleaned up state in state-transition UTs

abi87:
Made state handling closer to what we have in production.
I need this to keep building unit tests in both state-transition (see ongoing https://github.com/berachain/beacon-kit/pull/2516) or other packages (validators APIs, not yet pushed)

### DIFF
--- a/testing/state-transition/state-transition.go
+++ b/testing/state-transition/state-transition.go
@@ -65,14 +65,15 @@ func (kvs *testKVStoreService) OpenKVStore(ctx context.Context) corestore.KVStor
 	return storage.NewKVStore(store)
 }
 
-//nolint:gochecknoglobals // unexported and use only in tests
+//nolint:gochecknoglobals // unexported and used only in tests
 var testStoreKey = storetypes.NewKVStoreKey("state-transition-tests")
 
 func BuildTestStores() (
 	storetypes.CommitMultiStore,
 	*beacondb.KVStore,
 	*depositstore.KVStore,
-	error) {
+	error,
+) {
 	db, err := db.OpenDB("", dbm.MemDBBackend)
 	if err != nil {
 		return nil, nil, nil, fmt.Errorf("failed opening mem db: %w", err)
@@ -122,8 +123,8 @@ func SetupTestState(t *testing.T, cs chain.Spec) (
 	cms, kvStore, depositStore, err := BuildTestStores()
 	require.NoError(t, err)
 
-	ms := sdk.NewContext(cms.CacheMultiStore(), true, log.NewNopLogger())
-	beaconState := statedb.NewBeaconStateFromDB(kvStore.WithContext(ms), cs)
+	sdkCtx := sdk.NewContext(cms.CacheMultiStore(), true, log.NewNopLogger())
+	beaconState := statedb.NewBeaconStateFromDB(kvStore.WithContext(sdkCtx), cs)
 
 	sp := core.NewStateProcessor(
 		noop.NewLogger[any](),
@@ -138,7 +139,7 @@ func SetupTestState(t *testing.T, cs chain.Spec) (
 	)
 
 	ctx := transition.NewTransitionCtx(
-		ms,
+		sdkCtx,
 		0, // time
 		dummyProposerAddr,
 	).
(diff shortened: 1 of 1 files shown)


testing/state-transition/state-transition.go
```diff
@@ -58,23 +58,24 @@ type (
 	TestStateProcessorT          = core.StateProcessor
 )
 
-type testKVStoreService struct {
-	ctx sdk.Context
```
abi87, line 62: this was a first attempt at creating state but it is wrong.
We should use the context provided by the transactions and extract the state frorm there, not use and reuse a single context.
It worked so far because we never handled rejected blocks or multiple blocks for the same height, which we need in this test.

testing/state-transition/state-transition.go
```diff
@@ -135,7 +138,7 @@ func SetupTestState(t *testing.T, cs chain.Spec) (
 	)
 
 	ctx := transition.NewTransitionCtx(
-		context.Background(),
+		ms,
```
abi87, line 141: this is important because the state passed to `state_processor` is extract from an `sdk.Context`. If we use any ordinary `context.Context` tests would panic.

testing/state-transition/state-transition.go
```diff
@@ -118,9 +119,11 @@ func SetupTestState(t *testing.T, cs chain.Spec) (
 
 	dummyProposerAddr := []byte{0xff}
 
-	kvStore, depositStore, err := initTestStores()
+	cms, kvStore, depositStore, err := BuildTestStores()
 	require.NoError(t, err)
-	beaconState := statedb.NewBeaconStateFromDB(kvStore, cs)
+
+	ms := sdk.NewContext(cms.CacheMultiStore(), true, log.NewNopLogger())
+	beaconState := statedb.NewBeaconStateFromDB(kvStore.WithContext(ms), cs)
```
abi87, line 126: this beaconState is basically what `storage.StateFromContext` does.
I should consider refactoring it out to make it clear

testing/state-transition/state-transition.go
```diff
@@ -58,23 +58,24 @@ type (
...
+func (kvs *testKVStoreService) OpenKVStore(ctx context.Context) corestore.KVStore {
+	store := sdk.UnwrapSDKContext(ctx).KVStore(testStoreKey)
 	return storage.NewKVStore(store)
 }
 
 //nolint:gochecknoglobals // unexported and use only in tests
 var testStoreKey = storetypes.NewKVStoreKey("state-transition-tests")
 
-func initTestStores() (*beacondb.KVStore, *depositstore.KVStore, error) {
+func BuildTestStores() (
```
abi87, line 71: pre-emptively exporting `BuildTestStores` which I am going to use in other packages

testing/state-transition/state-transition.go
```diff
@@ -58,23 +58,25 @@ type (
...
-	//nolint:contextcheck // fine with tests
-	store := sdk.UnwrapSDKContext(kvs.ctx).KVStore(testStoreKey)
+func (kvs *testKVStoreService) OpenKVStore(ctx context.Context) corestore.KVStore {
+	store := sdk.UnwrapSDKContext(ctx).KVStore(testStoreKey)
 	return storage.NewKVStore(store)
 }
 
-//nolint:gochecknoglobals // unexported and use only in tests
+//nolint:gochecknoglobals // unexported and used only in tests
 var testStoreKey = storetypes.NewKVStoreKey("state-transition-tests")
```
rezbera, line 69: What is the purpose of this special key? iirc the key by default in prod is `beacond`

abi87: I could have reused the prod one, but when I first built these tests I wanted to make sure I fully controlled them and I could use a different one.
One interesting point here is that Comet requires this very key to work: if you copy it, it won't be able to access the state with the copy.
This is something I would not have found out if I had not tried to use a test key

rezbera (approved):
doesn't hit the critical path so nothing contentious here. bit tricky to see the end-goal though

abi87:
> doesn't hit the critical path so nothing contentious here. bit tricky to see the end-goal though

I agree. I made a single PR out of this just to make tests follow prod in the way state is handled (although not showing it by other tests, one has to take my word or dig into prod).

I hope to come up soon with tests where multiple, independent states must be created on top of a finalized state, in which case I really need this PR and the caches it introduces, to make the tests independent.

PR #2519, opened 2025-02-17
chore(api): Add UTs for validators APIa

abi87:


node-api/handlers/beacon/types/conversions.go
```diff
@@ -73,3 +75,45 @@ func ValidatorFromConsensus(v *ctypes.Validator) *Validator {
 		WithdrawableEpoch:          v.GetWithdrawableEpoch().Base10(),
 	}
 }
+
+// useful in UTs
+func ValidatorToConsensus(v *Validator) (*ctypes.Validator, error) {
```
abi87, line 80: TODO: add property tests to ensure `ValidatorToConsensus` and `ValidatorFromConsensus` are dual

PR #2522, opened 2025-02-19
perf(blobs): don't process blobs outside of DA period

shotes:
During `FinalizeBlock`, a node does not need to process a blob sidecar at all unless it is within the DA period from the tip of the chain.

This is a required step towards making blobs fully ephemeral on all nodes. It also slightly improves performance during sync, not having to write then delete all blobs.


beacon/blockchain/finalize_block.go
```diff
@@ -53,24 +53,35 @@ func (s *Service) FinalizeBlock(
...
 	if blk == nil {
 		s.logger.Error("SignedBeaconBlock contains nil BeaconBlock during FinalizeBlock")
 		return nil, ErrNilBlk
 	}
+
+	// STEP 2: Finalize sidecars first (block will check for sidecar availability).
+	// SyncingToHeight is always the tip of the chain both during sync and when
+	// caught up. We don't need to process sidecars unless they are within DA period.
+	//#nosec: G115 // SyncingToHeight will never be negative.
+	if s.chainSpec.WithinDAPeriod(blk.GetSlot(), math.Slot(req.SyncingToHeight)) {
```
shotes, line 66: I think this logic is correct, given that the `BeaconBlock`'s slot corresponds to the `req.Height`.

During normal consensus, they should all be equal: `blk.GetSlot() == req.Height == req.SyncingToHeight`.

During sync, we want the actual tip of the chain.

calbera (dismissed):
LGTM

calbera (commented):
utACK. removing review until tested

beacon/blockchain/finalize_block.go
```diff
@@ -137,15 +141,6 @@ func (s *Service) finalizeBeaconBlock(
 	if err != nil {
 		return nil, err
 	}
-
-	// If the blobs needed to process the block are not available, we
```
rezbera, line 141: Why we can remove this without wrapping it in a `WithinDAPeriod` check?

calbera: It’s now above in line 78

rezbera: 🤦

beacon/blockchain/finalize_block.go
```diff
@@ -58,29 +59,31 @@ func (s *Service) FinalizeBlock(
...
+			s.logger.Error("Failed to process blob sidecars", "error", err)
+			return nil, fmt.Errorf("failed to process blob sidecars: %w", err)
+		}
+
+		// Ensure we can access the data using the commitments from the block.
+		if !s.storageBackend.AvailabilityStore().IsDataAvailable(
+			ctx, blk.GetSlot(), blk.GetBody(),
+		) {
+			return nil, ErrDataNotAvailable
+		}
```
rezbera, line 82: previously there was the comment:
```go
It is safe to use the slot off of the beacon block since it has been verified as correct already.
```

is that still the case here?

shotes: I believe the original concern was not really a valid concern. We are taking data from the finalized block which has been verified as correct. So long as cometBFT verifies the signatures for us, then that data should be the same data that passed consensus already. Thus we can trust the slot from the beacon block.

beacon/blockchain/finalize_block.go
abi87: @shotes , looks correct to me too. The one bit I do not understand rn is:
- won't block finalization check for sidecar availability during sync as well? If so, aren't we forced to process the sidecar anyhow?

I understand the optimization here, but I am surprised no changes in the FinalizeBlock flow that do some checks and not others depending if node is syncing/in normal ops.

abi87: Also this requires that sidecars processing does not affect beaconState at all, which I believe is the case, just stating this outloud

beacon/blockchain/finalize_block.go
```diff
@@ -48,6 +48,7 @@ func (s *Service) FinalizeBlock(
 		s.logger.Error("Failed to decode block and blobs", "error", err)
 		return nil, fmt.Errorf("failed to decode block and blobs: %w", err)
 	}
+	blk := signedBlk.GetMessage()
```
abi87, line 51: nit:
replace
```
finalizeErr = s.forceSyncUponFinalize(ctx, signedBlk.GetMessage())
```
with
```
finalizeErr = s.forceSyncUponFinalize(ctx, blk)
```

beacon/blockchain/finalize_block.go
shotes: > won't block finalization check for sidecar availability during sync as well? If so, aren't we forced to process the sidecar anyhow?

In this PR, we move that check so that we only check for sidecar availability if we're in the DA period.

abi87: Oh, you're right. I was tripping

calbera:
Would say this is more of a perf than fix PR? would change the title if you agree

shotes:
> Would say this is more of a perf than fix PR? would change the title if you agree

It's a little bit of both. In the current state of the code, it is just a performance upgrade. When we eventually make blobs prunable for archival nodes, then this will become necessary, depending on the implementation.

calbera (approved):
utACK

rezbera (approved):
I'm okay with this but not a fan of merging without tests. Expectation atm is that we can add tests for this using simulation work after merging but before releasing

abi87 (approved):

PR #2521, opened 2025-02-18
fix(coverage): Fixes to Ensure Test Coverage reflects actuality (2/n)

rezbera:
**THIS PR BUILDS on https://github.com/berachain/beacon-kit/pull/2411**

Currently, our code coverage reports only the output from the unit tests. This PR does the following:
- Speed up CI by removing the previously duplicated unit test execution.
- Update the code coverage reporter to combine the coverage reported from simulation tests with the unit tests.
- Remove `testing` directory from the scope of coverage reporting.

I tried to figure out how we could include E2E tests in the coverage but it seems to be a major refactor atm. 

**This PR alone bumps coverage from 32.31% ->  56.94%**

PR #2523, opened 2025-02-19
test(simulated): Add Simulation Tests for Malicious Proposer, Crashed EL and Malicious Consensus Majority (3/n)

rezbera:
This PR builds on https://github.com/berachain/beacon-kit/pull/2521 which needs to be merged first

This PR increases test coverage by adding the following tests:

- TestFinalizeBlock_BadBlock_Errors effectively serves as a test for how a valid node would react to a malicious consensus majority agreeing to a block with an invalid EVM transaction.

- TestProcessProposal_InvalidTimestamps_Errors effectively serves as a test for how a valid node would react to a malicious proposer attempting to use a future timestamp in the block that does not match the consensus timestamp.

-  TestProcessProposal_CrashedExecutionClient_Errors effectively serves as a test for how a valid node would react to a valid block being proposed but the execution client has crashed.

it also moves the tests to different files based on logical separation


testing/simulated/chaos_test.go
```diff
@@ -0,0 +1,75 @@
+//go:build simulated
```
rezbera, line 1: test comment

testing/simulated/chaos_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+// SPDX-License-Identifier: BUSL-1.1
```
rezbera, line 3: test comment

testing/simulated/simulated_test.go
```diff
@@ -140,122 +139,89 @@ func (s *SimulatedSuite) initializeChain() {
...
-	s.Require().NoError(err)
-	s.Require().Equal(types.PROCESS_PROPOSAL_STATUS_REJECT, processResp.Status)
-
-	// Verify that the log contains the expected error message.
-	s.Require().Contains(s.LogBuffer.String(), errors.ErrInvalidPayloadStatus.Error())
-	// Note this error message may change across execution clients.
-	s.Require().Contains(s.LogBuffer.String(), "max fee per gas less than block base fee: address 0x71562b71999873DB5b286dF957af199Ec94617F7, maxFeePerGas: 10000000, baseFee: 875000000")
+// CoreLoop will iterate through the core loop `iterations` times, i.e. Propose, Process, Finalize and Commit.
+// Returns the list of proposed comet blocks.
+func (s *SimulatedSuite) CoreLoop(startHeight, iterations int64, proposer *signer.BLSSigner) []*types.PrepareProposalResponse {
```
abi87, line 183: nit: consider renaming `CoreLoop` to something like `MoveChainToHeight`

testing/simulated/chaos_test.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"time"
+
+	"github.com/berachain/beacon-kit/testing/simulated"
+	"github.com/cometbft/cometbft/abci/types"
+)
+
+// TestProcessProposal_CrashedExecutionClient_Errors effectively serves as a test for how a valid node would react to
+// a valid block being proposed but the execution client has crashed.
+func (s *SimulatedSuite) TestProcessProposal_CrashedExecutionClient_Errors() {
```
abi87, line 34: lovely example of what the framework can do

testing/simulated/chaos_test.go
```diff
@@ -0,0 +1,75 @@
...
+	s.Require().NoError(err)
+
+	// Go through 1 iteration of the core loop to bypass any startup specific edge cases such as sync head on startup.
+	proposals := s.CoreLoop(blockHeight, coreLoopIterations, blsSigner)
+	s.Require().Len(proposals, coreLoopIterations)
+
+	// Prepare a valid block proposal.
+	proposalTime := time.Now()
+	proposal, err := s.SimComet.Comet.PrepareProposal(s.Ctx, &types.PrepareProposalRequest{
+		Height:          blockHeight + coreLoopIterations,
```
abi87, line 53: nit: consider introducing `currentHeight := blockHeight + coreLoopIterations` and reusing in the test

testing/simulated/chaos_test.go
```diff
@@ -0,0 +1,75 @@
...
+		Time:            proposalTime,
+		ProposerAddress: pubkey.Address(),
+	})
+	s.Require().NoError(err)
+	s.Require().NotEmpty(proposal)
+
+	// Reset the log buffer to discard old logs we don't care about.
+	s.LogBuffer.Reset()
+	// Kill the execution client.
+	err = s.ElHandle.Close()
```
abi87, line 63: nit: I believe the suite cleanup will try to close this again. Nothing wrong I think, just noting here that we should make ElHandle.Close idempotent?

testing/simulated/malicious_proposer_test.go
```diff
@@ -0,0 +1,140 @@
...
+	pubkey, err := blsSigner.GetPubKey()
+	s.Require().NoError(err)
+
+	// Go through 1 iteration of the core loop to bypass any startup specific edge cases such as sync head on startup.
+	proposals := s.CoreLoop(blockHeight, coreLoopIterations, blsSigner)
+	s.Require().Len(proposals, coreLoopIterations)
+
+	// Prepare a valid block proposal, but 2 seconds in the future (i.e. attempt to roll timestamp forward)
+	correctTime := time.Now()
+	proposalTime := correctTime.Add(2 * time.Second)
```
abi87, line 118: nit: consider using `payloadtime.Next` function here

testing/simulated/malicious_proposer_test.go
```diff
@@ -0,0 +1,140 @@
...
+	processResp, err := s.SimComet.Comet.ProcessProposal(s.Ctx, &types.ProcessProposalRequest{
+		Txs:             maliciousProposal.Txs,
+		Height:          blockHeight + coreLoopIterations,
+		ProposerAddress: pubkey.Address(),
+		// Use the correct time as the actual consensus time, which mismatches the proposal time.
+		Time: correctTime,
+	})
+	s.Require().NoError(err)
+	s.Require().Equal(types.PROCESS_PROPOSAL_STATUS_REJECT, processResp.Status)
+	s.Require().Contains(s.LogBuffer.String(), "timestamp too far in the future")
```
abi87, line 139: nit: consider using `ErrTooFarInTheFuture.Error()` here

abi87 (commented):
nits

testing/simulated/simulated_test.go
```diff
@@ -140,122 +139,89 @@ func (s *SimulatedSuite) initializeChain() {
 }
 
 // TestFullLifecycle_ValidBlock_IsSuccessful tests that a valid block proposal is processed, finalized, and committed.
```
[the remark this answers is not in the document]
rezbera: Agreed

testing/simulated/simulated_test.go
rezbera: Good call

testing/simulated/malicious_proposer_test.go
rezbera: I'm not sure if `payloadtime.Next` is suitable here. That function is useful for generating a valid block timestamp based on parent and consensus time, but my goal is to create an invalid block timestamp

testing/simulated/simulated_test.go
```diff
@@ -139,123 +135,50 @@ func (s *SimulatedSuite) initializeChain() {
 	s.Require().Len(deposits, 1, "Expected 1 deposit")
 }
 
-// TestFullLifecycle_ValidBlock_IsSuccessful tests that a valid block proposal is processed, finalized, and committed.
-func (s *SimulatedSuite) TestFullLifecycle_ValidBlock_IsSuccessful() {
```
rezbera, line 143: test moved to `valid_chain_test`

testing/simulated/malicious_proposer_test.go
abi87: Just do NexTime.Add(time.Second) and it will be more explicit why the timestamp is invalid? At least I think

rezbera: I have renamed variables to make it more clear on the intent of the test:
```go
	correctConsensusTime := time.Now()
	maliciousProposalTime := correctConsensusTime.Add(2 * time.Second)
```
The problem with using `payloadtime.Next` is that it requires `parentPayloadTimestamp math.U64`, which I currently don't have access to without unmarshalling the last proposal. Theoretically we could store and retrieve the parent payload time though but the benefit seems small

.gitignore
```diff
@@ -226,8 +226,11 @@ out
 # docker test
 **/temp/
 
-# integration testing coverage
+# testing coverage
 coverage-test-unit-cover.txt
+coverage-merged.txt
+test-simulated.txt
+test-unit-cover.txt
```
rezbera, line 233: Carry over from https://github.com/berachain/beacon-kit/pull/2521 review

testing/simulated/chaos_test.go
rezbera: Correct that it is torndown at the end of the suite. Currently it logs, but nothing else. It's hacky so I will clean it up. 
```go
	if err := s.ElHandle.Close(); err != nil {
		s.T().Logf("Error closing EL handle: %s", err)
	}
```

rezbera: Addressed in https://github.com/berachain/beacon-kit/pull/2523/commits/d09555afc8918a94375cba9cde63e78d1b540cda

testing/simulated/execution/resource.go
```diff
@@ -0,0 +1,44 @@
...
+
+import "github.com/ory/dockertest"
+
+// Resource manages a *dockertest.Resource with added utility.
+type Resource struct {
+	*dockertest.Resource
+	isClosed bool
+}
+
+func (r *Resource) Close() error {
```
calbera, line 33: q: why do we need two functions for Close() ? seems it can just be something like 

```go
func (r *Resource) Close() error {
        if r.isClosed {
		return nil
	}

        r.isClosed = true
	return r.Resource.Close()
}
```

testing/simulated/execution/resource.go
```diff
@@ -0,0 +1,44 @@
...
+// EXPRESS OR IMPLIED, INCLUDING (WITHOUT LIMITATION) WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, NON-INFRINGEMENT, AND
+// TITLE.
+
+package execution
+
+import "github.com/ory/dockertest"
+
+// Resource manages a *dockertest.Resource with added utility.
+type Resource struct {
```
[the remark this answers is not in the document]
rezbera: Why would it need to be thread-safe? Every test manages its own resource separately

testing/simulated/execution/resource.go
rezbera: true. fixed

abi87 (approved):
LGTM

PR #2526, opened 2025-02-20
chore(payload): allow `payloadID` to get stale

abi87:


payload/builder/payload.go
```diff
@@ -194,48 +194,6 @@ func (pb *PayloadBuilder) RetrievePayload(
...
-// TODO: This should be moved onto a "sync service"
-// of some kind.
-func (pb *PayloadBuilder) SendForceHeadFCU(
-	ctx context.Context,
-	st *statedb.StateDB,
-	slot math.Slot,
-) error {
-	if !pb.Enabled() {
-		return ErrPayloadBuilderDisabled
-	}
```
abi87, line 209: I think this is a bug. We should send FCU at start even if the node is not building blocks!
I believe I introduced this and I think I got tricked by this method being in PayloadBuilder.

PR #2528, opened 2025-02-20
fix(testnet): Activate Deneb1 for EVM inflation

calbera:
## Previously
- EVM inflation turned on at genesis.
- Deneb1 was never set to turn on, but the values were set.

## New Behavior
- To keep fork versions consistent with mainnet, we shall activate Deneb1 at epoch 1 (slot 192). Genesis fork version will remain as Deneb, as is on mainnet.
- EVM inflation can be kept on or off at genesis. As BGT contracts and minting will not be live until well after block 192, choosing to keep it off at genesis (prevents minting more BERA than we need).

### DIFF
--- a/config/spec/testnet.go
+++ b/config/spec/testnet.go
@@ -30,12 +30,12 @@ func TestnetChainSpecData() *chain.SpecData {
 	specData := MainnetChainSpecData()
 	specData.DepositEth1ChainID = TestnetEth1ChainID
 
-	// Genesis values of inflation are unused since Deneb1 fork is activated at block 1 (epoch 0).
+	// Genesis values of EVM inflation are consistent with Deneb1 to keep BERA minting on.
 	specData.EVMInflationAddressGenesis = common.NewExecutionAddressFromHex(mainnetEVMInflationAddressDeneb1)
 	specData.EVMInflationPerBlockGenesis = mainnetEVMInflationPerBlockDeneb1
 
-	// Unlike mainnet, testnet activates Deneb1 (Bera minting) at block 1 (epoch 0).
-	specData.Deneb1ForkEpoch = 0
+	// Unlike mainnet, testnet activates Deneb1 at epoch 1.
+	specData.Deneb1ForkEpoch = 1
 	specData.EVMInflationAddressDeneb1 = common.NewExecutionAddressFromHex(mainnetEVMInflationAddressDeneb1)
 	specData.EVMInflationPerBlockDeneb1 = mainnetEVMInflationPerBlockDeneb1
 
--- a/config/spec/testnet.go
+++ b/config/spec/testnet.go
@@ -26,18 +26,19 @@ import (
 )
 
 // TestnetChainSpecData is the chain.SpecData for Berachain's public testnet.
-//
-// TODO: adjust values before testnet genesis.
 func TestnetChainSpecData() *chain.SpecData {
 	specData := MainnetChainSpecData()
 	specData.DepositEth1ChainID = TestnetEth1ChainID
 
-	// Unlike mainnet, testnet activates Bera minting at block 1
-	specData.EVMInflationAddressDeneb1 = common.NewExecutionAddressFromHex(mainnetEVMInflationAddressDeneb1)
+	// Genesis values of inflation are unused since Deneb1 fork is activated at block 1 (epoch 0).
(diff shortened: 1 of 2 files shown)


config/spec/testnet.go
```diff
@@ -26,18 +26,19 @@ import (
...
 
-	// Unlike mainnet, testnet activates Bera minting at block 1
-	specData.EVMInflationAddressDeneb1 = common.NewExecutionAddressFromHex(mainnetEVMInflationAddressDeneb1)
+	// Genesis values of inflation are unused since Deneb1 fork is activated at block 1 (epoch 0).
 	specData.EVMInflationAddressGenesis = common.NewExecutionAddressFromHex(mainnetEVMInflationAddressDeneb1)
+	specData.EVMInflationPerBlockGenesis = mainnetEVMInflationPerBlockDeneb1
 
+	// Unlike mainnet, testnet activates Deneb1 (Bera minting) at block 1 (epoch 0).
+	specData.Deneb1ForkEpoch = 0
+	specData.EVMInflationAddressDeneb1 = common.NewExecutionAddressFromHex(mainnetEVMInflationAddressDeneb1)
```
rezbera, line 39: Why do we even override this line? its the same as `MainnetChainSpecData`

config/spec/testnet.go
```diff
@@ -26,18 +26,19 @@ import (
...
 	specData.DepositEth1ChainID = TestnetEth1ChainID
 
-	// Unlike mainnet, testnet activates Bera minting at block 1
-	specData.EVMInflationAddressDeneb1 = common.NewExecutionAddressFromHex(mainnetEVMInflationAddressDeneb1)
+	// Genesis values of EVM inflation are consistent with Deneb1 to keep BERA minting on.
 	specData.EVMInflationAddressGenesis = common.NewExecutionAddressFromHex(mainnetEVMInflationAddressDeneb1)
+	specData.EVMInflationPerBlockGenesis = mainnetEVMInflationPerBlockDeneb1
 
+	// Unlike mainnet, testnet activates Deneb1 at epoch 1.
+	specData.Deneb1ForkEpoch = 1
```
rezbera, line 38: We're 100% that epochs are zero-indexed? i.e. Genesis is Epoch 0

calbera: yes indeed

config/spec/testnet.go
rezbera: idk i found it more confusing as to why it was being set

config/spec/testnet.go
calbera: check the logic in `ActiveForkVersionFor(Slot/Epoch)` functions to convince yourself.

rezbera: convinced

config/spec/testnet.go
calbera: So now, the only thing thats different from mainnet is `Deneb1ForkEpoch = 1`. Should that be the only one we explicitly do?

rezbera: Only differences should be explicit

calbera: done

config/spec/testnet.go
```diff
@@ -20,24 +20,19 @@
...
 
-	// Unlike mainnet, testnet activates Bera minting at block 1
-	specData.EVMInflationAddressDeneb1 = common.NewExecutionAddressFromHex(mainnetEVMInflationAddressDeneb1)
-	specData.EVMInflationAddressGenesis = common.NewExecutionAddressFromHex(mainnetEVMInflationAddressDeneb1)
+	// Genesis values of EVM inflation are consistent with those of mainnet.
+	// Testnet activates Deneb1 for BERA minting at epoch 1.
+	specData.Deneb1ForkEpoch = 1
 
-	specData.EVMInflationPerBlockDeneb1 = mainnetEVMInflationPerBlockDeneb1
-	specData.EVMInflationPerBlockGenesis = mainnetEVMInflationPerBlockDeneb1
```
calbera, line 40: as discussed with Rez, removing assignment of values that are already consistent with mainnet spec. Allows the differences to be clearly highlighted here (which are only _when_ the fork epochs are occurring).

rezbera (approved):
Agree with the intent of his PR. Before his PR we were just setting the genesisInflation  to that of the deneb1 values without actually activating the deneb1 hardfork. 
That to me is a bug.

PR #2531, opened 2025-02-21
[DRAFT] Sidecars SSZ

rezbera:


da/types/sidecars_test.go
```diff
@@ -88,6 +88,53 @@ func TestEmptySidecarMarshalling(t *testing.T) {
...
+	// Unmarshal the empty sidecar
+	unmarshalled := types.BlobSidecars{}
+	err = unmarshalled.UnmarshalSSZ(marshalled)
+	require.NoError(
+		t,
+		err,
+		"Unmarshalling empty sidecars should not produce an error",
+	)
+
+	// Compare the original and unmarshalled empty sidecars
```
rezbera, line 129: These will not be equal

PR #2534, opened 2025-02-23
test(simulated): test(simulated): Simulated Blob Tests (5/n)

rezbera:
It introduced valid and byzantine proposer tests for blobs


consensus/cometbft/service/encoding/encoding.go
```diff
@@ -102,5 +102,9 @@ func UnmarshalBlobSidecarsFromABCIRequest(
 	// TODO: Do some research to figure out how to make this more
 	// elegant.
 	sidecars = datypes.BlobSidecars{}
-	return sidecars, sidecars.UnmarshalSSZ(sidecarBz)
+	err := sidecars.UnmarshalSSZ(sidecarBz)
```
rezbera, line 105: This is just cleaner imo

PR #2539, opened 2025-02-25
fix(node-api): Fix param in validator endpoint

shotes:
Makes the validator by id API endpoint work. Example test with `curl -X 'GET' \
  'http://localhost:3500/eth/v1/beacon/states/head/validators/0' \
  -H 'accept: application/json'`


calbera (approved):
utACK

abi87 (approved):

PR #2545, opened 2025-02-27
fix(block building): let payloadID cache be purged

abi87:


shotes:
This solution will require the node to fail once before purging the payload ID from the cache.

Here is an alternative solution (https://github.com/berachain/beacon-kit/pull/2547) that immediately purges the payload ID from the cache when retrieved. This makes it a single-use ID, which is how reth treats it. This also makes it so that beacon-kit does not need to fail a request once before purging the payload ID from its cache.

abi87:
https://github.com/berachain/beacon-kit/pull/2547 is a cleaner solution I think

PR #2546, opened 2025-02-27
chore(config): Disable optimistic builds by default

shotes:
This disables the optimistic builder for beacon-kit by default. The nodes run by infra do not run optimistic builds, and the nodes on the public testnet also do not run optimistic builds.

I believe that we should disable them as their default behavior.


testing/networks/80084/app.toml
calbera, line 1: have we made sure that this is set to false in all other networks' app.tomls?

shotes: checked to make sure it is set to false in all of the places that showed up when I grep for "optimistic". All of the `app.toml`s in `testing/networks/*` are set to `false`.

calbera (approved):
LGTM, good default to have optimistic building OFF

shotes:
merge if we reach consensus that this is the path forward.

rezbera (approved):

abi87:
I am not sure we should disable this everywhere.
I understand the principle of having our tests exercise code path as close to prod as possible BUT
shouldn't we extend tests of optimistic payload building and then keep it for whenever we need performances?
Or activate it in devnet/testnet for example?

rezbera:
> I am not sure we should disable this everywhere. I understand the principle of having our tests exercise code path as close to prod as possible BUT shouldn't we extend tests of optimistic payload building and then keep it for whenever we need performances? Or activate it in devnet/testnet for example?

My take at the moment is that we're not in a position yet where we have the necessary test coverage to cover for both test paths (feel free to correct me if I'm wrong). Whichever mode (i.e. optimistic or non-optimistic) we have more coverage for, we should focus on that and build up coverage till we have capacity to support both modes.

PR #2547, opened 2025-02-27
fix(builder): Flush PayloadID cache

shotes:
This is an alternative solution to https://github.com/berachain/beacon-kit/pull/2545

This PR introduces a `delete` on the item that is fetched from the payload ID cache. When it is fetched, it now gets deleted.

This fixes a problem that can be replicated by running `make start` and `make start-reth` locally with the following diff:
```
diff --git a/payload/builder/payload.go b/payload/builder/payload.go
index fad226702..9d98638fa 100644
--- a/payload/builder/payload.go
+++ b/payload/builder/payload.go
@@ -90,6 +90,8 @@ func (pb *PayloadBuilder) RequestPayloadAsync(
 	return payloadID, nil
 }

+var RUN_ONCE = true
+
 // RequestPayloadSync request a payload for the given slot and
 // blocks until the payload is delivered.
 func (pb *PayloadBuilder) RequestPayloadSync(
@@ -122,14 +124,20 @@ func (pb *PayloadBuilder) RequestPayloadSync(
 	if payloadID == nil {
 		return nil, ErrNilPayloadID
 	}
+	payloadTimeout := pb.cfg.PayloadTimeout
+	if slot == 10 && RUN_ONCE {
+		RUN_ONCE = false
+		payloadTimeout = time.Second * 10
+	}

 	// Wait for the payload to be delivered to the execution client.
 	pb.logger.Info(
 		"Waiting for local payload to be delivered to execution client",
-		"for_slot", slot.Base10(), "timeout", pb.cfg.PayloadTimeout.String(),
+		"for_slot", slot.Base10(), "timeout", payloadTimeout.String(),
+		"payloadID", payloadID.String(),
 	)
 	select {
-	case <-time.After(pb.cfg.PayloadTimeout):
+	case <-time.After(payloadTimeout):
 		// We want to trigger delivery of the payload to the execution client
 		// before the timestamp expires.
 		break
```

NOTE: When testing **disable optimistic building**. Otherwise this diff will do nothing. You can do this by modifying the `app.toml` that your `make start` uses. Mine is in `.tmp/beacond/config/app.toml`. Set `enable-optimistic-payload-builds` to `"false"`. When you start with `make start` again, make sure not to overwrite the config. Otherwise it will re-enable optimistic builds by default.

This diff will introduce a 10 second wait at slot 10. When run with `make start` and `make start-reth`, this will result in:
1. untimely prepare proposal at slot 10, but successful engine API calls to trigger build via `engine_forkchoiceUpdated` and retrieve build via `engine_getPayload`.
2. re-prepare proposal for slot 10
3. beacon-kit node finds payload ID in cache, uses this ID in another `engine_getPayload` request
4. reth has already made that payload stale after the first request, returning `Error -38001: unknown payload`.
5. repeat this round after round in slot 10 indefinitely.

With the fix, the payload ID is removed from the beacon-kit cache after the first `engine_getPayload` call. Thus subsequent calls on secondary rounds will result in new payload builds getting started instead of attempting to use the same stale build.


abi87 (commented):
I like this solution much better, to the point I am going to close the other PR I made.
I am not sure I agree with switching optimistic build in tests.
Probably we have a majority pushing for it in core but I would still keep it out of this PR and then tackle that question separately

abi87:
Actually there already is a PR handling optimistic payload configs https://github.com/berachain/beacon-kit/pull/2546.
Please remove that from this PR and rebase this on main

shotes:
> Actually there already is a PR handling optimistic payload configs #2546.
> Please remove that from this PR and rebase this on main

Right, I had branched this PR off of that optimistic payload configs branch. Had intended to set this PR to merge into the other one, but didn't set up the PR right.

But yea, I can rebase this PR onto main so that it doesn't include the optimistic payload configs branch stuf.

payload/cache/payload_id.go
```diff
@@ -68,21 +68,22 @@ func (p *PayloadIDCache[RootT, SlotT]) Has(
...
 ) (engineprimitives.PayloadID, bool) {
-	p.mu.RLock()
-	defer p.mu.RUnlock()
-	innerMap, ok := p.slotToStateRootToPayloadID[slot]
-	if !ok {
-		return engineprimitives.PayloadID{}, false
-	}
-	pid, ok := innerMap[stateRoot]
-	if !ok {
+	if !p.Has(slot, stateRoot) {
```
fridrik01, line 77: We should use the same lock operation for checking if it exists in cache and evicting from the cache as there is a potential data race makin

fridrik01 (commented):
I like it, very simple, but there is a data race

payload/cache/payload_id.go
shotes: oh duh, good catch!

payload/cache/payload_id.go
```diff
@@ -35,7 +35,7 @@ const historicalPayloadIDCacheSize = 2
 // on slot and parent block hash. It is designed to improve the efficiency of
 // payload ID retrieval by caching recent entries.
 type PayloadIDCache[
-	RootT ~[32]byte, SlotT ~uint64,
+RootT ~[32]byte, SlotT ~uint64,
```
abi87, line 38: nit: unrelated to the PR, but should we drop the generic too?

payload/cache/payload_id.go
```diff
@@ -83,6 +84,9 @@ func (p *PayloadIDCache[RootT, SlotT]) Get(
 	if !ok {
 		return engineprimitives.PayloadID{}, false
 	}
+
+	// Successfully retrieved. Remove from cache.
+	delete(p.slotToStateRootToPayloadID[slot], stateRoot)
```
abi87, line 89: another nit, possibly unrelated: go should allow using structs as keys, so should we turn this map of maps into a map of `struct{slot, root}`?

payload/cache/payload_id.go
shotes: Agreed. Also, for some reason, my IDE auto corrects the formatting to this, even though it's wrong. 

Will remove generics too.

payload/cache/payload_id.go
abi87: we could keep the struct only internal and keep the cache API the same (passing slot and root separately is fine)

payload/cache/payload_id.go
```diff
@@ -34,63 +36,67 @@ const historicalPayloadIDCacheSize = 2
...
 // payload ID retrieval by caching recent entries.
-type PayloadIDCache[
-	RootT ~[32]byte, SlotT ~uint64,
-] struct {
+type PayloadIDCache struct {
 	// mu protects access to the slotToStateRootToPayloadID map.
 	mu sync.RWMutex
 	// slotToStateRootToPayloadID is used for storing payload ID mappings
-	slotToStateRootToPayloadID map[SlotT]map[RootT]engineprimitives.PayloadID
+	slotToStateRootToPayloadID map[PayloadIDCacheKey]engineprimitives.PayloadID
```
calbera, line 43: not a big deal, but what does golang do internally when hashing a struct? shouldn't be too expensive but otherwise can just create the key manually as func of (slot, root)

payload/cache/payload_id.go
```diff
@@ -34,63 +36,67 @@ const historicalPayloadIDCacheSize = 2
...
 	return ok
 }
 
-// Get was successful.
-func (p *PayloadIDCache[RootT, SlotT]) Get(
-	slot SlotT,
-	stateRoot RootT,
+// Get retrieves the payloadID from the cache. If successfully retrieved,
+// evict it from the cache.
+func (p *PayloadIDCache) Get(
```
calbera, line 78: note: nomenclature with `Get` function usually implies no writes, but here we are performing deletion. 

From claude lol:

> How about retrieveAndEvict? This name clearly indicates both actions: it retrieves the item from the cache and evicts it when found. Other good options:
> - popFromCache - Similar to how pop works in data structures
> - extractFromCache - Implies removing it from its source
> - fetchAndRemove - Direct and descriptive
> - getAndEvict - Simple and straightforward

payload/builder/types.go
```diff
@@ -31,11 +31,11 @@ import (
 	statedb "github.com/berachain/beacon-kit/state-transition/core/state"
 )
 
-type PayloadCache[RootT, SlotT any] interface {
-	Get(slot SlotT, stateRoot RootT) (engineprimitives.PayloadID, bool)
-	Has(slot SlotT, stateRoot RootT) bool
-	Set(slot SlotT, stateRoot RootT, pid engineprimitives.PayloadID)
-	UnsafePrunePrior(slot SlotT)
+type PayloadCache interface {
+	GetAndEvict(slot math.Slot, stateRoot common.Root) (engineprimitives.PayloadID, bool)
```
abi87, line 35: Usually I don't naming naming things `SomethingANDSomething` but here it seems appropriate?
- Evict only would be misleading as this is really a get that consumes the retrieved payload
- Get would not convey the fact that retrieved item is consumed
- `Pop` comes to mind? Maybe `Push` and `Pop` instead of `Set/Get`?

I think we will end up having this cache removed and having a single payloadID "pending". So GetAndEvict is totally fine for me now

payload/cache/payload_id.go
```diff
@@ -34,63 +36,67 @@ const historicalPayloadIDCacheSize = 2
...
 	// mu protects access to the slotToStateRootToPayloadID map.
 	mu sync.RWMutex
 	// slotToStateRootToPayloadID is used for storing payload ID mappings
-	slotToStateRootToPayloadID map[SlotT]map[RootT]engineprimitives.PayloadID
+	slotToStateRootToPayloadID map[PayloadIDCacheKey]engineprimitives.PayloadID
+}
+
+// PayloadIDCacheKey is the (slot, root) tuple that is used to access a
+// payloadID from the cache.
+type PayloadIDCacheKey struct {
```
abi87, line 48: @shotes can this be un exported?

abi87: @shotes my other question is: why we need both slot and parentBlockHash?

This seems to indicate we may have multiple payload being built for the same slot and multiple parents, which maybe can happen in case of optimistic payload + network partitions (multiple valid blocks are them verified and we build optimistically on top of all of them).
Just checking

payload/cache/payload_id.go
```diff
@@ -101,18 +107,13 @@ func (p *PayloadIDCache[RootT, SlotT]) Set(
...
-	}
-	innerMap[stateRoot] = pid
+	p.slotToStateRootToPayloadID[PayloadIDCacheKey{slot, stateRoot}] = pid
 }
 
 // UnsafePrunePrior removes payload IDs from the cache for slots less than
 // the specified slot. Only used for testing.
-func (p *PayloadIDCache[_, SlotT]) UnsafePrunePrior(
-	slot SlotT,
+func (p *PayloadIDCache) UnsafePrunePrior(
```
abi87, line 115: @shotes can this be unexported or removed?
I don't see where it is used in our prod code

payload/cache/payload_id.go
```diff
@@ -34,63 +36,67 @@ const historicalPayloadIDCacheSize = 2
...
 
-// Get was successful.
-func (p *PayloadIDCache[RootT, SlotT]) Get(
-	slot SlotT,
-	stateRoot RootT,
+// GetAndEvict retrieves the payloadID from the cache. If successfully retrieved,
+// evict it from the cache.
+func (p *PayloadIDCache) GetAndEvict(
+	slot math.Slot,
+	stateRoot common.Root,
```
rezbera, line 80: I see references to `stateRoot` but what's actually passed in is the `ParentBlockRoot`. State root to me implies BeaconState root. This issue wasn't introduced in your PR but could be a good cleanup here, even with functions like `slotToStateRootToPayloadID`

rezbera (commented):
Good change - 1 issue with the naming of `stateRoot`. Could we also add a test in `TestPayloadIDCache` that makes sure duplicate `GetAndEvict` correctly evicts.

payload/cache/payload_id.go
calbera: Wait this should be block root and not state root anyways. I think we neee to clean this up in this PR

rezbera: yep, name needs to change from `stateRoot` -> ``blockRoot`

payload/cache/payload_id.go
shotes: Right. The idea is that under some circumstances, we *may* build on top of a not yet finalized block.

The optimistic build is triggered in ProcessProposal. It will start a build on the proposed block (where the proposed blocks payload hash is the parent hash). This happens before ProcessProposal is finished, and before there is network consensus. Should that proposal not pass (like bad network conditions, unable to get 2/3 vote in timely manner, or worse, some clients think its valid and some others not), then we could end up in a situation where we are building a new payload for the same slot on a different parent.

I think this situation is not possible without optimistic builds though

payload/cache/payload_id.go
shotes: This is used in unit tests `TestPayloadIDCache` and `FuzzPayloadIDCacheBasic`. Currently rewriting test cases so that they still properly test the functionality of `prunePrior` without needing to add this extra `UnsafePrunePrior` function.

shotes: Removed and fixed tests.

beacon/blockchain/payload.go
```diff
@@ -214,7 +214,8 @@ func (s *Service) rebuildPayloadForRejectedBlock(
 		// We are rebuilding for the current slot.
 		stateSlot,
 		nextPayloadTimestamp.Unwrap(),
-		// We set the parent root to the previous block root.
+		// We set the parent root to the previous block root. The HashTreeRoot
```
shotes, line 217: Added clarifying comment regarding the block root calculation using the beaconBlockHeader.

beacon/blockchain/payload.go
```diff
@@ -214,7 +214,8 @@ func (s *Service) rebuildPayloadForRejectedBlock(
...
+		// of the header is the same as the HashTreeRoot of the block.
```
abi87, line 218: Awesome, thanks for checking this!

abi87: Is easily understandable from code or should we consider adding a property test to show this?

shotes: https://github.com/berachain/beacon-kit/pull/2547/commits/ac8d5d6704abe7c58c121196a90eeb69d7752a6f

abi87 (approved):
Retested successfully in a couple of scenarios:

- scenario described in the PR, i.e. optimistic-payload off, sleep for 10 secs once at slot X --> consensus skips round 0, getPayload errs at round 1, block finalized at round 2
- `builder.deadline` scenario, i.e. optimistic-payload off, comet timeouts @ 25 secs, sleep for 20 secs once at slot X --> reth immediately fail due to deadline exceeded, it fails again because we don't cleanup the cache upon `getPayload` error then it restarts again

PR #2548, opened 2025-02-28
chore(primitives): consolidated `ParseUint` usage into `U64FromString`

abi87:
We have a de facto way to turn strings into u64 but this is not consolidated in the code.
The PR removes the code duplication

### DIFF
--- a/node-api/engines/echo/validator.go
+++ b/node-api/engines/echo/validator.go
@@ -116,7 +116,7 @@ func ValidateTimestampID(fl validator.FieldLevel) bool {
 }
 
 func ValidateUint64Dec(value string) bool {
-	if value == "" { // TODO ABEAR: CHECK IF THIS CAN BE ABSORBED INTO U64FromString
+	if value == "" {
 		return true
 	}
 	_, err := math.U64FromString(value)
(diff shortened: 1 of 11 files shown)


node-api/engines/echo/validator.go
```diff
@@ -119,10 +119,8 @@ func ValidateUint64Dec(value string) bool {
 	if value == "" {
 		return true
 	}
-	if _, err := strconv.ParseUint(value, 10, 64); err == nil {
-		return true
-	}
-	return false
+	_, err := math.U64FromString(value)
```
abi87, line 122: `math.U64FromString` is based on `strconv.ParseUint` which returns ErrSyntax in case of empty string.
We could avoid the `value == ""` test above, but I would avoid pulling up third libraries errors and just explicitly do the check ourselves

node-api/handlers/utils/id.go
```diff
@@ -106,17 +105,6 @@ func IsTimestampIDPrefix(timestampID string) bool {
...
-// U64FromString returns a math.U64 from the given string. Errors if the given
-// string is not in proper decimal notation.
-func U64FromString(id string) (math.U64, error) {
-	u64, err := strconv.ParseUint(id, 10, 64)
-	if err != nil {
-		return 0, err
-	}
-
-	return math.U64(u64), nil
-}
```
abi87, line 118: consolidated in `math` package

fridrik01 (approved):
lgtm

PR #2550, opened 2025-02-28
chore(payload): clean fcu signature

abi87:
`NotifyForkchoiceUpdate` returns `latestValidHash` which is never used. This PR drops it.
Extracted from https://github.com/berachain/beacon-kit/pull/2490 to reduce diff

### DIFF
--- a/execution/client/helpers.go
+++ b/execution/client/helpers.go
@@ -47,14 +47,14 @@ func processPayloadStatusResult(
 	result *engineprimitives.PayloadStatusV1,
 ) (*common.ExecutionHash, error) {
 	switch result.Status {
+	case engineprimitives.PayloadStatusValid:
+		return result.LatestValidHash, nil
 	case engineprimitives.PayloadStatusAccepted:
 		return nil, engineerrors.ErrAcceptedPayloadStatus
 	case engineprimitives.PayloadStatusSyncing:
 		return nil, engineerrors.ErrSyncingPayloadStatus
 	case engineprimitives.PayloadStatusInvalid:
-		return result.LatestValidHash, engineerrors.ErrInvalidPayloadStatus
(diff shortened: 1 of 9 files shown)


execution/client/helpers.go
```diff
@@ -47,14 +47,14 @@ func processPayloadStatusResult(
 	result *engineprimitives.PayloadStatusV1,
 ) (*common.ExecutionHash, error) {
 	switch result.Status {
+	case engineprimitives.PayloadStatusValid:
```
abi87, line 50: just moved up the only case where we don't return errors

bar-bera (approved):

execution/client/helpers.go
```diff
@@ -47,14 +47,14 @@ func processPayloadStatusResult(
...
+		return result.LatestValidHash, nil
 	case engineprimitives.PayloadStatusAccepted:
 		return nil, engineerrors.ErrAcceptedPayloadStatus
 	case engineprimitives.PayloadStatusSyncing:
 		return nil, engineerrors.ErrSyncingPayloadStatus
 	case engineprimitives.PayloadStatusInvalid:
-		return result.LatestValidHash, engineerrors.ErrInvalidPayloadStatus
-	case engineprimitives.PayloadStatusValid:
-		return result.LatestValidHash, nil
+		return nil, engineerrors.ErrInvalidPayloadStatus
```
calbera, line 57: just noting: interesting that we were returning the LatestValidHash even if Invalid status was returned. Glad you’re getting rid of that and just returning nil

calbera (approved):
LGTM

execution/client/helpers.go
abi87: 100%

PR #2551, opened 2025-02-28
fix(payload): `forceSyncUponProcess` even when block building is disabled

abi87:
I believe we currently skip it due to a lazy `!pb.Enabled()` which I placed here for analogy with the rest of `payload` package functionalities.

`SendFCU` should not belong to `payload` package.


beacon/blockchain/payload.go
```diff
@@ -48,7 +48,35 @@ func (s *Service) forceSyncUponProcess(
 
 	// TODO: Verify if the slot number is correct here, I believe in current
 	// form it should be +1'd. Not a big deal until hardforks are in play though.
-	if err = s.localBuilder.SendForceHeadFCU(ctx, st, slot+1); err != nil {
+	slot++
+
+	lph, err := st.GetLatestExecutionPayloadHeader()
```
abi87, line 53: this is the implementation that was in `localBuilder.SendForceHeadFCU`, without the `IsEnabled`

shotes (approved):
I think this is good and puts `forceSyncUponProcess` in line with `forceSyncUponFinalize`. LGTM

PR #2554, opened 2025-03-01
feat(simulated): Add Simulation Client for EVM State Transitions (4/n)

rezbera:
The primary unlock from this PR is the ability to do tests like `TestFullLifecycle_ValidBlockWithInjectedTransaction_IsSuccessful`, in which we can inject custom transactions and create valid blocks that pass processing which was previously not possibly.

To do this, we must introduce the simulation client, which utilises `eth_simulateV1` to calculate the EVM state and receipts roots for a given block. These fields are necessary to correctly pass `NewPayload` validation. 

The `eth_simulateV1` API is currently missing features we require which I hope to have merged in upstream geth https://github.com/ethereum/go-ethereum/pull/31304

Till then, we have to use a custom image which I've built based on the linked PR. Once upstream is merged and released, we can go back to the original image.


testing/simulated/execution/geth.go
```diff
@@ -82,14 +82,15 @@ func (g *GethNode) Start(t *testing.T) (*Resource, *url.ConnectionURL) {
...
+			geth --http --http.addr 0.0.0.0 --http.api eth,net,web3,debug \
 				 --authrpc.addr 0.0.0.0 \
 				 --authrpc.jwtsecret /testing/files/jwt.hex \
 				 --authrpc.vhosts '*' \
 				 --datadir /tmp/gethdata \
 				 --ipcpath /tmp/gethdata/geth.ipc \
 				 --syncmode full \
-				 --verbosity 4
+				 --verbosity 4 \
+				 --nodiscover
```
rezbera, line 93: nodiscover as peer discovery is not required

testing/simulated/execution/simulationclient.go
calbera, line 1: nit: rename to simulation_client.go? or sim_client.go?

testing/simulated/execution/geth.go
```diff
@@ -131,3 +132,13 @@ func ValidGethImage() docker.PullImageOptions {
...
+
+// ValidGethImageWithSimulate returns the default Docker image options for the Geth node with Simulate API
+// Build references commit https://github.com/ethereum/go-ethereum/tree/2407255bb3032dc17205ba0d648270357c98b713
+// TODO: Remove once https://github.com/ethereum/go-ethereum/pull/31304/files is merged.
+func ValidGethImageWithSimulate() docker.PullImageOptions {
+	return docker.PullImageOptions{
+		Repository: "rezbera/geth-simulate",
+		Tag:        "latest",
+	}
+}
```
abi87, line 144: can we have a `berachain` fork instead of personal forks?

testing/simulated/utils.go
```diff
@@ -56,142 +62,170 @@ type SharedAccessors struct {
...
+				Calls: calls,
+				BlockOverrides: &execution.BlockOverrides{
+					Time:          &overrideTime,
+					GasLimit:      &overrideGasLimit,
+					FeeRecipient:  (*gethcommon.Address)(&overrideFeeRecipient),
+					PrevRandao:    &overridePrevRandao,
+					BaseFeePerGas: (*hexutil.Big)(overrideBaseFeePerGas),
+					BeaconRoot:    &overrideBeaconRoot,
+					Withdrawals:   overrideWithdrawals,
+					// TODO: Do we need to override blob base fee?
```
rezbera, line 105: This addressed in 5/n

testing/simulated/execution/geth.go
rezbera: done

testing/simulated/execution/simulationclient.go
rezbera: done

testing/simulated/execution/simulation_client.go
```diff
@@ -0,0 +1,199 @@
...
+	WithdrawalsRoot       common.Hash       `json:"withdrawalsRoot"`
+}
+
+// NewSimulationClient returns a client for eth_simulateV1 calls.
+func NewSimulationClient(cli *client.EngineClient) *SimulationClient {
+	return &SimulationClient{cli}
+}
+
+// Simulate calls `eth_simulateV1` with the provided block number and options.
+func (c *SimulationClient) Simulate(
```
shotes, line 138: What is the advantage of using `eth_simulateV1` vs making actual calls during the test?

rezbera: Unsure what you mean by actual calls.

TLDR of why we need simulateV1 is because it gives us a mechanism to generate the correct `stateRoot` and `receiptsRoot` which is necessary to pass `NewPayload` validation.

testing/simulated/malicious_consensus_test.go
```diff
@@ -71,9 +75,40 @@ func (s *SimulatedSuite) TestFinalizeBlock_BadBlock_Errors() {
...
+		&gethtypes.DynamicFeeTx{
+			Nonce:     0,
+			To:        &gethcommon.Address{1},
+			Value:     big.NewInt(100000000000),
+			Gas:       21000,
+			GasTipCap: big.NewInt(10000000),
+			GasFeeCap: big.NewInt(10000000),
+			Data:      []byte{},
+		},
+	)
```
abi87, line 91: @rez may be skill issue but I do not see why this tx is malicious or should fail. Can you clarify (maybe comment in code)

abi87: wait I guess it is `max fee per gas less than block base fee` (see final assert)

testing/simulated/malicious_proposer_test.go
```diff
@@ -72,9 +76,41 @@ func (s *SimulatedSuite) TestProcessProposal_BadBlock_IsRejected() {
 	)
 	s.Require().NoError(err)
 
+	// Sign a malicious transaction that is expected to fail.
+	recipientAddress := gethcommon.HexToAddress("0x56898d1aFb10cad584961eb96AcD476C6826e41E")
```
abi87, line 80: @rezbera what is the difference among this test and previous one? Asserts seem the same

testing/simulated/malicious_proposer_test.go
```diff
@@ -117,7 +153,7 @@ func (s *SimulatedSuite) TestProcessProposal_InvalidTimestamps_Errors() {
 	s.Require().Len(proposals, coreLoopIterations)
 	currentHeight := int64(blockHeight + coreLoopIterations)
 
-	// Prepare a valid block proposal, but 2 seconds in the future (i.e. attempt to roll timestamp forward)
+	// Prepare a block proposal, but 2 seconds in the future (i.e. attempt to roll timestamp forward)
```
abi87, line 156: to give some context:
they point here is that one should not be able to push timestamp arbitrarily far in the future, otherwise a malicious proposer can stop the network by proposing timestamp = max.Uint64.
We try and bound the timestamp as strictly as possible (while maintaining its strictly monotonicity)

testing/simulated/valid_chain_test.go
```diff
@@ -69,3 +78,114 @@ func (s *SimulatedSuite) TestFullLifecycle_ValidBlock_IsSuccessful() {
 	s.Require().NoError(err)
 	s.Require().Equal(proposedBlock.Message.GetHeader().GetBodyRoot(), stateHeader.GetBodyRoot())
 }
+
+// TestFullLifecycle_ValidBlockWithInjectedTransaction_IsSuccessful effectively serves as a demonstration for how one can
+// inject custom transactions and state transitions into the core loop.
```
abi87, line 83: awesome

abi87 (approved):
LGTM

testing/simulated/malicious_proposer_test.go
rezbera: One of them calls ProcessProposal and once of them calls FinalizeBlock

PR #2555, opened 2025-03-02
chore(networks):decommission bArtio and fix `make start` for Bepolia

chuck-bear:
Removes bArtio files and fixes the `make start` commands for Bepolia


rezbera (dismissed):
Test `make start-bepolia` and `make start-geth-bepolia`. Beacond crashes as EL is syncing which is expected, however, we need more EL peers.

rezbera (approved):
Was able to get testnet syncing with `make start-bepolia` and `make start-geth-bepolia/make start-reth-bepolia`

We were not reading el peers before. Now it works

abi87 (commented):
LGTM, retested `make start-bepolia`, confirmed it works

PR #2556, opened 2025-03-02
chore(config): cleared example config

abi87:
It was added pre-mainnet, to experiment with what default configs should be.
We do have now mainnet and testnet, so no need anymore


rezbera (approved):
Don't see it used anywhere so LGTM

PR #2557, opened 2025-03-03
chore(node-api): cleanup of DataResponse and refactor its usage

nidhi-singh02:
- Cleanup of `DataResponse`, noticed that [genesis API](https://ethereum.github.io/beacon-APIs/#/Beacon/getGenesis)  will have response in below format - 

```
{
  "data": {
    "genesis_time": "1590832934",
    "genesis_validators_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
    "genesis_fork_version": "0x00000000"
  }
}
```


node-api/handlers/beacon/historical.go
```diff
@@ -63,5 +63,5 @@ func (h *Handler) GetStateFork(c handlers.Context) (any, error) {
 	if err != nil {
 		return nil, err
 	}
-	return beacontypes.NewResponse(types.Wrap(fork)), nil
+	return beacontypes.NewResponse(fork), nil
```
nidhi-singh02, line 66: Wrap() is not needed.

node-api/handlers/beacon/types/response.go
```diff
@@ -78,6 +78,17 @@ type GenesisData struct {
...
+	Data GenesisData `json:"data"`
+}
+
+// NewGenesisResponse creates a new genesis response as per the spec.
+func NewGenesisResponse(data GenesisData) GenesisResponse {
+	return GenesisResponse{
+		Data: data,
+	}
+}
+
```
nidhi-singh02, line 91: Introduced this so that same pattern is being followed to return response as `NewResponse`

fridrik01: nit: I would remove `NewGenesisResponse` here and just initialize `GenesisResponse` directly inside `GetGenesis`. I think `NewResponse` was used since it sets some default values (Finalized & ExecutionOptimistic) in the return struct which helps reduce LOC which is not applicable here.

fridrik01 (approved):
just a small nit

node-api/handlers/beacon/types/response.go
nidhi-singh02: You have a valid point, I have addressed the comment.

PR #2559, opened 2025-03-03
chore(node-api) : Updated endpoints as per the spec

nidhi-singh02:
- I was going through the code, noticed that the block rewards API is not part of the route, added it. 

- While looking at beacon API spec, realised few endpoints are missing so added them as well.

Updated endpoints as per the [beacon API spec](https://ethereum.github.io/beacon-APIs/#/Beacon/).


node-api/handlers/beacon/routes.go
```diff
@@ -184,25 +193,49 @@ func (h *Handler) RegisterRoutes(
...
 			Handler: h.NotImplemented,
 		},
 		{
+			// Deprecated endpoint.
 			Method:  http.MethodPost,
 			Path:    "/eth/v1/beacon/pool/attestations",
 			Handler: h.NotImplemented,
 		},
 		{
+			Method:  http.MethodGet,
```
abi87, line 208: @nidhi-singh02 looks like some endpoints are missing like `/eth/v1/beacon/states/{state_id}/validator_identities`.
What is the principle around including some and not others?

abi87: Also some are not relevant anymore like `blinded_blocks` one?

abi87 (dismissed):
I would suggest either closing this down or having the full list of endpoints listed in https://ethereum.github.io/beacon-APIs/#/

node-api/handlers/beacon/routes.go
calbera: Would either add all from spec and then mark as deprecated OR only include ones relevant for us. okay either way

nidhi-singh02: I saw block rewards missing in route so added it. Then noticed few endpoints missing. 
Fair point, will go through the whole beacon list and add all the missing endpoints. I would prefer adding a comment as deprecated.

nidhi-singh02:
> I would suggest either closing this down or having the full list of endpoints listed in https://ethereum.github.io/beacon-APIs/#/

Added all the endpoints from node API except https://ethereum.github.io/beacon-APIs/#/ValidatorRequiredApi as it does not introduce new or different endpoints that are absent from the other tags (like “Validator” or “Beacon”). It would result in duplicate endpoints.
To represent the endpoints as deprecated added a comment. Let me know if we are in favour of removing the deprecated endpoints from route altogether.

node-api/handlers/debug/routes.go
```diff
@@ -37,7 +37,7 @@ func (h *Handler) RegisterRoutes(logger log.Logger) {
 		},
 		{
 			Method:  http.MethodGet,
-			Path:    "/eth/v2/debug/beacon/states/heads",
+			Path:    "/eth/v2/debug/beacon/heads",
 			Handler: h.NotImplemented,
 		},
```
nidhi-singh02, line 42: This is the correct path https://ethereum.github.io/beacon-APIs/#/Debug/getDebugChainHeadsV2

node-api/handlers/node/routes.go
```diff
@@ -47,7 +47,7 @@ func (h *Handler) RegisterRoutes(logger log.Logger) {
 		},
 		{
 			Method:  http.MethodGet,
-			Path:    "/eth/v1/node/peers/peer_count",
+			Path:    "/eth/v1/node/peer_count",
 			Handler: h.NotImplemented,
 		},
```
nidhi-singh02, line 52: Rectified the endpoint path as per https://ethereum.github.io/beacon-APIs/#/Node/getPeerCount

node-api/handlers/debug/routes.go
abi87: cc @calbera @rezbera @camembera

abi87 (commented):
LGTM but I would:
- consider introducing an `h.Deprecated` error for APIs that have been deprecated, instead of using just a comment and `h.NotImplemented` (I don't think we need to implement deprecated APIs)
- wait for explicit approval from @camembera and at least one among @calbera and @rezbera for the change in the debug API URL (this is currently "heavily" used to check deposits flow)

abi87 (commented):
Wrong comment about the debug API. Only `h.Deprecated` comment stands

nidhi-singh02:
> * consider introducing an `h.Deprecated` error for APIs that have been deprecated, instead of using just a comment and `h.NotImplemented` (I don't think we need to implement deprecated APIs)

Fair point, Introduced `h.Deprecated`.

node-api/handlers/rewards/routes.go
```diff
@@ -0,0 +1,49 @@
...
+package rewards
+
+import (
+	"net/http"
+
+	"github.com/berachain/beacon-kit/log"
+	"github.com/berachain/beacon-kit/node-api/handlers"
+)
+
+func (h *Handler) RegisterRoutes(logger log.Logger) {
```
calbera, line 30: I don't think we should register a new handler when these endpoints are part of the beacon namespace (eth/v1/beacon/... prefix)

calbera (dismissed):
overall makes sense. I am a bit confused by which endpoints are "NotImplemented" and which ones are "Depcreated"; would appreciate it if we can compile a list of these because its unclear at the emoment

EDIT: its now clear to me the difference between `Deprecated` and `NotImplemented`, but lets put the purpose of each and why theyre used on the functions so its clear from reading the code

node-api/handlers/rewards/routes.go
nidhi-singh02: Fair point, removed the handler from rewards.

nidhi-singh02:
> overall makes sense. I am a bit confused by which endpoints are "NotImplemented" and which ones are "Depcreated"; would appreciate it if we can compile a list of these because its unclear at the emoment
> 
> EDIT: its now clear to me the difference between `Deprecated` and `NotImplemented`, but lets put the purpose of each and why theyre used on the functions so its clear from reading the code

As discussed in DM, I have added a comment on `Deprecated` and `NotImplemented` to explain its purpose.

node-api/handlers/beacon/routes.go
```diff
@@ -151,11 +171,16 @@ func (h *Handler) RegisterRoutes(
...
+			Path:    "/eth/v1/beacon/rewards/blocks/:block_id",
+			Handler: h.GetBlockRewards,
 		},
 		{
 			Method:  http.MethodPost,
-			Path:    "/eth/v1/beacon/rewards/attestation/:epoch",
+			Path:    "/eth/v1/beacon/rewards/attestations/:epoch",
 			Handler: h.NotImplemented,
 		},
```
nidhi-singh02, line 185: Caught a typo here.

node-api/handlers/beacon/routes.go
```diff
@@ -151,11 +171,16 @@ func (h *Handler) RegisterRoutes(
 		{
 			Method:  http.MethodGet,
 			Path:    "/eth/v1/beacon/deposit_snapshot",
-			Handler: h.NotImplemented,
+			Handler: h.Deprecated,
+		},
+		{
+			Method:  http.MethodGet,
+			Path:    "/eth/v1/beacon/rewards/blocks/:block_id",
+			Handler: h.GetBlockRewards,
```
abi87, line 179: is this right?

nidhi-singh02: yes the endpoint path is right, refer https://ethereum.github.io/beacon-APIs/#/Beacon/getBlockRewards

abi87 (approved):

PR #2560, opened 2025-03-03
fix(state-transition): fixed ReadOnlyBeaconState interface - revive mockery

abi87:
We forgot to update `EVMInflationWithdrawal`.
Also it's nice to reuse `ReadOnlyBeaconState` and show `validateExecutionPayload` does not modify the beacon state, it just reads from it.
This is not related to https://github.com/berachain/beacon-kit/pull/2145

Note: fixing the mocks required regenerating them (mockery.yaml was broken too)


state-transition/core/interfaces.go
```diff
@@ -136,6 +136,6 @@ type ReadOnlyEth1Data interface {
 
 // ReadOnlyWithdrawals only has read access to withdrawal methods.
 type ReadOnlyWithdrawals interface {
-	EVMInflationWithdrawal() *engineprimitives.Withdrawal
```
fridrik01, line 139: Note there is still a `EVMInflationWithdrawal()` method defined [here](https://github.com/berachain/beacon-kit/blob/07200201ec9111f3ee59cf2ea24736b593fafce9/node-core/components/interfaces.go#L684), also in some mocks

state-transition/core/interfaces.go
```diff
@@ -28,6 +28,11 @@ import (
 	"github.com/berachain/beacon-kit/primitives/math"
 )
 
+type BeaconState interface {
+	ReadOnlyBeaconState
+	WriteOnlyBeaconState
+}
```
abi87, line 34: Added to regenerate mocks

state-transition/core/interfaces.go
abi87: Should be all solved now. Hardest was revive mockery

.mockery.yaml
```diff
@@ -30,47 +30,47 @@ outpkg: "mocks"
 resolve-type-alias: False # see https://vektra.github.io/mockery/latest/deprecations/#resolve-type-alias
 issue-845-fix: True # see https://vektra.github.io/mockery/latest/deprecations/#issue-845-fix
 packages:
-  github.com/berachain/beacon-kit/mod/execution/client/ethclient:
```
fridrik01, line 33: just did a grep, this file should be the last place where we refer to the old mod layout

fridrik01: also interesting that we didn't get a warn when running `make generate` since these packages didn't exist

fridrik01 (approved):
lgtm

.mockery.yaml
abi87: Indeed!

PR #2563, opened 2025-03-04
 test(simulated): Add Simulation Tests for Blobs (5/n)

rezbera:
Adds both valid and invalid blob tests. The Invalid tests are :

```go
// TestProcessProposal_InvalidBlobCommitment_Errors effectively serves as a test for a malicious blobs.
// Specifically, to bypass the commitment count check, we put 2 commitments into 1 tx and leave the
// other tx with zero commitments. This ultimately gets rejected by the execution client.
```
and

```go
// TestProcessProposal_InvalidBlobInclusionProof_Errors effectively serves as a test for a malicious blobs.
// Specifically, we tweak the KZG commitment inclusion proof such that it is invalid and must be rejected.
```


testing/simulated/valid_chain_test.go
```diff
@@ -189,3 +195,175 @@ func (s *SimulatedSuite) TestFullLifecycle_ValidBlockWithInjectedTransaction_IsS
...
+		s.TestNode.ChainSpec.ActiveForkVersionForSlot(blockHeight),
+	)
+	s.Require().NoError(err)
+
+	// Create the Blobs, with proofs and commitments
+	// Each blob will go into 1 transaction.
+	blobs := []*eip4844.Blob{{1, 2, 3}, {4, 5, 6}}
+	proofs, commitments := simulated.GetProofAndCommitmentsForBlobs(require.New(s.T()), blobs, s.TestNode.KZGVerifier)
+	s.Require().Len(proofs, 2)
+	s.Require().Len(commitments, 2)
```
abi87, line 241: @rezbera  nit: this 2 here is really `len(blobs)` right?

abi87 (approved):
LGTM.
@rezbera just check the main merge (there were conflicts) and a couple of nits

PR #2566, opened 2025-03-04
chore(e2e-test): Added tests for validator balances GET API

nidhi-singh02:
E2E tests for https://ethereum.github.io/beacon-APIs/#/Beacon/getStateValidatorBalances as go-eth2-client has methods for POST call.

Here we are directly calling the endpoint.


testing/e2e/e2e_api_test.go
```diff
@@ -515,9 +515,228 @@ func (s *BeaconKitE2ESuite) TestGetStateValidatorByPubkey() {
...
+		url = url + "?" + strings.Join(queryParams, "&")
+	}
+
+	resp, err := client.Get(url)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get validator balances: %w", err)
+	}
+	if resp == nil {
+		return nil, errors.New("received nil response")
+	}
```
abi87, line 546: this is checked on line 585 as well? Consider handling nil responses only in one place

nidhi-singh02: Wow! Kept it in func `getValidatorBalances` and removed NotNil check.

testing/e2e/e2e_api_test.go
```diff
@@ -515,9 +515,220 @@ func (s *BeaconKitE2ESuite) TestGetStateValidatorByPubkey() {
...
+		return nil, err
+	}
+
+	var genericResp beacontypes.GenericResponse
+	if err = json.Unmarshal(bodyBytes, &genericResp); err != nil {
+		return nil, err
+	}
+
+	// Convert the data field to JSON
+	dataBytes, err := json.Marshal(genericResp.Data)
```
shotes, line 568: Unfortunate that you have to unmarshal as `GenericResponse`, re-marshal the `any` type data, and then re-unmarshal that data into the `[]ValidatorBalanceData`. I suppose there's not really a better way around it other than creating a custom "Response" type like `ValidatorBalanceResponse` for the original `json.Unmarshal`.

Just pointing it out, I don't think there's a much better way. Works good enough for e2e tests.

shotes (approved):
LGTM

abi87 (approved):

PR #2568, opened 2025-03-04
fix(abci): ignore comet ctx and use app ctx

shotes:
Currently, beacond hangs on SIGTERM if its inside a retry timeout. This is because the context used in the retry is not cancelled in our signal hander. 

To address this, this PR makes the context available to the `cometbft` service which it gets from the context passed in to `Start()`. Now, when signal handler catches a signal, the context cancellation will be checked by the ABCI++ endpoints properly. This causes us to properly handle signal cancellation in the retry timeout. 

Unfortunately, CometBFT does not yet support context cancellation, or context usage at all. It currently passes in `context.TODO()` into all of the ABCI methods, so the ctx we see in the ABCI function implementations is not connected to the ctx from our node. I also can't seem to find anything in cometBFT RFC or docs that indicate supporting this.

Since CometBFT does not support context cancellation, this means that there is no clean way out -- we MUST return error in `ProcessProposal` and `FinalizeBlock` (which causes the `CONSENSUS FAILURE!!!` panic). If we don't then our node will be attempting to participate in consensus with potentially corrupt, invalid, or missing data. I have tested this behavior by not returning an error from `FinalizeBlock` during context cancellation, and indeed my node still proceeded to call `Commit`.


consensus/cometbft/service/abci.go
```diff
@@ -36,11 +36,29 @@ var (
 	errNilFinalizeBlockState = errors.New("finalizeBlockState is nil")
 )
 
+// combineContexts combines the CometBFT context with the original context from the node.
+// TODO: This can be removed if CometBFT implements contexts.
```
calbera, line 40: As an uninformed reader, this comment is very confusing. CometBFT obviously passes in a context but its a TODO().

Q: why does this mean we need a combined context?

consensus/cometbft/service/abci.go
```diff
@@ -36,11 +36,29 @@ var (
...
+func (s *Service) combineContexts(ctx context.Context) context.Context {
+	if s.ctx == nil {
+		return ctx
+	}
+	newCtx, cancel := context.WithCancel(context.Background())
```
calbera, line 45: why not just use a parent context here? Seems like the perfect use case for this secenario

consensus/cometbft/service/abci.go
```diff
@@ -36,11 +36,29 @@ var (
...
+	go func() {
+		select {
+		case <-ctx.Done():
+			cancel()
+		case <-s.ctx.Done():
+			cancel()
+		}
```
abi87, line 52: @shotes: is this leaking goroutines? 
We create a goroutine per each callback and never exit them.
Probably we need to make things slighly more complex.
We want to:
- have a single goroutine monitoring contextes
- reset the context from consensus every time we have a callback

abi87: An possible alternative is to return a channel allowing to close the goroutine?

Something like:

```
	done := make(chan struct{})
	newCtx, cancel := context.WithCancel(context.Background())
	go func() {
		select {
		case <-ctxConsensus.Done():
			cancel()
		case <-ctxApp.Done():
			cancel()
		case <-done:
			return
		}
	}()
	return newCtx, done
```

and then change callback as:

```
	ctx, done := combineContexts(ctxConsensus, s.ctx)
	defer func() { close(done) }()
	return s.finalizeBlock(ctx, req)
```

I am not 100% sure this is good, cause I do not know what happens if `s.Ctx` is cancelled among callbacks, when no goroutine is spinning.

consensus/cometbft/service/abci.go
```diff
@@ -36,11 +36,29 @@ var (
 	errNilFinalizeBlockState = errors.New("finalizeBlockState is nil")
 )
 
+// combineContexts combines the CometBFT context with the original context from the node.
+// TODO: This can be removed if CometBFT implements contexts.
+func (s *Service) combineContexts(ctx context.Context) context.Context {
+	if s.ctx == nil {
+		return ctx
+	}
```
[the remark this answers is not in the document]
abi87: Consider introducing
```
// ResetAppCtx is used in simulated tests to make sure we
// duly configure s.ctx even is Start is not called
func (s *Service) ResetAppCtx(ctx context.Context) {
	s.ctx = ctx
}
```

and calling it in 

```
func (s *SimComet) Start(ctx context.Context) error {
	// Just make sure ctx is initied similarly to what
	// happens when s.Comet.Start is called
	s.Comet.ResetAppCtx(ctx)
	return nil
}
```

consensus/cometbft/service/abci.go
abi87: @rezbera @shotes I tried this and I still got a crash.
May it be that given that in simulated tests, node.Start is called in a different go-routine there is race?
We sleep two seconds after start, maybe not enough?

consensus/cometbft/service/abci.go
shotes: Clarified comment.

consensus/cometbft/service/commit.go
```diff
@@ -31,7 +31,7 @@ import (
 
 func (s *Service) commit(
 	context.Context, *cmtabci.CommitRequest,
-) (*cmtabci.CommitResponse, error) {
+) *cmtabci.CommitResponse {
```
shotes, line 34: linter suddenly started complaining with
```
consensus/cometbft/service/commit.go:34:29: (*Service).commit - result 1 (error) is always nil (unparam)
) (*cmtabci.CommitResponse, error) {
```

Not sure why it popped up in this particular branch, but this fixes it.

consensus/cometbft/service/prepare_proposal.go
```diff
@@ -34,6 +34,12 @@ func (s *Service) prepareProposal(
 	ctx context.Context,
 	req *cmtabci.PrepareProposalRequest,
 ) (*cmtabci.PrepareProposalResponse, error) {
+	// Check if ctx is still good. CometBFT does not check this.
```
rezbera, line 37: why not catch these in `abci.go` before it leaks to the service? not a big deal but that seems cleaner

consensus/cometbft/service/prepare_proposal.go
```diff
@@ -54,7 +60,8 @@ func (s *Service) prepareProposal(
 		s.getContextForProposal(
 			s.prepareProposalState.Context(),
 			req.Height,
-		),
+			// Preserve the CosmosSDK context while using the correct base ctx.
+		).WithContext(ctx),
```
abi87, line 64: @shotes why is this `WithContext` needed? `resetState` does already that

shotes: `resetState` does it indeed. However, we conditionally override that `ctx` during `getContextForProposal` with the old ctx from the cosmos SDK context - `s.finalizeBlockState.Context().CacheContext()`

shotes: Actually, a better fix would be to do this `WithContext()` specifically during that one case in `getContextForProposal()`. Will modify this.

consensus/cometbft/service/finalize_block.go
```diff
@@ -56,6 +62,9 @@ func (s *Service) finalizeBlockInternal(
 		s.finalizeBlockState = s.resetState(ctx)
 	}
 
+	// Preserve the CosmosSDK context while using the correct base ctx.
+	s.finalizeBlockState.SetContext(s.finalizeBlockState.Context().WithContext(ctx))
```
abi87, line 66: isn't this equivalent to `s.finalizeBlockState.Context().WithContext(ctx)`?

shotes: No, because `WithContext` is not built on a pointer receiver. It does not set the ctx for the receiver, only the return value.

consensus/cometbft/service/abci.go
```diff
@@ -36,11 +36,37 @@ var (
...
+	return newCtx, cancel
+}
+
 func (s *Service) InitChain(
 	ctx context.Context,
 	req *cmtabci.InitChainRequest,
 ) (*cmtabci.InitChainResponse, error) {
-	return s.initChain(ctx, req)
+	cCtx, cancel := combineContexts(s.ctx, ctx)
+	defer cancel()
```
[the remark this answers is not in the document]
shotes: Agreed. I originally wanted to just leave the cancel to stop the goroutine, since it would be cleaner. Even though we don't currently have goroutine that would still be running on this child context by the time it is cancelled, it could be confusing in the future why a goroutine is getting cancelled early. Will change this.

consensus/cometbft/service/abci.go
```diff
@@ -36,11 +36,40 @@ var (
...
+	go func() {
+		select {
+		case <-done:
+			return
+		// If ctxApp context is cancelled, newCtx is implicitly cancelled by
+		// being an inherited child context.
+		case <-newCtx.Done():
+			return
+		case <-ctxComet.Done():
+			// If the ctxComet context is cancelled, bridge its cancellation to
```
rezbera, line 57: since the `ctxComet` is always `context.TODO()`, this case would never be hit right?
since internally, context.TODO is

```
func (emptyCtx) Done() <-chan struct{} {
	return nil
}
```

rezbera: if comet's context is not TODO and is appropriately cancelled, we will only ever need 1 context and do not need to combine.

shotes:
Opted to entirely remove the `combineContexts` functionality. The original reason for implementing the `combineContexts` functionality was so that we could additionally capture the `ctx` provided by comet, **IF** they ever decided to implement it.

This original reason was not a very good reason - IF they implement the `ctx`, then we'd want to remove `combineContexts` anyways, since `ctx` would have become a child context of `s.ctx`.

Removing this functionality entirely streamlines the whole thing, and it was unnecessary in the first place. Now, if comet ever implements `ctx` properly, we can simply replace the `s.ctx` with `ctx`.

consensus/cometbft/service/finalize_block.go
```diff
@@ -54,6 +54,9 @@ func (s *Service) finalizeBlockInternal(
 	// CometBFT.
 	if s.finalizeBlockState == nil {
 		s.finalizeBlockState = s.resetState(ctx)
+	} else {
+		// Preserve the CosmosSDK context while using the correct base ctx.
+		s.finalizeBlockState.SetContext(s.finalizeBlockState.Context().WithContext(ctx))
```
rezbera, line 59: bit unsure on this

rezbera (approved):
1 non-blocking query , but rest lgtm

consensus/cometbft/service/finalize_block.go
shotes: So the cosmos SDK context, which I will refer to as `cosmosCtx`, contains a reference to the `baseCtx`, which is the regular golang `context.Context`. This `cosmosCtx` stores a bunch of data in it for our state and commit multi store and whatnot.

When we `resetState(ctx)`, we are creating a new `cosmosCtx`, where `baseCtx = ctx`.

If we do not resetState(ctx)`, then the already existing `finalizeBlockState` is used. This state has the old `cosmosCtx`, which has the old `ctx`. We want to make sure we are using the new `ctx` that is passed in for this function, so we can simply set `baseCtx = ctx` here.

Indeed, this is not as big of a problem now, since the context isn't getting cancelled like it was in a previous iteration of this PR. However, it is still not accurately using the correct `ctx` unless we make this change.

rezbera: > However, it is still not accurately using the correct ctx unless we make this change.

I had assumed it was always going to be using [this context](https://github.com/berachain/beacon-kit/blob/b26a27600d5d228d20df32d838590459fda9af17/consensus/cometbft/service/service.go#L107-L106).

For my understanding, is it fair to say that we don't need to change the context here given we have this universal base context which _currently_ never changes?

calbera, line 59: can you make a helper function on `state` to reset/update the base ctx

consensus/cometbft/service/abci.go
```diff
@@ -37,19 +37,28 @@ var (
 )
 
 func (s *Service) InitChain(
-	ctx context.Context,
+	_ context.Context,
 	req *cmtabci.InitChainRequest,
 ) (*cmtabci.InitChainResponse, error) {
-	return s.initChain(ctx, req)
+	//nolint:contextcheck // see s.ctx comment for more details
+	return s.initChain(s.ctx, req)
```
abi87, line 44: unlike other ABCI callbacks, we don't test `s.ctx.Err() != nil` here. Should we?

abi87: cc @shotes please cross check this

consensus/cometbft/service/finalize_block.go
abi87: @rezbera if you check the call hierarchy you will see that `FinalizeBlock` does use `s.ctx`

abi87: Actually @rezbera we can solve the issue by just not passing ctx as parameter, but reusing s.ctx everywhere, which this PR basically does. So no need to look up the call hierarchy

consensus/cometbft/service/abci.go
shotes: We can and probably should, but imo it doesn't matter much since initChain runs once on genesis and basically never again.

shotes: I see that you've made the change. Good idea.

testing/simulated/simulated_test.go
```diff
@@ -95,7 +95,7 @@ func (s *SimulatedSuite) SetupTest() {
 	}()
 
 	// Allow a short period for services to fully initialize.
-	time.Sleep(2 * time.Second)
+	time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
rezbera: I've changed to use a reliable method in https://github.com/berachain/beacon-kit/pull/2572

It's not ideal, because ideally we'd have a readiness probe API but this is better than sleeping arbitrary time.

PR #2569, opened 2025-03-04
chore(node): Remove public FetchServiceRegistry

rezbera:
It is not necessary to expose this from the node. I previously thought it was necessary but this is not the case


calbera (approved):
thank you ahaha LGTM

PR #2576, opened 2025-03-06
chore(abci): Pass the context parameter

shotes:
Passing the beacond App's `s.ctx` context field as a parameter to the ABCI functions imitates the functionality of using the `ctx` from CometBFT. This makes it so that the arguments in each ABCI function call are easily swappable between `s.ctx` and `ctx`, should CometBFT ever fully implement their `ctx`.

### DIFF
--- a/consensus/cometbft/service/abci.go
+++ b/consensus/cometbft/service/abci.go
@@ -46,7 +46,7 @@ func (s *Service) InitChain(
 		return &cmtabci.InitChainResponse{}, s.ctx.Err()
 	}
 	//nolint:contextcheck // see s.ctx comment for more details
-	return s.initChain(req) // internally this uses s.ctx
+	return s.initChain(s.ctx, req)
 }
 
 // PrepareProposal implements the PrepareProposal ABCI method and returns a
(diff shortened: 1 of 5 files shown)


consensus/cometbft/service/service.go
```diff
@@ -174,7 +174,7 @@ func (s *Service) Start(
 		return err
 	}
 
-	s.ctx = ctx
+	s.ResetAppCtx(ctx)
```
shotes, line 177: Also just a nit. Might as well use the setter function.

consensus/cometbft/service/abci.go
```diff
@@ -46,7 +46,7 @@ func (s *Service) InitChain(
 		return &cmtabci.InitChainResponse{}, s.ctx.Err()
 	}
 	//nolint:contextcheck // see s.ctx comment for more details
-	return s.initChain(req) // internally this uses s.ctx
+	return s.initChain(s.ctx, req)
```
shotes, line 49: With this PR, I could easily interchange this `s.ctx` with the `ctx` that is provided in the function parameters.

shotes:
Also note that on `main`, `PrepareProposal` is currently inconsistent with the other ABCI calls. It is currently implemented exactly like the changes in this PR. So this PR will make all ABCI calls consistent with each other.

rezbera (approved):
LGTM

abi87 (approved):

PR #2578, opened 2025-03-07
chore(node-api): slot fix and correct slot usage in error

nidhi-singh02:
Extending [slot fix](https://github.com/berachain/beacon-kit/pull/2538/files), noticed that due to redefining the same variable name, the error message is not printing the correct slot. 

Please note : Will make the error message verbose in other places as and when we implement them.

### DIFF
--- a/node-api/backend/validator.go
+++ b/node-api/backend/validator.go
@@ -80,7 +80,7 @@ func (f *validatorFilters) parseID(id string) {
 func (b Backend) FilteredValidators(
 	slot math.Slot, ids []string, statuses []string,
 ) ([]*beacontypes.ValidatorData, error) {
-	st, slot, err := b.stateFromSlot(slot)
+	st, resolvedSlot, err := b.stateFromSlot(slot)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to get state from slot %d", slot)
 	}
@@ -92,7 +92,7 @@ func (b Backend) FilteredValidators(
 
 	// Parse all IDs and pubkeys once at the start
 	filters := parseValidatorIDs(ids)
-	epoch := b.cs.SlotToEpoch(slot)
+	epoch := b.cs.SlotToEpoch(resolvedSlot)
 
 	return filterAndBuildValidatorData(st, validators, filters, epoch, statuses)
 }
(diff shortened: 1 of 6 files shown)


node-api/backend/validator.go
```diff
@@ -237,7 +237,7 @@ func (b Backend) ValidatorByID(slot math.Slot, id string) (*beacontypes.Validato
 }
 
 func (b Backend) ValidatorBalancesByIDs(slot math.Slot, ids []string) ([]*beacontypes.ValidatorBalanceData, error) {
-	st, slot, err := b.stateFromSlot(slot)
+	st, _, err := b.stateFromSlot(slot)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to get state from slot %d", slot)
```
nidhi-singh02, line 242: As the slot returned from func `stateFromSlot` is not used, so skipping it.

node-api/backend/validator.go
```diff
@@ -80,7 +80,7 @@ func (f *validatorFilters) parseID(id string) {
 func (b Backend) FilteredValidators(
 	slot math.Slot, ids []string, statuses []string,
 ) ([]*beacontypes.ValidatorData, error) {
-	st, slot, err := b.stateFromSlot(slot)
+	st, resolvedSlot, err := b.stateFromSlot(slot)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to get state from slot %d", slot)
 	}
```
nidhi-singh02, line 86: Noticed that the error message was getting overridden by the slot from state (stateFromSlot) instead of the slot passed in, hence renamed the var to `resolvedSlot` to correctly log the error message.

node-api/backend/validator.go
```diff
@@ -196,7 +196,7 @@ func buildValidatorData(
 }
 
 func (b Backend) ValidatorByID(slot math.Slot, id string) (*beacontypes.ValidatorData, error) {
-	st, slot, err := b.stateFromSlot(slot)
+	st, resolvedSlot, err := b.stateFromSlot(slot)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to get state from slot %d", slot)
 	}
```
nidhi-singh02, line 202: To rectify the error message.

shotes:
I think I liked the original behavior where it prints out the resolved slot, because that seems like more useful information. But also it's good to clarify the variable names so we know which slot is being used.

abi87 (approved):
extended error msgs to address @shotes's comments

PR #2580, opened 2025-03-10
feat(node-api): Config spec API added

nidhi-singh02:
- Implemented https://ethereum.github.io/beacon-APIs/#/Config/getSpec
- Added E2E tests


node-api/backend/spec.go
```diff
@@ -0,0 +1,41 @@
...
+
+func (b *Backend) Spec() (*types.SpecData, error) {
+	chainSpec := b.ChainSpec()
+	if chainSpec == nil {
+		return nil, errors.New("chain spec not found")
+	}
+	return &types.SpecData{
+		DepositContractAddress: chainSpec.DepositContractAddress(),
+		// Network ID is same as eth1 chain ID.
+		DepositNetworkID:                chainSpec.DepositEth1ChainID(),
```
nidhi-singh02, line 36: Using Eth1 chain ID. The other approach is to introduce `DepositNetworkID` in the chain spec, and have its value equal to Chain ID.
As the current approach will reduce LoC as it has the same value anyway. I am open to suggestions.

nidhi-singh02, line 36: Using Eth1 chain ID. The other approach is to introduce DepositNetworkID in the chain spec, and have its value equal to Chain ID.
As the current approach will reduce LoC as it has the same value anyway. I am open to suggestions.

config/spec/defaults.go
```diff
@@ -57,8 +57,10 @@ const (
 	defaultTargetSecondsPerEth1Block = 2 // Berachain specific.
 
 	// Fork-related values.
-	defaultDeneb1ForkEpoch  = 9999999999999998 // Set as a future epoch as not yet determined.
-	defaultElectraForkEpoch = 9999999999999999 // Set as a future epoch as not yet determined.
+	defaultDeneb1ForkEpoch                 = 9999999999999998 // Set as a future epoch as not yet determined.
+	defaultElectraForkEpoch                = 9999999999999999 // Set as a future epoch as not yet determined.
+	defaultInactivityPenaltyQuotient       = 67108864         // Phase 0
+	defaultInactivityPenaltyQuotientAltair = 50331648         // Altair
```
nidhi-singh02, line 63: These values are taken from ETH2.0 consensus spec.

node-api/handlers/config/types/response.go
```diff
@@ -0,0 +1,69 @@
...
+type specJSON struct {
+	DepositContractAddress          string `json:"DEPOSIT_CONTRACT_ADDRESS"`
+	DepositNetworkID                string `json:"DEPOSIT_NETWORK_ID"`
+	DomainAggregateAndProof         string `json:"DOMAIN_AGGREGATE_AND_PROOF"`
+	InactivityPenaltyQuotient       string `json:"INACTIVITY_PENALTY_QUOTIENT"`
+	InactivityPenaltyQuotientAltair string `json:"INACTIVITY_PENALTY_QUOTIENT_ALTAIR"`
+}
+
+func (sd SpecData) MarshalJSON() ([]byte, error) {
+	return json.Marshal(specJSON{
```
nidhi-singh02, line 52: Marshalling is done to ensure API response is returned as per node spec.

chain/data.go
```diff
@@ -112,6 +112,8 @@ type SpecData struct {
 	//
 	// InactivityPenaltyQuotient is the inactivity penalty quotient.
 	InactivityPenaltyQuotient uint64 `mapstructure:"inactivity-penalty-quotient"`
+	// InactivityPenaltyQuotientAltair is the inactivity penalty quotient for Altair.
+	InactivityPenaltyQuotientAltair uint64 `mapstructure:"inactivity-penalty-quotient-altair"`
```
abi87, line 116: not sure adding configs we will never use only to serve APIs responses is the right approach.
Wouldn't it be better to configure this value only where needed?
cc @calbera

config/spec/defaults.go
abi87: cc @calbera I think our current fork is Deneb1 already, so there should be no need to add defaults here?

testing/simulated/transformers.go
```diff
@@ -41,7 +41,11 @@ import (
 
 // transformSimulatedBlockToGethBlock converts a simulated execution block into a Geth-style block.
 // It uses the provided transactions and parent beacon root to construct a new execution block header.
-func transformSimulatedBlockToGethBlock(simBlock *execution.SimulatedBlock, txs []*gethtypes.Transaction, parentBeaconRoot libcommon.Root) *gethtypes.Block {
+func transformSimulatedBlockToGethBlock(
```
abi87, line 44: this changes do not seem related to the config spec. Why are they here?

abi87 (dismissed):
a few questions around the extra configs. Deneb1 related changes seems wrong to me, given our current fork

testing/simulated/transformers.go
nidhi-singh02: This got introduced as part of this [commit](https://github.com/berachain/beacon-kit/pull/2580/commits/d9a2eca287f356d49759b9fa9d153e31ecc57902). @calbera shall we remove them ?

config/spec/defaults.go
nidhi-singh02: I see, @calbera could you pls confirm.

testing/simulated/transformers.go
calbera: Ahh that’s my bad, my linter always be doing the most. Good to remove!

calbera (commented):
As discussed on sync, I would prefer keeping the Inactivity Penality Quotient to be consistent with spec (phase0, Altair) until we implement slashing for inactivity correctly. At which point we can correctly return for our versions (Deneb, Deneb1, …)

testing/simulated/transformers.go
nidhi-singh02: Removed.

chain/data.go
calbera: In agreement. @nidhi-singh02 Let's remove any changes to chain/ and config/spec/ in this PR. These values can be hardcoded as placeholders as needed. As discussed this is only temporary until we properly implement and use inactivityPenaltyQuotients

calbera (dismissed):
some minor changes

node-api/backend/mocks/chain_spec.mock.go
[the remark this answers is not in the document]
nidhi-singh02: As we are fetching values from chainspec in the node-api/backend/spec, the addition of funcs to chainspec interface https://github.com/berachain/beacon-kit/pull/2580/files#diff-3ed12b4715a8cb640c8484ef9b44bf00a0629d76b67695c03f562c7d776c6344R41-R43 . Mock will have changes.

node-api/backend/spec.go
```diff
@@ -0,0 +1,40 @@
...
+// TITLE.
+
+package backend
+
+import (
+	"github.com/berachain/beacon-kit/errors"
+	"github.com/berachain/beacon-kit/node-api/handlers/config/types"
+)
+
+func (b *Backend) Spec() (*types.SpecData, error) {
```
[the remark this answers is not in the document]
nidhi-singh02: added comment

config/spec/defaults.go
nidhi-singh02: My bad, Thanks for clarifying. I have hardcoded the values in node-api/backend/spec.go with comments. And removed from here.

chain/data.go
nidhi-singh02: I have removed changes to chain/ and config/spec.  And hardcoded these values in node-api.

calbera:
Made https://github.com/berachain/beacon-kit/pull/2580/commits/2d657395d3e832534b45dd9e29e7a0bade3a607e to show some simplifications:
- gothonic package design, where the ChainSpec interface is defined as needed per package and used as such
- removed unnecessary code to handle json marshalling (when the types are not complex/dynamic as such, this is more readable and easier)

calbera (approved):

shotes (commented):
nit in return values for INACTIVITY_PENTALTY_QUOTIENT. Otherwise LGTM

node-api/handlers/config/spec.go
```diff
@@ -0,0 +1,58 @@
...
+		DepositContractAddress: cs.DepositContractAddress().String(),
+
+		// Network ID is same as eth1 chain ID.
+		DepositNetworkID: math.U64(cs.DepositEth1ChainID()).Base10(),
+
+		DomainAggregateAndProof: cs.DomainTypeAggregateAndProof().String(),
+
+		// Currently these are placeholders, will be replaced with the correct values for our
+		// versions like Deneb, Deneb1 etc once we implement slashing for inactivity.
+		InactivityPenaltyQuotient:       Phase0InactivityPenaltyQuotient.Base10(),
```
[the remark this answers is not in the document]
calbera: i'm good either way. Adopted in https://github.com/berachain/beacon-kit/pull/2580/commits/0fd4960b883675669d2dc4ecbaa2c82b961b83f2

PR #2581, opened 2025-03-10
chore(abci): Add FATAL errors for app

shotes:
This adds FATAL errors to the ABCI app. This makes the response to Fatal Engine API errors consistent across `PrepareProposal`, `ProcessProposal`, and `FinalizeBlock`.

### DIFF
--- a/consensus/cometbft/service/prepare_proposal.go
+++ b/consensus/cometbft/service/prepare_proposal.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/berachain/beacon-kit/consensus/types"
+	"github.com/berachain/beacon-kit/errors"
 	"github.com/berachain/beacon-kit/primitives/math"
 	cmtabci "github.com/cometbft/cometbft/abci/types"
 )
@@ -71,7 +72,10 @@ func (s *Service) prepareProposal(
 		s.prepareProposalState.Context(),
 		slotData,
 	)
-	if err != nil {
+	switch {
+	case errors.IsFatal(err):
+		return &cmtabci.PrepareProposalResponse{}, err
+	case err != nil:
 		s.logger.Error(
 			"failed to prepare proposal",
 			"height", req.Height,
@@ -79,9 +83,7 @@ func (s *Service) prepareProposal(
 			"err", err,
 		)
 		return &cmtabci.PrepareProposalResponse{Txs: [][]byte{}}, nil
+	default:
+		return &cmtabci.PrepareProposalResponse{Txs: [][]byte{blkBz, sidecarsBz}}, nil
 	}
-
-	return &cmtabci.PrepareProposalResponse{
-		Txs: [][]byte{blkBz, sidecarsBz},
-	}, nil
 }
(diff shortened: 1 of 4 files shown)


errors/mod.go
calbera, line 1: nit: since its no longer its own go module, should prob rename this file to something like errors.go

rezbera (commented):
It'll look bad if validators have a higher occurrance of their beacon node crashing with the new release (something we faced with the lockstep EL/CL changes). 

Is there anything we can do to put the validators at ease as to why their beacon node crashed - e.g. excessive logging?

consensus/cometbft/service/prepare_proposal.go
calbera, line 1: I think payload builders should still be able to send an empty block if their EL went fatal. Something to discuss further

consensus/cometbft/service/prepare_proposal.go
```diff
@@ -71,17 +72,18 @@ func (s *Service) prepareProposal(
 		s.prepareProposalState.Context(),
 		slotData,
 	)
-	if err != nil {
+	switch {
+	case errors.IsFatal(err):
+		return &cmtabci.PrepareProposalResponse{}, err
```
calbera, line 77: nit: can this just be `return nil, err`

consensus/cometbft/service/process_proposal.go
```diff
@@ -68,20 +69,23 @@ func (s *Service) processProposal(
...
-	status := cmtabci.PROCESS_PROPOSAL_STATUS_ACCEPT
 	err := s.Blockchain.ProcessProposal(
 		s.processProposalState.Context(),
 		req,
 	)
-	if err != nil {
-		status = cmtabci.PROCESS_PROPOSAL_STATUS_REJECT
+	switch {
+	case errors.IsFatal(err):
+		return &cmtabci.ProcessProposalResponse{Status: cmtabci.PROCESS_PROPOSAL_STATUS_UNKNOWN}, err
```
calbera, line 78: q: is this Status even checked by Comet if an error is returned?

consensus/cometbft/service/prepare_proposal.go
shotes: fixed

shotes:
> It'll look bad if validators have a higher occurrance of their beacon node crashing with the new release (something we faced with the lockstep EL/CL changes).
>
> Is there anything we can do to put the validators at ease as to why their beacon node crashed - e.g. excessive logging?

I'm not sure this results in a higher occurrence of beacon node crashing. If a Fatal error happens during ProcessProposal (and gets glossed over because we don't crash immediately in process proposal), then it is extremely likely that the same Fatal error happens again during FinalizeBlock. For example, if my EL crashes, the engine api is going to keep returning the Fatal `connection refused` error until it reaches FinalizeBlock, where the CL then crashes.

Definitely can do with more logging and I see your point.

consensus/cometbft/service/prepare_proposal.go
shotes: An empty proposed block would get rejected by validators and prevoted nil, whereas a missed proposed would timeout and get prevoted nil**. The behavior at consensus level is the same, but we're adding extra network congestion by broadcasting empty proposal. Why would we do that if we don't have to?

**From [comet spec](https://docs.cometbft.com/v1.0/spec/consensus/consensus#prevote-step-heighthroundr):
> Else, if the proposal is invalid or wasn’t received on time, it prevotes <nil>.

PR #2585, opened 2025-03-11
feat(pectra): Pectra Types

rezbera:


consensus-types/types/execution_requests.go
```diff
@@ -0,0 +1,106 @@
...
+const sszWithdrawRequestSize = 76          // ExecutionAddress = 20, ValidatorPubKey = 48, Amount = 8
+const dynamicFieldsInExecutionRequests = 2 // 2 since two dynamic objects (Deposits, Withdrawals)
+
+type ExecutionRequests struct {
+	Deposits    []*DepositRequest
+	Withdrawals []*WithdrawalRequest
+}
+
+// DepositRequest is introduced in EIP6110 which is currently not processed.
+type DepositRequest struct {
```
calbera, line 48: note: this is actually equivalent to our current Deposit type

consensus-types/types/block.go
```diff
@@ -44,18 +44,10 @@ type BeaconBlock struct {
...
-	// BbVersion is the BbVersion of the beacon block.
-	// BbVersion must be not serialized but it's exported
+	// forkVersion is the forkVersion of the beacon block.
+	// forkVersion must be not serialized but it's exported
 	// to allow unit tests using reflect on beacon block.
-	BbVersion common.Version `json:"-"`
-}
-
-// Empty creates an empty beacon block.
-func (*BeaconBlock) Empty() *BeaconBlock {
```
rezbera, line 54: was not used

consensus-types/types/payload_header.go
```diff
@@ -92,20 +86,22 @@ func (h *ExecutionPayloadHeader) Empty() *ExecutionPayloadHeader {
 func (h *ExecutionPayloadHeader) NewFromSSZ(
 	bz []byte, forkVersion common.Version,
 ) (*ExecutionPayloadHeader, error) {
-	h = h.Empty()
-	h.EphVersion = forkVersion
-	return h, h.UnmarshalSSZ(bz)
+	h = h.Empty(forkVersion)
+	err := h.UnmarshalSSZ(bz)
+	if err != nil {
```
rezbera, line 91: more clear like this imo

consensus-types/types/signed_beacon_block.go
```diff
@@ -63,8 +63,9 @@ func NewSignedBeaconBlockFromSSZ(
 		EnsureNotNilWithdrawals(block.Message.Body.ExecutionPayload)
 
 		// duly setup fork version in every relevant block member
-		block.Message.Body.ExecutionPayload.EpVersion = forkVersion
-		block.Message.BbVersion = forkVersion
+		block.Message.Body.ExecutionPayload.forkVersion = forkVersion
```
rezbera, line 66: We gotta find a better way than this

engine-primitives/engine-primitives/withdrawals_test.go
```diff
@@ -54,6 +57,27 @@ func TestWithdrawals(t *testing.T) {
 		require.NotEmpty(t, root)
 	})
 
+	t.Run("HashTreeRoot Comparison", func(t *testing.T) {
```
rezbera, line 60: new test to ensure we align with `zrnt`

primitives/constraints/basic.go
```diff
@@ -20,19 +20,7 @@
...
 	Empty() SelfT
 }
-
-// Nillable is a constraint that requires a type to have an IsNil method.
-type Nillable interface {
-	IsNil() bool
-}
-
-// Versionable is a constraint that requires a type to have a Version method.
-type Versionable interface {
```
rezbera, line 36: unused

storage/encoding/ssz.go
```diff
@@ -71,7 +71,6 @@ func (SSZValueCodec[T]) ValueType() string {
 // infer the type of an interface.
 type SSZInterfaceCodec[T interface {
 	constraints.SSZMarshallable
-	constraints.Versionable
```
rezbera, line 74: wasn't needed

testing/quick/compare_test.go
```diff
@@ -24,82 +24,22 @@ package compare_test
...
 )
 
 var (
-	c    = quick.Config{MaxCount: 5_000}
+	c    = quick.Config{MaxCount: 10_000}
 	hFn  = ztree.GetHashFn()
 	spec = zspec.Mainnet
 )
 
-func TestExecutionPayloadHashTreeRootZrnt(t *testing.T) {
```
rezbera, line 50: moved to new file as I expect we'll add more

testing/quick/compare_test.go
```diff
@@ -24,84 +24,27 @@ package compare_test
...
-
-		return bytes.Equal(typeRoot[:], containerRoot[:]) &&
-			bytes.Equal(typeRoot[:], zRoot[:])
-	}
-	if err := quick.Check(f, &c); err != nil {
-		t.Error(err)
-	}
-}
-
+// TODO(pectra): None of the iterations makes it to `HashTreeRoot` as they all get caught by the nil checks.
```
rezbera, line 43: need to come back to this but don't want to get sidequested

consensus-types/types/block.go
```diff
@@ -72,8 +64,11 @@ func NewBeaconBlockWithVersion(
 			ProposerIndex: proposerIndex,
 			ParentRoot:    parentBlockRoot,
 			StateRoot:     common.Root{},
-			Body:          &BeaconBlockBody{},
-			BbVersion:     forkVersion,
+			Body: &BeaconBlockBody{
```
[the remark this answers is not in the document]
calbera: you could have this type of reverse "inheritance", where BeaconBlockBody has a pointer to the BeaconBlock but only as a Versionable type.

consensus-types/types/block.go
```diff
@@ -153,16 +148,21 @@ func (b *BeaconBlock) GetStateRoot() common.Root {
...
 	b.StateRoot = root
 }
 
+// GetForkVersion identifies the version of the BeaconBlock.
+func (b *BeaconBlock) GetForkVersion() common.Version {
+	return b.forkVersion
+}
+
+// SetForkVersion sets the fork version of the BeaconBlock
+func (b *BeaconBlock) SetForkVersion(version common.Version) {
```
calbera, line 162: not a fan of this as I see its only required in tests. If thats the case we can keep but would prefer noting that its only used for testing

consensus-types/types/block.go
abi87: I have not done much digging but IIRC `Empty` is used by SSZ code to encode/decode stuff?
How do we make sure that we are not dropping some interface support here?

abi87: we do have some quick tests in https://github.com/berachain/beacon-kit/blob/main/testing/quick/compare_test.go, comparing the way we format some data with Prysm. I wonder if we should add similar comparisons for blocks?

consensus-types/types/block.go
```diff
@@ -153,16 +148,21 @@ func (b *BeaconBlock) GetStateRoot() common.Root {
 	return b.StateRoot
 }
 
-// Version identifies the version of the BeaconBlock.
-func (b *BeaconBlock) Version() common.Version {
-	return b.BbVersion
-}
```
abi87, line 159: Also this may be useful for some SSZ related stuff?

rezbera:
moving this back into draft pending @calbera 's PR https://github.com/berachain/beacon-kit/pull/2602

consensus-types/types/block.go
```diff
@@ -118,9 +114,18 @@ func (b *BeaconBlock) MarshalSSZ() ([]byte, error) {
...
+// empty returns an empty BeaconBlock for the given fork version.
+func (*BeaconBlock) empty(version common.Version) *BeaconBlock {
+	return &BeaconBlock{
+		Versionable: NewVersionable(version),
+		Body:        (&BeaconBlockBody{}).empty(version),
+	}
+}
+
+// NewFromSSZ creates a new BeaconBlock from SSZ format.
+func (*BeaconBlock) NewFromSSZ(buf []byte, version common.Version) (*BeaconBlock, error) {
```
rezbera, line 126: For reviewers, one of the reasons we switch from `UnmarshalSSZ` to `NewFromSSZ` is because any setup required before decoding the bytes, e.g. setting `Versionable` field, is now done within the `NewFromSSZ`, whereas previously we would have a dedicated `Empty()` function which callers had to remember to use before any Unmarshalling.

consensus-types/types/payload_header.go
```diff
@@ -70,20 +69,15 @@ type ExecutionPayloadHeader struct {
 	BlobGasUsed math.U64 `json:"blobGasUsed"`
 	// ExcessBlobGas is the amount of excess blob gas in the block.
 	ExcessBlobGas math.U64 `json:"excessBlobGas"`
-
-	// EphVersion is the fork EphVersion of the execution payload header.
-	// EphVersion must be not serialized but it's exported
-	// to allow unit tests using reflect on execution payload header.
-	// TODO: Enable once
-	// https://github.com/karalabe/ssz/pull/9/files# is merged.
-	EphVersion common.Version `json:"-"`
```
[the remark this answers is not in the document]
calbera: is now public with `Versionable`

consensus-types/types/block.go
rezbera: @abi87  see https://github.com/berachain/beacon-kit/pull/2585/files#r1999747460 .

Basically we no longer want callers to need "Awareness" of the `Empty()` function before decoding SSZ because that's an unnecessary logical separation that could easily be missed before decoding. This is solved by capturing that behaviour in `NewFromSSZ`, so now anytime you want to decode using ssz, you automatically get the `Empty()` behaviour.

calbera: "Empty" behavior now implicitly part of all used constructors. More info in PR description

consensus-types/types/block.go
rezbera: Indeed it is useful for SSZ, we have no replaced this with the 

`constraints.Versionable `json:"-"`` interface, which gives us a standard pattern for all consensus types.

consensus-types/types/payload.go
```diff
@@ -72,16 +81,11 @@ type ExecutionPayload struct {
 	BlobGasUsed math.U64 `json:"blobGasUsed"`
 	// ExcessBlobGas is the amount of excess blob gas in the block.
 	ExcessBlobGas math.U64 `json:"excessBlobGas"`
-
-	// EpVersion is the version of the execution payload.
-	// EpVersion must be not serialized, but it's exported
-	// to allow unit tests using reflect on execution payload.
-	EpVersion common.Version `json:"-"`
```
rezbera, line 79: The reflection in the `quick` tests has been replaced and was not necessary

testing/quick/execution_payload_test.go
```diff
@@ -0,0 +1,202 @@
...
+			Transactions: *(*zcommon.PayloadTransactions)(
+				unsafe.Pointer(&payload.Transactions)),
+			BlobGasUsed:   zview.Uint64View(payload.BlobGasUsed.Unwrap()),
+			ExcessBlobGas: zview.Uint64View(payload.ExcessBlobGas.Unwrap()),
+		}
+		var zWithdrawals zcommon.Withdrawals
+		for _, withdrawal := range payload.Withdrawals {
+			zWithdrawal := zcommon.Withdrawal{
+				Index:          zcommon.WithdrawalIndex(withdrawal.Index),
+				ValidatorIndex: zcommon.ValidatorIndex(withdrawal.Validator),
```
rezbera, line 186: Previously we were not testing withdrawals correctly with the quick tests. The quick tests by default would not populate the withdrawals.
When I fixed the withdrawals, it became clear that the casting we were doing before:
```go
Withdrawals: *(*zcommon.Withdrawals)(unsafe.Pointer(&payload.Withdrawals)),
```
was failing because the field names are different between `zrnt` and us. This is now fixed.

testing/quick/compare_test.go
```diff
@@ -24,84 +24,27 @@ package compare_test
...
 	zspec "github.com/protolambda/zrnt/eth2/configs"
 	ztree "github.com/protolambda/ztyp/tree"
-	zview "github.com/protolambda/ztyp/view"
 	pprim "github.com/prysmaticlabs/prysm/v5/consensus-types/primitives"
 	pethpb "github.com/prysmaticlabs/prysm/v5/proto/prysm/v1alpha1"
 )
 
 var (
-	c    = quick.Config{MaxCount: 5_000}
+	c    = quick.Config{MaxCount: 10_000}
```
rezbera, line 38: quick tests are actually very fast if we do them correctly with custom generator.

testing/quick/compare_test.go
```diff
@@ -24,84 +24,27 @@ package compare_test
...
-		// nil-pointer dereference in fastssz
-		if payload == nil ||
-			payload.Withdrawals == nil ||
-			slices.Contains(payload.Withdrawals, nil) ||
-			payload.Transactions == nil ||
-			slices.ContainsFunc(payload.Transactions, func(e []byte) bool {
-				return e == nil
-			}) {
-			return true
-		}
```
rezbera, line 63: Most iterations were skipped from these checks.

state-transition/core/core_test.go
```diff
@@ -90,29 +90,31 @@ func buildNextBlock(
...
-	// finally build the block
-	blk, err := types.NewBeaconBlockWithVersion(
-		parentBlkHeader.GetSlot()+1,
-		parentBlkHeader.GetProposerIndex(),
-		parentBlkHeader.HashTreeRoot(),
-		version.Deneb1(),
-	)
 	require.NoError(t, err)
 	blk.Body = &types.BeaconBlockBody{
+		Versionable:      blk,
```
calbera, line 117: technically now, it is an anti-pattern to use the "parent" struct as the `Versionable` now. Instead each type should have its own "base" `versionable` struct. Like define it once and set it in the `ExecutionPayload` and `BeaconBlockBody`. 

Now normally  `NewBeaconBlockWithVersion` already sets the Versionable field correctly, but since in this test we are overwriting the `ExecutionPayload` and `BeaconBlockBody` structs entirely, it becomes necessary to set explicitly.

consensus-types/types/signed_beacon_block.go
```diff
@@ -34,13 +32,13 @@ import (
...
-	_ ssz.DynamicObject                   = (*SignedBeaconBlock)(nil)
-	_ constraints.SSZMarshallableRootable = (*SignedBeaconBlock)(nil)
+	_ ssz.DynamicObject                                                = (*SignedBeaconBlock)(nil)
+	_ constraints.SSZVersionedMarshallableRootable[*SignedBeaconBlock] = (*SignedBeaconBlock)(nil)
 )
 
 type SignedBeaconBlock struct {
-	Message   *BeaconBlock        `json:"message"`
-	Signature crypto.BLSSignature `json:"signature"`
+	*BeaconBlock `json:"message"`
```
calbera, line 40: question: can we also change the json tag to something like `json:"beacon_block"` or will that break backwards compatibility?

calbera: my theory is that we don't actually use the json tags since this type is never json (un)marshalled in practice. Should investigate this and if confirmed, I think we should remove the tags entirely to make this explicitly clear.

calbera: confirmed and handled in https://github.com/berachain/beacon-kit/pull/2611

abi87:
@rezbera nit: a quick search for `Empty()` shows that `func (bs *BlobSidecars) Empty() *BlobSidecars` is still there and looks unused. You may want to drop it so that, as you state in the PR heading, we don't have `Empty()` methods anymore

consensus-types/types/attestation_data.go
```diff
@@ -79,9 +79,10 @@ func (a *AttestationData) MarshalSSZ() ([]byte, error) {
 	return buf, ssz.EncodeToBytes(buf, a)
 }
 
-// UnmarshalSSZ unmarshals the AttestationData object from SSZ format.
-func (a *AttestationData) UnmarshalSSZ(buf []byte) error {
-	return ssz.DecodeFromBytes(buf, a)
+// NewFromSSZ creates a new AttestationData object from SSZ format.
+func (*AttestationData) NewFromSSZ(buf []byte) (*AttestationData, error) {
+	a := &AttestationData{}
+	return a, ssz.DecodeFromBytes(buf, a)
```
abi87, line 85: this makes an extra copy of `AttestationData` and does not initialize the pointer receiver.
Can you explain why we changed this?

abi87: So actually this is what is done in others NewFromSSZ methods, which we didn't modify.
It should be fine, but I will recheck to see if there are any unnecessary copies

consensus-types/types/block.go
abi87: Awesome

calbera:
> @rezbera, @calbera nit: a quick search for `Empty()` shows that `func (bs *BlobSidecars) Empty() *BlobSidecars` is still there and looks unused. You may want to drop it so that, as you state in the PR heading, we don't have `Empty()` methods anymore

Yup, removed in https://github.com/berachain/beacon-kit/pull/2611

consensus-types/types/payload_header.go
```diff
@@ -59,7 +66,7 @@ type ExecutionPayloadHeader struct {
 	// ExtraData is the extra data of the block.
 	ExtraData bytes.Bytes `json:"extraData"`
 	// BaseFeePerGas is the base fee per gas.
-	BaseFeePerGas *uint256.Int `json:"baseFeePerGas"`
+	BaseFeePerGas *math.U256 `json:"baseFeePerGas"`
```
abi87, line 69: I see `math.U256` is an alias for `uint256.Int` so formatting should not have changed here

primitives/constraints/ssz.go
```diff
@@ -0,0 +1,83 @@
...
+
+// SSZUnmarshaler is an interface for objects that can be unmarshaled from SSZ format.
+type SSZUnmarshaler[SelfT any] interface {
+	// NewFromSSZ creates a new object from SSZ format.
+	NewFromSSZ(bz []byte) (SelfT, error)
+}
+
+// SSZVersionedUnmarshaler is an interface for objects that can be
+// unmarshaled from SSZ format for specific versions.
+type SSZVersionedUnmarshaler[SelfT any] interface {
```
abi87, line 42: nit: `SSZMarshaler`, `SSZUnmarshaler` and `SSZVersionedUnmarshaler` could be unexported as they are only used in larger interfaces defined below

primitives/constraints/ssz.go
```diff
@@ -0,0 +1,83 @@
...
+// marshaled to SSZ format.
+type SSZMarshaler interface {
+	// MarshalSSZ marshals the object into SSZ format.
+	MarshalSSZ() ([]byte, error)
+}
+
+// SSZUnmarshaler is an interface for objects that can be unmarshaled from SSZ format.
+type SSZUnmarshaler[SelfT any] interface {
+	// NewFromSSZ creates a new object from SSZ format.
+	NewFromSSZ(bz []byte) (SelfT, error)
```
abi87, line 37: just a note, feel free to neglect this.
Looks like practically we always call `NewFromSSZ`. It would be great to make a templated constructor out of this, dropping the pointer receiver.
I can't recall rn if golang supports generics specialization so not sure if this is feasible.
My whole point is that the pointer receiver seems to exist only to allow us picking the right constructor

consensus-types/types/attestation_data.go
calbera: With the latest approach in this PR, no new memory is created unnecessarily

primitives/constraints/ssz.go
calbera: as discussed this is necessary for the generic (en/decoders) [here](https://github.com/berachain/beacon-kit/blob/d6ad0abe380b1f89d58e6d3c29437e1fcac48180/storage/encoding/ssz.go#L42) and [here](https://github.com/berachain/beacon-kit/blob/d6ad0abe380b1f89d58e6d3c29437e1fcac48180/storage/encoding/ssz.go#L93). To completely remove the generics we could build a en/decoder for the cosmos store for each type that needs it (only like 6 or 7 of them). Open to it, but should be subsequent PR

shotes:
How does the addition of new fields fit in with this new `Versionable` interface? For example, in Pectra, the `BeaconBlockBody` has new fields. Are there going to be modifications to the SSZ definition based on the `Versionable` version?

rezbera:
> How does the addition of new fields fit in with this new `Versionable` interface? For example, in Pectra, the `BeaconBlockBody` has new fields. Are there going to be modifications to the SSZ definition based on the `Versionable` version?

Correct, e.g. https://github.com/berachain/beacon-kit/pull/2599/files#diff-1cac05c0290006696b174f28b6c116ddcc9ddbd69e437a4f71c9ec1b45b2c7b0R115

shotes:
utACK

calbera (approved):

primitives/constraints/ssz.go
abi87: Indeed done with two PRs https://github.com/berachain/beacon-kit/pull/2615 and https://github.com/berachain/beacon-kit/pull/2616

PR #2590, opened 2025-03-11
chore(types): Delete excessive types that are unused and define interfaces locally

rezbera:
# Context
Things like interfaces should only be defined by the user generally speaking as not all users will require the full implementation of a concrete type. 

In BeaconKit, we overly define them with no one using them. It's better to delete them now and add them as necessary which will be part of my goal with Pectra. 

For example, **now** `beacon/blockchain/types.go` only references interfaces from other packages that it requires, whereas before it was mixing interfaces for its own package and those it required from other packages. 

We should have a clean pattern for declaring the interfaces required from other packages. My suggestion is we have a file in the package called `interfaces.go` which declares all the interfaces it depends on. 

# Changes
1. We remove all (that I could find) unused types/interfaces and methods.
2. We renamed any `types.go` files that were used for defining interfaces locally to `interfaces.go`. Any `types.go` files that were used for declaring concrete types were left as `types.go`. 
3. We modify or add `interface.go` in each package where relevant and appropriately define the reduced scope needed for the package, mostly for chain spec.


primitives/constraints/basic.go
```diff
@@ -22,28 +22,6 @@ package constraints
...
-type EngineType[SelfT any] interface {
-	EmptyWithVersion[SelfT]
-	Versionable
-	Nillable
-	JSONMarshallable
-}
-
-// EmptyWithVersion is a constraint that requires a type to have an Empty
-// method.
-type EmptyWithVersion[SelfT any] interface {
```
calbera, line 43: we use the Empty interface but not this? almost feel like we should choose EmptyWithVersion and enforce eberywhere? although that stuff may not be for this PR

node-api/backend/interfaces.go
```diff
@@ -18,23 +18,22 @@
...
-		common.ExecutionAddress,
-		engineprimitives.Withdrawals,
-		common.Root,
-	) (SelfT, error)
+type StateProcessor interface {
+	ProcessSlots(*statedb.StateDB, math.Slot) (transition.ValidatorUpdates, error)
+}
+
+type ChainSpec interface {
+	chain.BlobSpec
```
calbera, line 35: whats the point of still importing the interface here? should we only define the funcs we need from BlobSpec here?

primitives/constraints/basic.go
rezbera: > we use the Empty interface but not this? 

yea no idea why not. just that it works without it

node-api/backend/interfaces.go
rezbera: true, thought we were using all of them. fixed.

consensus-types/types/mocks/proposer_domain.mock.go
```diff
@@ -0,0 +1,82 @@
+// Code generated by mockery v2.49.0. DO NOT EDIT.
```
rezbera, line 1: was needed to keep `make generate-check` happy

storage/pruner/types.go
```diff
@@ -1,46 +0,0 @@
...
-// OTHER DEALINGS IN THE SOFTWARE.
-
-package pruner
-
-import (
-	"context"
-	"errors"
-)
-
-var ErrInvalidRange = errors.New("range start greater than end")
```
rezbera, line 33: This package `pruner` was only being used for the Error Type here. Deleted the package and moved the error declaration to [storage/errors.go](https://github.com/berachain/beacon-kit/pull/2590/files#diff-adcb777be092bdabf23d40c719392ce1c44a9a1259f713c4a96adba6d7aa7b31)

da/blob/interfaces.go
```diff
@@ -0,0 +1,35 @@
...
+// TITLE.
+
+package blob
+
+import (
+	"time"
+)
+
+// ChainSpec represents a chain spec. Currently unused but may be used with new forks
+type ChainSpec interface{}
```
calbera, line 28: do we need this for now? define as its needed imo

rezbera: ChainSpec is embedded throughout DA/blob package but never used. I can remove it but slightly more invasive, not fussed.

shotes: There used to be chainspec checks here but they got moved out to other package. I'd be cool with removing this ChainSpec from this `da/blob/interfaces.go` and `da/store/interfaces.go`, since they're never used. Up to you whether that belongs in this PR or not.

primitives/constraints/basic.go
calbera: fine for now but think we need to standardize as part of the SSZ interfaces/types cleanup

da/blob/interfaces.go
calbera: removed here https://github.com/berachain/beacon-kit/pull/2595/files

da/store/interfaces.go
```diff
@@ -35,3 +35,6 @@ type IndexDB interface {
 	// returned with no error.
 	GetByIndex(index uint64) ([][]byte, error)
 }
+
+// ChainSpec currently unused but may be used with new forks.
+type ChainSpec interface{}
```
shotes, line 40: This one can be removed as well.

shotes: https://github.com/berachain/beacon-kit/pull/2596

calbera (approved):

PR #2591, opened 2025-03-11
chore(network): add default cometbft values for p2p

chuck-bear:


abi87:
Still experimental, will merge once testing is complete

PR #2592, opened 2025-03-11
chore(ssz): Define SSZ offset/content once for safe reuse

calbera:
Resolves the TODO around `Transactions` type. Relevant now as we refactor types for multiple fork version handling.

```golang
// TODO: This can accidentally decouple from the definition in
// ExecutionPayload and we should be cognizant of that and/or
// make a PR to allow for them to be defined in one place.
```


calbera:
not feasible

PR #2599, opened 2025-03-13
feat(pectra): Pectra Consensus Types (1/n)

rezbera:
Merged after https://github.com/berachain/beacon-kit/pull/2616

This introduces they consensus types required for Pectra


consensus-types/types/execution_requests.go
```diff
@@ -0,0 +1,197 @@
...
+const dynamicFieldsInExecutionRequests = 3 // 3 since three dynamic objects (Deposits, Withdrawals, Consolidations)
+
+type ExecutionRequests struct {
+	Deposits       []*DepositRequest
+	Withdrawals    []*WithdrawalRequest
+	Consolidations []*ConsolidationRequest
+}
+
+// DepositRequest is introduced in EIP6110 which is currently not processed.
+type DepositRequest struct {
```
calbera, line 51: rather than reintroducing should we type alias this to the existing Deposit?

consensus-types/types/execution_requests.go
```diff
@@ -0,0 +1,197 @@
...
+const sszDynamicObjectOffset = 4
+const maxDepositRequestsPerPayload = 8192
+const maxWithdrawalRequestsPerPayload = 16
+const maxConsolidationRequestsPerPayload = 2
+const sszDepositRequestSize = 192          // Pubkey = 48, WithdrawalCredentials = 32, Amount = 8, Signature = 96, Index = 8.
+const sszWithdrawRequestSize = 76          // ExecutionAddress = 20, ValidatorPubKey = 48, Amount = 8
+const sszConsolidationRequestSize = 116    // ExecutionAddress = 20, PubKey = 48, Pubkey = 48
+const dynamicFieldsInExecutionRequests = 3 // 3 since three dynamic objects (Deposits, Withdrawals, Consolidations)
+
+type ExecutionRequests struct {
```
calbera, line 44: do we need JSON tags?

consensus-types/types/execution_requests.go
rezbera: Sounds reasonable. Only concern would be a potential future divergence but in the spirit of not overengineering for the future, let's do it

consensus-types/types/execution_requests.go
rezbera: is it only for APIs? is there a spec for the tags?

consensus-types/types/payload_requests.go
```diff
@@ -32,6 +32,7 @@ import (
 
 // NewPayloadRequest as per the Ethereum 2.0 specification:
 // https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/beacon-chain.md#modified-newpayloadrequest
+// TODO(pectra): Modified NewPayloadRequest: new execution_requests. This may need to be Versionable.
```
calbera, line 35: it should inherit version from ExecutionPayload ?

rezbera:
Moving back to draft so I can rebase  and cleanup

rezbera:
replaced by https://github.com/berachain/beacon-kit/pull/2599

PR #2601, opened 2025-03-13
feat(node-api): genesis api 

nidhi-singh02:
Implementation of https://ethereum.github.io/beacon-APIs/?urls.primaryName=v3.0.0#/Beacon/getGenesis along with E2E test.


abi87 (dismissed):
I would avoid modifying beacon state for this.
If not possible, we need to coordinate the upgrade

state-transition/core/state_processor_genesis.go
```diff
@@ -120,6 +120,11 @@ func (sp *StateProcessor) InitializePreminedBeaconStateFromEth1(
 		return nil, err
 	}
 
+	genesisTime := execPayloadHeader.GetTimestamp()
+	if err = st.SetGenesisTime(genesisTime); err != nil {
+		return nil, err
+	}
+
```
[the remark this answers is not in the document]
nidhi-singh02: Got it. Let me check for other alternatives here.

cli/commands/genesis/payload.go
```diff
@@ -85,6 +85,10 @@ func AddExecutionPayload(chainSpec ChainSpec, elGenesisPath string, config *cmtc
 		return errors.Wrap(err, "failed to read genesis doc from file")
 	}
 
+	//#nosec:G115 // genesis time won't ever be negative i.e. before Unix epoch (1970-01-01)
+	// Set the genesis time from app Genesis in the execution payload which gets converted to ExecutionPayloadHeader.
+	payload.Timestamp = uint64(appGenesis.GenesisTime.Unix())
+
```
nidhi-singh02, line 91: This is not related to the node-api. This sets the timestamp in the execution payload header in the beacon genesis file.

node-api/handlers/beacon/genesis.go
```diff
@@ -35,11 +37,23 @@ func (h *Handler) GetGenesis(_ handlers.Context) (any, error) {
...
+
+	genesisTime, err := h.backend.GenesisTime(utils.Genesis)
+	if err != nil {
+		return nil, err
+	}
+
 	return beacontypes.GenesisResponse{
 		Data: beacontypes.GenesisData{
-			GenesisTime:           "1590832934", // stub
+			GenesisTime:           genesisTime.Base10(),
```
nidhi-singh02, line 54: need genesis time in string format.

state-transition/core/state_processor_genesis.go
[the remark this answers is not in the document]
nidhi-singh02: Tried an alternative approach as coded in the current PR, without changing the state-transition package. Getting the timestamp from the execution header from kvstore.

testing/e2e/e2e_api_test.go
```diff
@@ -867,3 +867,28 @@ func (s *BeaconKitE2ESuite) TestGetValidatorsWithStateGenesis() {
...
+
+	genesisResp, err := client.Genesis(s.Ctx(), &beaconapi.GenesisOpts{})
+	s.Require().NoError(err)
+	s.Require().NotNil(genesisResp)
+	s.Require().NotEmpty(genesisResp.Data)
+
+	genesis := genesisResp.Data
+
+	s.Require().NotZero(genesis.GenesisTime, "Genesis time should not be zero")
+	s.Require().True(genesis.GenesisTime.Unix() < time.Now().Unix(), "Genesis time should not be in future")
```
abi87, line 883: nit, message could be `Genesis time should be in the past`.

testing/e2e/e2e_api_test.go
```diff
@@ -867,3 +867,28 @@ func (s *BeaconKitE2ESuite) TestGetValidatorsWithStateGenesis() {
...
+
+	s.Require().NotEmpty(genesis.GenesisValidatorsRoot, "Genesis validators root should not be empty")
+	s.Require().NotEmpty(genesis.GenesisForkVersion, "Genesis fork version should not be empty")
+
+	expectedVersion := phase0.Version{0x04, 0x00, 0x00, 0x00}
```
abi87, line 888: nit: use `version.Phase0()` instead

node-api/handlers/beacon/genesis.go
```diff
@@ -35,11 +37,23 @@ func (h *Handler) GetGenesis(_ handlers.Context) (any, error) {
 	if len(genesisRoot) == 0 {
 		return nil, types.ErrNotFound
 	}
+
+	genesisForkVersion, err := h.backend.GenesisForkVersion(utils.Genesis)
+	if err != nil {
+		return nil, err
+	}
+	genesisForkVersionHex := "0x" + hex.EncodeToString(genesisForkVersion[:])
+
```
abi87, line 46: I wonder if we create a canonical way to represent fork versions? We should check what ethereum does
cc @calbera

node-api/backend/genesis.go
```diff
@@ -35,3 +35,31 @@ func (b Backend) GenesisValidatorsRoot(slot math.Slot) (common.Root, error) {
...
+		return 0, errors.Wrapf(err, "failed to get state from slot %d", genesisSlot)
+	}
+	// Get the execution payload header from the beacon state,
+	// and return the timestamp as the genesis time.
+	execPayloadHeader, err := st.GetLatestExecutionPayloadHeader()
+	if err != nil {
+		return 0, errors.Wrapf(err, "failed to get execution payload header from state")
+	}
+	return execPayloadHeader.Timestamp, nil
+}
```
abi87, line 65: This is probably the best we can do and it does not look too bad.
I think we should:
- compare what other API implementation do in Ethereum
- reach agreement about this and document it. cc @calbera

abi87 (changes requested):
Some nits and a couple of questions to be agreed

testing/e2e/e2e_api_test.go
nidhi-singh02: Good suggestion, done!

testing/e2e/e2e_api_test.go
nidhi-singh02: done!

node-api/handlers/beacon/genesis.go
calbera: From the spec they have returned
```
     "genesis_fork_version": "0x00000000"
```

cli/commands/genesis/payload.go
calbera: This change can be removed. Since the `payload.Timestamp` is already set and determined by the eth-genesis file. If we overwrite it here (and the eth-genesis timestamp is actually different from the cl-genesis timestamp), we will produce a different execution payloaad header and people will have issues. Let's leave this CLI as is.

node-api/backend/genesis.go
```diff
@@ -35,3 +35,31 @@ func (b Backend) GenesisValidatorsRoot(slot math.Slot) (common.Root, error) {
 	}
 	return st.GetGenesisValidatorsRoot()
 }
+
+// GenesisForkVersion returns the genesis fork version of the beacon chain.
+func (b Backend) GenesisForkVersion(genesisSlot math.Slot) (common.Version, error) {
```
calbera, line 40: nit: all backend methods should have ptr receiver. Please change in another PR

node-api/backend/genesis.go
```diff
@@ -35,3 +35,31 @@ func (b Backend) GenesisValidatorsRoot(slot math.Slot) (common.Root, error) {
...
+	st, _, err := b.stateFromSlot(genesisSlot)
+	if err != nil {
+		return common.Version{}, errors.Wrapf(err, "failed to get state from slot %d", genesisSlot)
+	}
+	fork, err := st.GetFork()
+	if err != nil {
+		return common.Version{}, errors.Wrapf(err, "failed to get fork from state")
+	}
+	return fork.CurrentVersion, nil
```
calbera, line 49: note: this is not the convention. This whole function should just be `ActiveForkVersionAtSlot(slot math.Slot)` and inside it should just call the chain spec method. No need to access state. Then in genesis handler, call with slot == 0.

node-api/backend/genesis.go
calbera: Prysm stores the genesis time (and other misc info) in memory in various non-state enforced "helper" services

node-api/handlers/beacon/genesis.go
calbera: For us, similarly we should return `"0x04000000"`

node-api/backend/genesis.go
calbera: and this function on `Backend`: `ActiveForkVersionAtSlot` should not be in genesis.go, but probably backend.go

node-api/backend/genesis.go
calbera: 2nd thing is we cannot query stateFromSlot with slot == 0 because comet's QueryContext function assumes height of 0 to be the head of the chain.

calbera: So for Genesis Timestamp which is not stored in app-enforced state anywhere, we should load it into memory in the Backend upon genesis/initialization.

node-api/backend/genesis.go
```diff
@@ -35,3 +35,31 @@ func (b Backend) GenesisValidatorsRoot(slot math.Slot) (common.Root, error) {
 	}
 	return st.GetGenesisValidatorsRoot()
```
calbera, line 36: This function should be redone.
- does not need to take in a slot
- should only read from state if the value is not already cached.
- if not cached, call stateFromSlot(0) and explicitly comment that we are reading from the state at the tip of the chain.
- cache and return st.GetGenesisValidatorsRoot()

This is okay since we actually store the genesis value (and remains unmodified) in the app-enforced state.

node-api/handlers/beacon/genesis.go
```diff
@@ -35,11 +37,23 @@ func (h *Handler) GetGenesis(_ handlers.Context) (any, error) {
 	if len(genesisRoot) == 0 {
 		return nil, types.ErrNotFound
 	}
+
+	genesisForkVersion, err := h.backend.GenesisForkVersion(utils.Genesis)
+	if err != nil {
+		return nil, err
+	}
+	genesisForkVersionHex := "0x" + hex.EncodeToString(genesisForkVersion[:])
```
calbera, line 45: this isn't needed just return `genesisForkVersion.String()`. The type already handles this.

node-api/handlers/beacon/genesis.go
```diff
@@ -35,11 +37,23 @@ func (h *Handler) GetGenesis(_ handlers.Context) (any, error) {
...
+		return nil, err
+	}
+
 	return beacontypes.GenesisResponse{
 		Data: beacontypes.GenesisData{
-			GenesisTime:           "1590832934", // stub
+			GenesisTime:           genesisTime.Base10(),
 			GenesisValidatorsRoot: genesisRoot,
-			GenesisForkVersion:    "0x00000000", // stub
+			GenesisForkVersion:    genesisForkVersionHex,
```
calbera, line 56:
```suggestion
			GenesisForkVersion:    genesisForkVersion.String(),
```

calbera (dismissed):
some changes

node-api/handlers/beacon/genesis.go
nidhi-singh02: Thanks, done

node-api/handlers/beacon/genesis.go
nidhi-singh02: yeah, done

node-api/backend/genesis.go
nidhi-singh02: addressed in [commit](https://github.com/berachain/beacon-kit/pull/2601/commits/1c56d57d0e7b4e35deac6135bd840dfa764a5933)

node-api/backend/genesis.go
nidhi-singh02: Created an issue https://github.com/berachain/beacon-kit/issues/2614 to track it. Will pick in separate PR

cli/commands/genesis/payload.go
nidhi-singh02: understood, removed the cli change.

node-api/handlers/beacon/genesis.go
nidhi-singh02: Yes, this is the response for us `"genesis_fork_version": "0x04000000"`

node-api/backend/genesis.go
nidhi-singh02: I understood the 1st and 2nd pointer, Was trying to set genesis time in `init_chain.go`  in ProcessGenesisData , noticed that the ExecutionPayloadHeader does not have the timestamp. Might not  be the correct place.
Can you guide me from where to load it. DMed you the same. Thanks!

node-api/backend/genesis.go
```diff
@@ -26,12 +26,53 @@ import (
...
-func (b Backend) GenesisValidatorsRoot(slot math.Slot) (common.Root, error) {
-	// needs genesis_time and genesis_fork_version
-	st, _, err := b.stateFromSlot(slot)
+// GetGenesisValidatorsRoot returns the genesis validators root of the beacon chain.
+func (b Backend) GenesisValidatorsRoot() (common.Root, error) {
+	// Return cached value if available
+	if b.genesisValidatorsRoot != (common.Root{}) {
+		return b.genesisValidatorsRoot, nil
+	}
+	// If not cached, read from the state of the tip of the chain
```
fridrik01, line 35:
```suggestion
	// If not cached, read state from the genesis slot
```

node-api/backend/genesis.go
```diff
@@ -26,12 +26,53 @@ import (
...
+		return common.Root{}, errors.Wrapf(err, "failed to get state from tip of chain")
+	}
+	// Get the genesis validators root
+	root, err := st.GetGenesisValidatorsRoot()
+	if err != nil {
+		return common.Root{}, errors.Wrap(err, "failed to get genesis validators root from state")
+	}
+
+	// Cache the value for future use
+	b.genesisValidatorsRoot = root
```
fridrik01, line 47: there is a race condition here if multiple api calls were calling the GenesisValidatorsRoot at the same time

node-api/backend/genesis.go
nidhi-singh02: addressed

node-api/backend/genesis.go
nidhi-singh02: Thanks, introduced `genesisValidatorsRootMu sync.RWMutex` to handle the race condition.

node-api/backend/block.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 // BlockHeader returns the block header at the given slot.
-func (b Backend) BlockHeaderAtSlot(slot math.Slot) (*ctypes.BeaconBlockHeader, error) {
+func (b *Backend) BlockHeaderAtSlot(slot math.Slot) (*ctypes.BeaconBlockHeader, error) {
 	st, _, err := b.stateFromSlot(slot)
```
nidhi-singh02, line 33: Had to make the pointer receiver as we have mutex in Backend. Same goes for other places in the PR.

node-api/backend/genesis.go
```diff
@@ -26,12 +26,67 @@ import (
 	"github.com/berachain/beacon-kit/primitives/math"
 )
 
-// GetGenesis returns the genesis state of the beacon chain.
-func (b Backend) GenesisValidatorsRoot(slot math.Slot) (common.Root, error) {
-	// needs genesis_time and genesis_fork_version
-	st, _, err := b.stateFromSlot(slot)
+// GetGenesisValidatorsRoot returns the genesis validators root of the beacon chain.
+func (b *Backend) GenesisValidatorsRoot() (common.Root, error) {
+	// Fast path: read lock for checking cached value
```
calbera, line 31: use atomic.Pointer[common.Root] to clean up, but looks correct (thread-safe wise)

node-api/backend/genesis.go
calbera: `ActiveForkVersionForTimestamp(timestamp math.U64) common.Version`

node-api/backend/genesis.go
```diff
@@ -26,12 +26,67 @@ import (
...
+	// Slow path: write lock for initialization
+	b.genesisValidatorsRootMu.Lock()
+	defer b.genesisValidatorsRootMu.Unlock()
+
+	// Double check after acquiring write lock
+	if b.genesisValidatorsRoot != (common.Root{}) {
+		return b.genesisValidatorsRoot, nil
+	}
+
+	// If not cached, read state from the genesis slot
```
calbera, line 49: reading from genesis slot is not possible

calbera: 0 -> implies "head" in comet

nidhi-singh02: As discussed on the call, keeping it as it is - fetching the state from slot 0. As we don't have validatorsRoot value in the genesis file.

node-api/backend/genesis.go
nidhi-singh02: done!

node-api/backend/genesis.go
nidhi-singh02: Changed the approach as suggested by Cal, using atomic pointer which is more cleaner.

node-api/backend/genesis.go
nidhi-singh02: Thanks for your help @calbera. As we discussed, now we are caching the genesis time from the genesis file and storing it in backend during initialisation.

nidhi-singh02:
The E2E tests are failing with the current changes. Now getting `"genesis_fork_version": "0x04010000"` instead of `0x04000000`. I see changes been done as part of Pectra. Wanted to confirm is this the expected behaviour now @calbera 

Other than this, all review comments are addressed.

node-api/backend/block.go
nidhi-singh02: We don't have mutex anymore, still keeping the pointer changes. Will ensure it is there for every endpoint as part of a separate PR.

nidhi-singh02:
> The E2E tests are failing with the current changes. Now getting `"genesis_fork_version": "0x04010000"` instead of `0x04000000`. I see changes been done as part of Pectra. Wanted to confirm is this the expected behaviour now @calbera
> 
> Other than this, all review comments are addressed.

Changed the expected genesis version to  Deneb1 ("0x04010000") which is coming as per the specs.

calbera:
e2e tests now use Electra at genesis version to allow testing Electra features. Will switch back to Deneb pending chain spec configuration changes in e2e suite

calbera (approved):

PR #2604, opened 2025-03-14
chore(networks/bepolia): add more peers and bootnodes

chuck-bear:


KodiakBera (approved):
LGTM

rezbera (commented):
Noting that we need to should update `make start-g/reth-bepolia` to use bootnodes with this change

chuck-bear:
> Noting that we need to should update `make start-g/reth-bepolia` to use bootnodes with this change

Just noticed that geth is using peers as bootnodes.

testing/networks/80069/el-bootnodes.txt
calbera, line 1: @chuck-bear is it okay to make this a comma separated list? or was there a reason for line breaking? comma separated would be more compat with our script that reads this file

calbera (approved):

PR #2605, opened 2025-03-14
test(pectra): Pectra Phase 1 Accepance Criteria Test

rezbera:
Getting this test working is the goal of Phase 1 of the Pectra deliverable. 

This test will ensure we can support a Prague based EL.
It tests that we can process blocks that include ExecutionRequests (in this case a withdrawals). We do not have to actually process the withdrawal, but we should not fail. 

It is currently marked with t.Skip() to prevent CI failing but should be used as a reference point in branches going forward.


go.mod
```diff
@@ -57,7 +57,7 @@ require (
 // build/test dependencies
 require (
 	github.com/attestantio/go-eth2-client v0.23.0
-	github.com/ethereum/go-ethereum v1.14.13
+	github.com/ethereum/go-ethereum v1.15.5
```
rezbera, line 60: We need to bump the go-version as older versions of go-ethereum do not correctly calculate prague genesis block hashes. Specifically because because of this line. https://github.com/ethereum/go-ethereum/blob/bee38ebb392caf564c14d7822adbf28fef2f84e9/core/genesis.go#L475-L474

testing/simulated/pectra-eth-genesis.json
```diff
@@ -0,0 +1,371 @@
...
+    "istanbulBlock": 0,
+    "muirGlacierBlock": 0,
+    "berlinBlock": 0,
+    "londonBlock": 0,
+    "arrowGlacierBlock": 0,
+    "grayGlacierBlock": 0,
+    "mergeNetsplitBlock": 0,
+    "shanghaiTime": 0,
+    "cancunTime": 0,
+    "pragueTime": 0,
```
rezbera, line 22: note that prague is enabled at genesis. We should be able to support this.

testing/simulated/pectra_test.go
```diff
@@ -0,0 +1,127 @@
...
+	simulated.SharedAccessors
+}
+
+// TestSimulatedCometComponent runs the test suite.
+func TestPectraSuite(t *testing.T) {
+	suite.Run(t, new(PectraSuite))
+}
+
+// SetupTest initializes the test environment.
+func (s *PectraSuite) SetupTest() {
```
rezbera, line 52: tldr we need a new Suite because we have a different `SetupTest()` with different eth genesis path.

testing/simulated/pectra_test.go
```diff
@@ -0,0 +1,127 @@
...
+	const coreLoopIterations = 10
+
+	// Initialize the chain state.
+	s.InitializeChain(s.T())
+
+	// Retrieve the BLS signer and proposer address.
+	blsSigner := simulated.GetBlsSigner(s.HomeDir)
+
+	// Go through 1 iteration of the core loop to bypass any startup specific edge cases such as sync head on startup.
+	proposals := s.MoveChainToHeight(s.T(), blockHeight, coreLoopIterations, blsSigner)
```
rezbera, line 125: Currently, we pass PrepareProposal but fail in ProcessProposal as the BlockHash calculation does to match. This is expected because we have not included `ExecutionRequestsHash` in the calculation.

go.mod
calbera: Q: have we checked if this has any unintended effects on other imports?

testing/simulated/pectra_test.go
```diff
@@ -0,0 +1,160 @@
...
+	// create withdrawal request
+	// corresponds with funded address in genesis 0x20f33ce90a13a4b5e7697e3544c3083b8f8a51d4
+	senderKey, err := crypto.HexToECDSA("fffdbb37105441e14b0ee6330d855d8504ff39e705c3afa8f859ac9865f99306")
+	s.Require().NoError(err)
+
+	elChainID := big.NewInt(int64(s.TestNode.ChainSpec.DepositEth1ChainID()))
+	signer := types.NewPragueSigner(elChainID)
+	// Field values roughly copies from Geth
+	// https://github.com/ethereum/go-ethereum/blob/39638c81c56db2b2dfe6f51999ffd3029ee212cb/core/blockchain_test.go#L4131-L4130
+	withdrawalTx := types.MustSignNewTx(senderKey, signer, &types.DynamicFeeTx{
```
calbera, line 139: Q: in the case where there is no 7002 withdrawal tx, do we also fail? Or is it only failing bc the EL sees this then explicitly attaches a ExecutionLayerWithdrawalRequest? Can we test a block without this withdrawal tx

calbera (commented):
This is really nice! One question about adding the test case without the explicit 7002 withdrawal tx, but otherwise looks great

go.mod
rezbera: not besides tests passing. make start mainnet doesn't seem to be finding peers for me so can't test atm

testing/simulated/pectra-eth-genesis.json
calbera: Q: should we verify within the simulated pectra test that the EL fork version is indeed Prague?

go.mod
rezbera: mainnet synced to ~10k blocks

testing/simulated/pectra_test.go
rezbera: yes, we still fail without withdrawal TX for the same reason. If we have no withdrawal tx, we could just theoretically hardcode &emptyRequests hash to get past the current error

testing/simulated/pectra-eth-genesis.json
rezbera: That seems more like a test of the simulation framework - but reasonable ask. It's possible via the `admin_nodeInfo` https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-admin#admin-nodeinfo API to get the pragueBlock but the effort to reward seems ratio seems high

rezbera: I personally just sanity checked the geth logs on startup since it logs the forks

calbera (approved):
LGTM

PR #2606, opened 2025-03-16
chore(ci): Remove requirement for Dockerhub Login

rezbera:
Community PRs cannot run E2E CI because the action requires login. There is no reason we should require login as the CI action does not require any private images.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -143,12 +143,6 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
-      - name: Login to Docker Hub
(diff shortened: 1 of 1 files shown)


fridrik01:
@rezbera there is a throttling/rate limiting applied if we don't login to dockerhub, see https://docs.docker.com/docker-hub/usage/pulls/. I believe that is why we are authenticating to dockerhub in our GH workflows. Maybe we can update it to login if the secret exits, otherwise skip it so it works for external contributors?

fridrik01:
@rezbera see alternative https://github.com/berachain/beacon-kit/pull/2607 which makes logging in optional

PR #2607, opened 2025-03-17
Only login to Docker Hub for internal use

fridrik01:
An alternative to https://github.com/berachain/beacon-kit/pull/2606 

We may want to continue authenticating with Docker Hub as otherwise we risk frequent rate limiting when pulling images,see  https://docs.docker.com/docker-hub/usage/pulls/

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -143,8 +143,16 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
+      - name: Check if Docker Hub credentials exist
+        id: check-docker-credentials
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/pipeline.yml
```diff
@@ -143,8 +143,16 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
+      - name: Check if Docker Hub credentials exist
+        id: check-docker-credentials
+        run: |
+          if [ "${{ secrets.DOCKERHUB_USERNAME }}" != "" ] && [ "${{ secrets.DOCKERHUB_PASSWORD }}" != "" ]; then
+            echo "has_credentials=true" >> $GITHUB_OUTPUT
+          fi
+        shell: bash
```
fridrik01, line 152: I had to create a separate step as `if` condition does not have access to `secrets`

rezbera (approved):
nice! lgtm

PR #2610, opened 2025-03-18
try this

calbera:


consensus-types/types/signed_beacon_block_test.go
```diff
@@ -183,15 +180,20 @@ func TestSignedBeaconBlock_SizeSSZ(t *testing.T) {
 
 func TestSignedBeaconBlock_EmptySerialization(t *testing.T) {
 	t.Parallel()
-	orig := &types.SignedBeaconBlock{}
+	fv := version.Phase0()
```
rezbera, line 183: can we wrap in in runForAllSupportedVersions

calbera: had only done this since its the "empty" version, but okay with this

PR #2612, opened 2025-03-18
chore(pectra): Update Simulation Test Chain Spec (2/n)

rezbera:
After https://github.com/berachain/beacon-kit/pull/2599

SImply updates the tests so we can run beacon kit with electra genesis


testing/simulated/homedir.go
```diff
@@ -47,20 +46,13 @@ const TestnetBeaconChainID = "testnet-beacon-80069"
...
 	t.Logf("Initializing home directory: %s", tempHomeDir)
 	// Create the default CometBFT configuration using the temporary home directory.
 	cometConfig := createCometConfig(t, tempHomeDir)
 
-	// Retrieve the testnet chain specification.
-	chainSpec, err := spec.TestnetChainSpec()
-	require.NoError(t, err, "failed to retrieve testnet chain spec")
-
-	// Set the chain spec type environment variable.
-	t.Setenv(components.ChainSpecTypeEnvVar, components.TestnetChainSpecType)
```
rezbera, line 62: no longer need to do this envar stuff

testing/simulated/utils.go
```diff
@@ -121,7 +121,7 @@ func (s *SharedAccessors) MoveChainToHeight(
 			ProposerAddress: pubkey.Address(),
 		})
 		require.NoError(t, err)
-		require.NotEmpty(t, proposal)
+		require.Len(t, proposal.Txs, 2)
```
abi87, line 124: txs in proposal are at most two not exactly two? We may requesting a too strict of a invariant here

abi87 (approved):
nit, but LGTM

testing/simulated/utils.go
shotes: If there is no error, then the txs in a proposal will always be exactly two.

testing/simulated/components.go
```diff
@@ -76,3 +77,22 @@ func FixedComponents(t *testing.T) []any {
...
+	chainSpec, err := chain.NewSpec(specData)
+	if err != nil {
+		return nil, err
+	}
+	return chainSpec, nil
+}
+
+// ProvideSimulationChainSpec provides a default chain-spec equivalent to testnet.
+// Bypasses the need for environment variables.
+func ProvideSimulationChainSpec() (chain.Spec, error) {
```
shotes, line 96: Oooh, does this set the chainspec for the node as well? If so, I think I could use this in my fork by timestamp PR

PR #2615, opened 2025-03-19
chore(ssz types): replace New attributes with generic constructor

abi87:
A proposal to cleanup 
```golang
type SSZUnmarshaler[SelfT any] interface {
	// NewFromSSZ creates a new object from SSZ format.
	NewFromSSZ(bz []byte) (SelfT, error)
}
```

It seems to me that what we are trying to do here is just having a generic constructor to be use when deserializing.
The receiver pointer is only used to understand which type we are deserializing, so why not using the generic instead?
Also, if we inspect implementations of `NewFromSSZ` they are  very similar (two exception, see below)
So we can indeed use a constructor with a generic type.

** There are only two caveat **:

- UnusedType has a different way to decode bytes for efficiency. This is already done ad-hoc, so I do it ad-hoc in the generic constructor
- Sidecar does some sintax validation after decoding. This is not a bad idea in general, so why not allowing syntactic validation for all types?

This PR basically replaces the monstruosity above, with a less horrible constructor. It seems to me it also simplifies the type hyerarchy.


consensus-types/types/attestation_data.go
```diff
@@ -79,13 +79,8 @@ func (a *AttestationData) MarshalSSZ() ([]byte, error) {
 	return buf, ssz.EncodeToBytes(buf, a)
 }
 
-// NewFromSSZ creates a new AttestationData object from SSZ format.
-func (a *AttestationData) NewFromSSZ(buf []byte) (*AttestationData, error) {
-	if a == nil {
-		a = &AttestationData{}
-	}
-	return a, ssz.DecodeFromBytes(buf, a)
-}
```
abi87, line 88: This and other methods are replaced by a single, generic function, `sszconstructors.NewFromSSZ`.

Horrible name I reckon, and suggestions are welcomed.
I really really like the reduced code duplication

consensus-types/types/attestation_data_test.go
```diff
@@ -84,13 +85,13 @@ func TestAttestationData_MarshalSSZ_UnmarshalSSZ(t *testing.T) {
 			require.NoError(t, err)
 			require.NotNil(t, data)
 
-			var unmarshalled *types.AttestationData
```
abi87, line 87: no need to instantiate a pointer before creating the object via `New...`. I love this

primitives/constraints/ssz.go
```diff
@@ -32,9 +32,9 @@ type SSZMarshaler interface {
 }
 
 // SSZUnmarshaler is an interface for objects that can be unmarshaled from SSZ format.
-type SSZUnmarshaler[SelfT any] interface {
-	// NewFromSSZ creates a new object from SSZ format.
-	NewFromSSZ(bz []byte) (SelfT, error)
+type SSZUnmarshaler interface {
+	IsUnusedFromSZZ() bool      // some types are not used and we can/must shortcut their marshalling
+	EnsureSyntaxFromSSZ() error // once unmarshalled we will check whether type syntax is correct
```
abi87, line 37: this PR replaces `NewFromSSZ` with a couple of attributes used in the single `sszconstructor.NewFromSSZ`.
That constructor itself has a template like pattens so it will use these methods to handle the instantiation corner cases (right now, just unused type and post serialization checks for a single type)

consensus-types/types/deposit.go
```diff
@@ -54,6 +54,10 @@ type Deposit struct {
 	Index uint64 `json:"index"`
 }
 
+func NewEmptyDeposit() *Deposit {
+	return &Deposit{}
+}
```
abi87, line 59: Needed for kvStore to allow them creating objects to unmarshal data into.
IMHO it's better to have a few of these small constructors, rather than making New a method of a generic interface

consensus-types/types/payload_header_test.go
```diff
@@ -245,7 +245,7 @@ func TestExecutionPayloadHeader_NewFromSSZ_Invalid(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			buf := tc.malleate()
 			_, err := (&types.ExecutionPayloadHeader{}).NewFromSSZ(buf, version.Deneb())
-			require.ErrorContains(t, err, tc.expErr.Error())
```
abi87, line 248: brittle tests, I enriched an error description and they failed. `ErrorIs` is the idiomatic way to do this AFAIK

consensus-types/types/unused_type.go
```diff
@@ -58,17 +58,18 @@ func (ut *UnusedType) MarshalSSZ() ([]byte, error) {
...
+	// we special case construction of unused types, for efficiency
 	if len(buf) != 1 {
-		return nil, fmt.Errorf("expected 1 byte got %d", len(buf))
+		return fmt.Errorf("expected 1 byte got %d", len(buf))
 	}
 	//#nosec:G701 // UnusedType is uint8 and byte is uint8.
-	u := UnusedType(buf[0])
-	ut = &u
-	return ut, nil
+	*v = UnusedType(buf[0])
```
abi87, line 67: client controls allocation of objec to unmarshal data into, so we simply overwrite it

storage/beacondb/kvstore.go
```diff
@@ -115,7 +115,9 @@ func New(kss store.KVStoreService) *KVStore {
 			schemaBuilder,
 			sdkcollections.NewPrefix([]byte{keys.ForkPrefix}),
 			keys.ForkPrefixHumanReadable,
-			encoding.SSZValueCodec[*ctypes.Fork]{},
+			encoding.SSZValueCodec[*ctypes.Fork]{
+				NewEmptyF: ctypes.NewEmptyFork,
```
abi87, line 119: this is my proposal to avoid all the `Interface[Self T]` madness.
Just provide the empty constructor where needed and that will take care of instantiating the object where needed.
On the contrary with the `pointer.New` structure, we have to worry whether `pointer` is nil or not and what is the semantics of `New` (returns a different pointer? overwrites `pointer`? Both?)

Maybe not a big deal, but I fill simple things should be simple

storage/encoding/ssz.go
```diff
@@ -23,23 +23,28 @@ package encoding
...
 )
 
 // SSZValueCodec provides methods to encode and decode SSZ values.
 type SSZValueCodec[T interface {
-	constraints.SSZMarshallable[T]
-}] struct{}
+	ssz.StaticObject
+	constraints.SSZMarshallable
+}] struct {
+	NewEmptyF func() T // constructor
```
abi87, line 36: we pass constructors (and generic ones) to the codec to give it the template to decode object

da/types/sidecar.go
```diff
@@ -159,25 +159,17 @@ func (b *BlobSidecar) MarshalSSZ() ([]byte, error) {
...
-
-// NewFromSSZ creates a new BlobSidecar object from SSZ format.
-func (b *BlobSidecar) NewFromSSZ(buf []byte) (*BlobSidecar, error) {
-	b = b.empty()
-	if err := ssz.DecodeFromBytes(buf, b); err != nil {
-		return nil, err
+func (b *BlobSidecar) EnsureSyntaxFromSSZ() error {
+	// Ensure SignedBeaconBlockHeader is not nil
+	if b.SignedBeaconBlockHeader == nil {
+		b.SignedBeaconBlockHeader = &ctypes.SignedBeaconBlockHeader{}
```
abi87, line 165: not sure this is needed anymore, but keeping it for good measure.
This is the reason these methods are called `Ensure` instead of just `Check`. I intend to signal that they can modify the pointer in place, not just read from it and to checks

consensus-types/types/attestation_data.go
```diff
@@ -79,13 +79,7 @@ func (a *AttestationData) MarshalSSZ() ([]byte, error) {
...
+func (*AttestationData) EnsureSyntaxFromSSZ() error { return nil }
```
fridrik01, line 82: Consider renaming this to sth more descriptive, like `ValidateAfterDecodingSSZ`

consensus-types/types/deposit_test.go
```diff
@@ -100,10 +101,9 @@ func TestDeposit_MarshalUnmarshalSSZ(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, sszDeposit)
 
-	var unmarshalledDeposit *types.Deposit
-	unmarshalledDeposit, err = unmarshalledDeposit.NewFromSSZ(sszDeposit)
+	unmarshalledDeposit := new(types.Deposit)
```
fridrik01, line 104: Does the `decoder.SSZUnmarshal` implementation require a pointer receiver? Like, it wouldn't work with a reference to a struct?... I think so, but I wanted to double check

primitives/decoder/decoder.go
```diff
@@ -0,0 +1,47 @@
...
+// SSZUnmarshal is the way we build objects from byte formatted as ssz
+// While logically related to constraints package, SSZUnmarshal has its own
+// small package to avoid import cycle related to Unused Type
+// Also SSZUnmarshal highlight the common template for SSZ decoding different
+// objects
+func SSZUnmarshal[T constraints.SSZUnmarshaler](buf []byte, v T) error {
+	switch dest := any(v).(type) {
+	case *types.UnusedType:
+		// unused types have special formatting for efficiency
+		return types.DecodeUnusedType(buf, dest)
```
fridrik01, line 40: A bit confused by this, isn't this a noop and not changing `v` at all?

storage/beacondb/index/validator.go
```diff
@@ -39,8 +39,8 @@ const (
 
 // Validator is an interface that combines the ssz.Marshaler and
 // ssz.Unmarshaler interfaces.
-type Validator[SelfT any] interface {
-	constraints.SSZMarshallable[SelfT]
+type Validator interface {
+	constraints.SSZMarshallable
```
fridrik01, line 43: this is now much cleaner!

fridrik01 (commented):
Overall this looks like a solid improvement and makes the code more readable, had just a few questions/nits.

consensus-types/types/attestation_data.go
rezbera: Agree on name change - not immediately clear from current name

consensus-types/types/eth1data.go
```diff
@@ -56,6 +56,10 @@ func NewEth1Data(depositRoot common.Root) *Eth1Data {
 	}
 }
 
+func NewEmptyEthi1Data() *Eth1Data {
```
rezbera, line 59: typo -> `NewEmptyEth1Data`

primitives/decoder/decoder.go
```diff
@@ -0,0 +1,47 @@
...
+	default:
+		if err := ssz.DecodeFromBytes(buf, v); err != nil {
+			return fmt.Errorf("failed decoding %T: %w", dest, err)
+		}
+		return v.EnsureSyntaxFromSSZ()
```
rezbera, line 45: noting that if `v.EnsureSyntaxFromSSZ()` returns error, `v` has still been mutated. This _should_ be okay though.

rezbera (commented):
some nits, otherwise lgtm

primitives/decoder/decoder.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/berachain/beacon-kit/primitives/constraints"
+	"github.com/karalabe/ssz"
+)
+
+// SSZUnmarshal is the way we build objects from byte formatted as ssz
+// While logically related to constraints package, SSZUnmarshal has its own
+// small package to avoid import cycle related to Unused Type
+// Also SSZUnmarshal highlight the common template for SSZ decoding different
+// objects
+func SSZUnmarshal[T constraints.SSZUnmarshaler](buf []byte, v T) error {
```
rezbera, line 36: we should declare interfaces in the package locally

primitives/decoder/decoder.go
```diff
@@ -0,0 +1,47 @@
...
+// EXPRESS OR IMPLIED, INCLUDING (WITHOUT LIMITATION) WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, NON-INFRINGEMENT, AND
+// TITLE.
+
+package decoder
+
+import (
+	"fmt"
+
+	"github.com/berachain/beacon-kit/consensus-types/types"
```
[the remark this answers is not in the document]
rezbera: https://github.com/berachain/beacon-kit/pull/2617

primitives/decoder/decoder.go
rezbera: https://github.com/berachain/beacon-kit/pull/2617

consensus-types/types/unused_type.go
shotes: Usually prefer non-nil check and error, but `ssz.Decode` also has this behavior, and it is more of a dev usage error rather than a bug in data encoding/decoding. Perhaps this panic behavior is good and enforces correct usage as long as we have coverage.

primitives/decoder/decoder.go
shotes, line 40: Should we be making a special case for UnusedType in the name of efficiency? Is there actually a significant performance improvement here?

consensus-types/types/attestation_data.go
abi87: Done

consensus-types/types/eth1data.go
abi87: Done

primitives/decoder/decoder.go
abi87: not sure it is a noop, but it is needed to switch over the type. 
I can't do `if v.(*types.UnusedType) cause v could have a concrete type and not being an interface

abi87: @shotes  I would love to remove it and use the ssz decoder, but how do I know if I won't change the serialization format?
If you think this can be done, please open a PR on top and let's streamline it

consensus-types/types/unused_type.go
abi87: Exactly, I avoided the check to keep uniform behaviour with ssz decoder

primitives/decoder/decoder.go
abi87: This is an important observation, that it is worth noting in the code.
Also I rechecked the implementation and sligthly changed `BlobSidecar` to err without changing v.
I don't think we can enforce it here (missing C++ `const`s now) but the comment should warn us of not making assumptions (although the new name `Validate...` would suggest that it's an read-only method?)

consensus-types/types/deposit_test.go
abi87: I does work as reference to the struct. I just wanted to keep the pointer type, avoid `&`.
Also note that `require.Equal` fails if you compare pointer to struct, even if they have the same content

rezbera (approved):
Nice - lgtm

primitives/decoder/decoder.go
shotes: https://github.com/berachain/beacon-kit/pull/2624

PR #2616, opened 2025-03-20
chore(ssz types): replace NewVersioned attributes with generic constructor

abi87:
Part 2 of https://github.com/berachain/beacon-kit/pull/2615 where we cleanup Versioned types.
The nice bit is that I don't need an extra `decoder.SSZUnmarshal`.

It is possible to cleanup UTs some more but stopping here as a PoC


consensus-types/types/payload.go
```diff
@@ -168,25 +181,12 @@ func (p *ExecutionPayload) MarshalSSZ() ([]byte, error) {
...
-	ep := &ExecutionPayload{
-		Versionable:   NewVersionable(forkVersion),
-		BaseFeePerGas: &math.U256{},
-	}
-
+func (p *ExecutionPayload) EnsureSyntaxFromSSZ() error {
 	// For any fork version after Bellatrix (Capella onwards), non-nil withdrawals are required.
-	if version.IsAfter(forkVersion, version.Bellatrix()) {
-		ep.Withdrawals = make([]*engineprimitives.Withdrawal, 0)
+	if p.Withdrawals == nil && version.IsAfter(p.GetForkVersion(), version.Bellatrix()) {
```
abi87, line 186: p.Withdrawals == nil because this is done post serialization

consensus/cometbft/service/encoding/encoding.go
```diff
@@ -81,7 +83,14 @@ func UnmarshalBeaconBlockFromABCIRequest(
 		return signedBlk, ErrNilBeaconBlockInRequest
 	}
 
-	return ctypes.NewSignedBeaconBlockFromSSZ(blkBz, forkVersion)
```
abi87, line 84: similarly to what we do with blobs, we use decoder instead of yet another constructor

primitives/constraints/ssz.go
```diff
@@ -38,13 +38,6 @@ type SSZUnmarshaler interface {
 	EnsureSyntaxFromSSZ() error // once unmarshalled we will check whether type syntax is correct
 }
 
-// sszVersionedUnmarshaler is an interface for objects that can be
-// unmarshaled from SSZ format for specific versions.
-type sszVersionedUnmarshaler[SelfT any] interface {
-	// NewFromSSZ creates a new object from SSZ format with the given version.
-	NewFromSSZ(bz []byte, version common.Version) (SelfT, error)
-}
```
abi87, line 46: once New is moved out of the interface, `sszVersionedUnmarshaler` becomes the same as `SSZUnmarshaller` to merged the interfaces

storage/encoding/ssz.go
```diff
@@ -25,14 +25,10 @@ import (
 	"github.com/berachain/beacon-kit/primitives/constraints"
 	"github.com/berachain/beacon-kit/primitives/decoder"
 	"github.com/davecgh/go-spew/spew"
-	"github.com/karalabe/ssz"
 )
 
 // SSZValueCodec provides methods to encode and decode SSZ values.
-type SSZValueCodec[T interface {
-	ssz.StaticObject
```
abi87, line 33: not strictly necessary but still simplfies a bit.
There is no reason why we could not encode `ssz.DynamicObject`s, so just kept `constraints.SSZMarshallable` that already contains `ssz.Object`

consensus-types/types/payload.go
```diff
@@ -88,16 +88,17 @@ type ExecutionPayload struct {
...
+		Versionable:   NewVersionable(forkVersion),
+		BaseFeePerGas: &math.U256{},
+	}
+
+	// For any fork version after Bellatrix (Capella onwards), non-nil withdrawals are required.
+	if version.IsAfter(forkVersion, version.Bellatrix()) {
+		ep.Withdrawals = make([]*engineprimitives.Withdrawal, 0)
 	}
+	return ep
 }
```
abi87, line 102: This is crude fix I made. It works but I didn't fully get the SSZ related types.
I believe this fits better in the EnsureSyntaxFromSSZ

consensus/cometbft/service/encoding/encoding.go
```diff
@@ -81,7 +83,14 @@ func UnmarshalBeaconBlockFromABCIRequest(
...
+	block, err := ctypes.NewEmptySignedBeaconBlockWithVersion(forkVersion)
+	if err != nil {
+		return nil, fmt.Errorf("attempt at building block with wrong version %s: %w", forkVersion, err)
+	}
+	if err = decoder.SSZUnmarshal(blkBz, block); err != nil {
+		return nil, err
+	}
+	return block, nil
```
calbera, line 93: just noting, I would avoid making functional/behavior changes in this PR. i.e. there is no non-nil verification of withdrawals as there was before

consensus-types/types/block_test.go
```diff
@@ -136,8 +137,8 @@ func TestBeaconBlock_MarshalUnmarshalSSZ(t *testing.T) {
 		require.NoError(t, err)
 		require.NotNil(t, sszBlock)
 
-		var unmarshalledBlock *types.BeaconBlock
-		unmarshalledBlock, err = unmarshalledBlock.NewFromSSZ(sszBlock, v)
+		unmarshalledBlock := types.NewEmptyBeaconBlockWithVersion(v)
```
rezbera, line 140: Noting that we're effectively reintroducing the `Empty()` with this which requires unmarshallers to always remember to call `NewXYZ..` before unmarshalling.

I'm okay with this but it reverts one of the goals of the original change

consensus-types/types/payload_header.go
```diff
@@ -82,6 +82,13 @@ type ExecutionPayloadHeader struct {
...
 
+func NewEmptyExecutionPayloadHeaderWithVersion(version common.Version) *ExecutionPayloadHeader {
+	return &ExecutionPayloadHeader{
+		Versionable:   NewVersionable(version),
+		BaseFeePerGas: &math.U256{},
+	}
+}
+
 // empty returns an empty ExecutionPayloadHeader.
 func (*ExecutionPayloadHeader) empty(version common.Version) *ExecutionPayloadHeader {
```
rezbera, line 93: This could be removed

consensus-types/types/block_test.go
abi87: Yes, effectively re-introducing `Empty()`!
Happy to revert and move to `pre-unmarshalling` if there is agreement over it

rezbera: Let's keep as is for now. We cal always revisit it.

consensus-types/types/payload_header.go
abi87: Needed to fix NewFromJson but done

consensus/cometbft/service/encoding/encoding.go
abi87: Indeed this changed a bit and I should have explained this better. I think the behaviour is effectively the same.
`Withdrawals` are instantiated in `NewEmptyExecutionPayloadWithVersion` (only if the fork is high enough) and the check is done in `ValidateAfterDecodingSSZ` which is called "recursively" from signed beacon block down

rezbera (approved):

PR #2617, opened 2025-03-20
chore(ssz types):  Move UnusedType to Common and Define Interfaces Locally

rezbera:
This change prevents import cycles, which can occur quite easily


shotes (approved):
LGTM

abi87 (approved):
buona

PR #2613, opened 2025-03-19
feat(pectra): Pectra Block Handling (3/n)

rezbera:

PR #2618, opened 2025-03-20
feat(pectra): Pectra Consensus Types (1/n)

rezbera:
Replaces https://github.com/berachain/beacon-kit/pull/2599

The PR introduces the core types necessary for Pectra. Most of the changes are hand picked from https://github.com/berachain/beacon-kit/pull/2613 which is a pr that can run prague EVM to avoid larger PRs.

### PrepareProposal
An ExecutionRequest is first picked up in `GetPayload`. It is received as an array of bytes that is _not_ a perfect SSZ encoding, but rather it is custom with a special prefix byte that identifies the type of the request. This is specified in the EIP https://eips.ethereum.org/EIPS/eip-7685. In the PR, we refer to this as `EncodedExecutionRequest`

Once it is in the Consensus Layer, we 'Decode' This special message using `DecodeExecutionRequests` into an `ExecutionRequests`. Once converted to this form we can take advantage of normal SSZ and also attach it to the `BeaconBlockBody`. 

### ProcessProposal
The BeaconBlock is full Serializable, and the `ExecutionPayload` would have been appropriately set in `PrepareProposal` if the proposer is non-byzantine. 

In ProcessProposal, we can simply `decoder.SSZUnmarshal` the body and do not need to call `DecodeExecutionRequests`. However, in order to validate the ExecutionRequests in `NewPayload`, we need to convert the `ExecutionRequests` back into the format the Execution Client knows about, as defined in EIP-7685. We do this using `GetExecutionRequestsList`.


testing/simulated/pectra_test.go
```diff
@@ -160,12 +160,12 @@ func (s *PectraSuite) TestFullLifecycle_WithRequests_IsSuccessful() {
 	// https://github.com/ethereum/go-ethereum/blob/39638c81c56db2b2dfe6f51999ffd3029ee212cb/core/blockchain_test.go#L4131-L4130
 	withdrawalTx := types.MustSignNewTx(senderKey, signer, &types.DynamicFeeTx{
 		ChainID:   elChainID,
-		Nonce:     1,
+		Nonce:     0,
```
rezbera, line 163: there was a bug where this tx wasn't being included because the nonce was `1`

consensus-types/types/deposit.go
```diff
@@ -104,8 +104,6 @@ func (d *Deposit) MarshalSSZ() ([]byte, error) {
 	return buf, ssz.EncodeToBytes(buf, d)
 }
 
-func (*Deposit) ValidateAfterDecodingSSZ() error { return nil }
```
calbera, line 107: why removed?

consensus-types/types/deposit_request.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+const MaxDepositRequestsPerPayload = 8192
+
+// DepositRequest is introduced in EIP6110 which is currently not processed.
+type DepositRequest = Deposit
+
+func (d *DepositRequest) ValidateAfterDecodingSSZ() error {
+	return nil
+}
```
calbera, line 37: think this should stay on the deposit type so that the `Deposit` type is backwards compatible with the decoder function used throughout

consensus-types/types/deposit.go
rezbera: Was overriden by DepositRequest. Fixed

rezbera: But also implies this wasn't being used

consensus-types/types/deposit_request.go
rezbera: Related to https://github.com/berachain/beacon-kit/pull/2618#discussion_r2008696393

go.mod
```diff
@@ -40,7 +40,7 @@ require (
 	github.com/phuslu/log v1.0.113
 	github.com/pkg/errors v0.9.1
 	github.com/prysmaticlabs/gohashtree v0.0.4-beta.0.20240624100937-73632381301b
-	github.com/prysmaticlabs/prysm/v5 v5.2.0
+	github.com/prysmaticlabs/prysm/v5 v5.3.0
```
rezbera, line 43: updated version as Prysm 5.2.0 has incorrect ssz marshalling for electra (they set consolidation requests max-size to `1`, but fixed in 5.3.0 to `2`.

consensus-types/types/body.go
```diff
@@ -102,6 +109,11 @@ type BeaconBlockBody struct {
 func (b *BeaconBlockBody) SizeSSZ(siz *ssz.Sizer, fixed bool) uint32 {
 	syncSize := b.syncAggregate.SizeSSZ(siz)
 	var size = 96 + 72 + 32 + 4 + 4 + 4 + 4 + 4 + syncSize + 4 + 4 + 4
+	if version.EqualsOrIsAfter(b.GetForkVersion(), version.Electra()) {
```
abi87, line 112: nit: consider adding the `includeExecRequest = version.EqualsOrIsAfter(b.GetForkVersion(), version.Electra())` boolean and reusing it

consensus-types/types/consolidation_request.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// MarshalSSZ marshals the ConsolidationRequests object to SSZ format by encoding each consolidation request individually.
+func (cr *ConsolidationRequests) MarshalSSZ() ([]byte, error) {
+	return eip7685.MarshalItems[*ConsolidationRequest](*cr)
+}
+
+// DecodeConsolidationRequests decodes SSZ data by decoding each request individually.
+func DecodeConsolidationRequests(data []byte) (*ConsolidationRequests, error) {
+	maxSize := maxConsolidationRequestsPerPayload * sszConsolidationRequestSize
```
abi87, line 81: not sure I follow: we are not requiring that this type is unused right? I understand it comes from the EL, so we may have to take it as it comes BUT we won't do any processing with it. Is this correct?

I am mainly trying to figure it out the difference among this type of handling and the UnsedType introduced by Shota

consensus-types/types/deposit_request.go
```diff
@@ -0,0 +1,63 @@
...
+package types
+
+import (
+	"fmt"
+
+	"github.com/berachain/beacon-kit/primitives/eip7685"
+	"github.com/karalabe/ssz"
+)
+
+const MaxDepositRequestsPerPayload = 8192
```
abi87, line 30: Isn't this a very high number of requests?
Where does it comes from? Should this be different in BeaconKit?

primitives/eip7685/ssz.go
```diff
@@ -0,0 +1,67 @@
...
+
+package eip7685
+
+import (
+	"fmt"
+
+	"github.com/karalabe/ssz"
+)
+
+// MarshalItems marshals a slice of items that satisfy SSZMarshaler.
```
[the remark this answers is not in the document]
abi87: Can you elaborate on why ssz is not good for this?
Would fastssz work instead? (karalabe ssz is not really maintained and i am just trying to understand whether ditching it in favour of fast ssz only would only spare us from implementing this code)

primitives/version/comparable.go
```diff
@@ -63,6 +63,11 @@ func IsBefore(a, b common.Version) bool {
 	return false
 }
 
+// IsBeforeOrEquals returns true if a is before or at the same version as b.
+func IsBeforeOrEquals(a, b common.Version) bool {
+	return !IsAfter(a, b)
```
abi87, line 68: nit: consider implementing this more directly as `IsBefore(a, b)  || Equals(a, b)`

testing/simulated/transformers.go
```diff
@@ -96,7 +96,7 @@ func transformExecutableDataToExecutionPayload(
 	data *gethprimitives.ExecutableData,
 ) (*ctypes.ExecutionPayload, error) {
 	// Check that the fork version is supported (pre-Deneb1).
-	if !version.IsBefore(forkVersion, version.Deneb1()) {
+	if version.EqualsOrIsAfter(forkVersion, version.Deneb1()) {
```
abi87, line 99: nit: comment is off. IIUC here we are changing the versions range we're testing (from pre Deneb1 to post or equal postDeneb1)

consensus-types/types/deposit_request.go
```diff
@@ -0,0 +1,63 @@
...
+// DecodeDepositRequests decodes SSZ data by decoding each request individually.
+func DecodeDepositRequests(data []byte) (*DepositRequests, error) {
+	maxSize := MaxDepositRequestsPerPayload * DepositSize
+	if len(data) > maxSize {
+		return nil, fmt.Errorf(
+			"invalid deposit requests SSZ size, requests should not be more than the max per "+
+				"payload, got %d max %d", len(data), maxSize,
+		)
+	}
+	depositSize := int(ssz.Size(&Deposit{}))
```
abi87, line 52: nit: could/should this be made a constant?

consensus-types/types/execution_requests.go
```diff
@@ -0,0 +1,196 @@
...
+
+	"github.com/berachain/beacon-kit/errors"
+	"github.com/berachain/beacon-kit/primitives/bytes"
+	"github.com/berachain/beacon-kit/primitives/common"
+	"github.com/karalabe/ssz"
+)
+
+const (
+	// ExecutionAddress = 20, PubKey = 48, Pubkey = 48
+	sszDynamicObjectOffset = 4
```
abi87, line 34: IIRC this is a universal constant for SSZ encoding.
No need to change anything in this PR, but we may want to cleanup and reuse this liberally eventually.

consensus-types/types/execution_requests.go
```diff
@@ -0,0 +1,196 @@
...
+	// Iterate over each encoded request group.
+	for _, encoded := range encodedRequests {
+		if len(encoded) < 1 {
+			return nil, errors.New("invalid execution request, length less than 1")
+		}
+		// The first byte indicates the request type.
+		reqType := encoded[0]
+
+		// Enforce that request types are in strictly increasing order.
+		if prevType != nil && *prevType >= reqType {
```
abi87, line 114: q: probably being paranoid here, but is there a way `prevType` may be not nil for reqTypes that are not the first one?
cc @shotes, just trying to understand if we are under-validating and if we should request more explicitly that prevTypes should be nil iff we are examining the first encoded request

consensus-types/types/execution_requests.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+		prevType = &reqType
+
+		// The remaining bytes are the SSZ serialization for this group.
+		data := encoded[1:]
+
+		// Switch based on the request type.
+		switch reqType {
+		case depositRequestType()[0]:
+			req, err := DecodeDepositRequests(data)
```
abi87, line 125: in this switch we are implicitly assuming that if `err == nil` then `req != nil`. It should be fine but I wonder if we could return the actual type rather than a pointer to it?
It would avoid the deference (and the types themselves are slices so kinda pointers already)

consensus-types/types/signed_beacon_block_test.go
```diff
@@ -185,9 +185,7 @@ func TestSignedBeaconBlock_EmptySerialization(t *testing.T) {
 	t.Parallel()
 	runForAllSupportedVersions(t, func(t *testing.T, fv common.Version) {
 		orig := &types.SignedBeaconBlock{
-			BeaconBlock: &types.BeaconBlock{
-				Versionable: types.NewVersionable(fv),
-			},
+			BeaconBlock: types.NewEmptyBeaconBlockWithVersion(fv),
```
abi87, line 188: <3

abi87 (approved):
This looks good to me, mostly questions and nits. I took the liberty to merge `main` where I had previously merged https://github.com/berachain/beacon-kit/pull/2624

primitives/eip7685/ssz.go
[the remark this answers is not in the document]
rezbera: When we use the libraries, it panics. @calbera spent some more time investigating but we spent 2 days trying to figure it out and eventually stopped to avoid wasting more time.

consensus-types/types/consolidation_request.go
rezbera: Correct. We will get it from the execution client and we will ignore processing on it in the consensus client. 
We cannot enforce that it is empty as that is controlled by EL and would need us to fork EL.

consensus-types/types/deposit_request.go
rezbera: It comes from the specification https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#execution

In the EIP https://eips.ethereum.org/EIPS/eip-6110 - it is not considered a DoS vector but we can re-evaluate if its problematic after Withdrawals is done and we decide if we want to change deposits

consensus-types/types/body.go
rezbera: 543d52dda2b6f22dd2880b17c391b74f23f6b163

primitives/version/comparable.go
rezbera: 27e47103755e790462a057d41223beb9da803f7a

consensus-types/types/deposit_request.go
rezbera: true, we can reuse the constant given we are already using it. 

57891ae976ffeee40ae2761b3567c2b696c5385b

rezbera: 57891ae976ffeee40ae2761b3567c2b696c5385b

consensus-types/types/execution_requests.go
rezbera: I don't think that's possible - i mostly inspired by prysm impl https://github.com/prysmaticlabs/prysm/blob/e0e735470809df29c5404f64102ffbae5a574e0a/proto/engine/v1/electra.go#L34-L33

testing/simulated/transformers.go
abi87: I missed the negation. All good

primitives/eip7685/ssz.go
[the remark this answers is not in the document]
calbera: I don't imagine fastssz solves this. as shown in the link, prysm uses these helpers for electra types (execution requests which are dynamic lists) even with fastssz lib

testing/simulated/transformers.go
calbera: nit but this would read easier if said
```suggestion
	if version.IsAfter(forkVersion, version.Deneb()) {
```

calbera (approved):

consensus-types/types/execution_requests.go
shotes: > is there a way `prevType` may be not nil for reqTypes that are not the first one?

Agreed that I don't think this is possible. reqType can't be nil, so prevType by extension must not be nil after the first one, since it is copied from reqType.

consensus-types/types/execution_requests.go
```diff
@@ -0,0 +1,196 @@
...
+			return nil, errors.New("requests should be in sorted order and unique")
+		}
+		prevType = &reqType
+
+		// The remaining bytes are the SSZ serialization for this group.
+		data := encoded[1:]
+
+		// Switch based on the request type.
+		switch reqType {
+		case depositRequestType()[0]:
```
shotes, line 124: Can we guarantee that this `depositRequestType` has at least one element?

abi87: These har hard-coded types, not elements of the requests?

consensus-types/types/execution_requests.go
```diff
@@ -0,0 +1,196 @@
...
+		default:
+			return nil, fmt.Errorf("unsupported request type %d", reqType)
+		}
+	}
+
+	return &result, nil
+}
+
+func depositRequestType() []byte {
+	return []byte{0x00}
```
shotes, line 151: nit: can these be changed to:
```go
func depositRequestType() byte {
	return byte(0x00)
}
```
or maybe changed to constants? In both of the places each of these functions are used, we only use the value of the first element and it is immediately unpackaged from the byte slice.

rezbera: Good cal

0b0bcc6560bd8c88dfb9af12f34f1ae99d1e55b6

shotes (approved):

PR #2620, opened 2025-03-21
chore(configs): Warn on configs

abi87:
Following our extensive test campaign, I would add some warning related to the configurations that are most responsible for issues.


fridrik01 (approved):
lgtm

PR #2622, opened 2025-03-21
remove ethereumjs

camembera:
@chuck-bear can confirm this won't matter for anything in infra


chuck-bear:
@camembera I confirm that this won't matter for anything in infra

calbera:
outdated and already handled

PR #2624, opened 2025-03-21
chore(ssz): Make UnusedType use the ssz codec

shotes:
Previously we had hard coded the encoding/decoding for UnusedType, making it so that we had to treat it as a special case. This PR changes the implementation to use the SSZ Codec for encoding/decoding and adds tests to verify that the implementation produces the same results.

Note by @abi87: once `UnusedType` is not special cased anymore, we can also marginally simplify the package structure since we do not have any import cycle anymore.


primitives/common/unused_type.go
```diff
@@ -53,21 +51,14 @@ func (ut *UnusedType) DefineSSZ(c *ssz.Codec) {
 
 // MarshalSSZ marshals the UnusedType object to SSZ format.
 func (ut *UnusedType) MarshalSSZ() ([]byte, error) {
-	return []byte{uint8(*ut)}, nil
+	buf := make([]byte, ssz.Size(ut))
+	return buf, ssz.EncodeToBytes(buf, ut)
```
calbera, line 55: just curious, why wasn't this done before when first implementing UnusedType?

primitives/decoder/decoder.go
```diff
@@ -33,17 +32,11 @@ import (
...
-		// Note: ValidateAfterDecodingSSZ may change v even if it returns error
-		// (depending on the specific implementations)
-		return v.ValidateAfterDecodingSSZ()
+	if err := ssz.DecodeFromBytes(buf, v); err != nil {
+		return fmt.Errorf("failed decoding %T: %w", v, err)
 	}
+
+	// Note: ValidateAfterDecodingSSZ may change v even if it returns error
+	// (depending on the specific implementations)
+	return v.ValidateAfterDecodingSSZ()
```
calbera, line 41: much cleaner now nice!

primitives/common/unused_type_test.go
```diff
@@ -0,0 +1,90 @@
...
+			got := new(common.UnusedType)
+			if err := ssz.DecodeFromBytes(tt.buf, got); err != nil {
+				if tt.wantErr {
+					return
+				}
+				t.Errorf("DecodeFromBytes() error = %v", err)
+			}
+			want := new(common.UnusedType)
+			*want = common.UnusedType(tt.buf[0])
+			if !reflect.DeepEqual(got, want) {
```
calbera, line 57: super nit: this is my personal preference, but up to you
```suggestion
			want := common.UnusedType(tt.buf[0])
			if !reflect.DeepEqual(got, &want) {
```

calbera (approved):
lgtm

primitives/common/unused_type.go
shotes: I think the answer to that question is partially that I thought it would be simpler and partially just because I neglected it. Thought it would be simpler to shortcut the ssz lib since it was a 1 to 1 encoding.

But turns out it's just all around better to be consistent with the implementation.

primitives/common/unused_type_test.go
shotes: applied suggestion!

primitives/common/unused_type.go
```diff
@@ -53,21 +51,14 @@ func (ut *UnusedType) DefineSSZ(c *ssz.Codec) {
...
-		return fmt.Errorf("expected 1 byte got %d", len(buf))
-	}
-	//#nosec:G701 // UnusedType is uint8 and byte is uint8.
-	*v = UnusedType(buf[0])
-	return nil
+func (ut *UnusedType) ValidateAfterDecodingSSZ() error {
+	return ut.EnforceUnused()
 }
 
-func (*UnusedType) ValidateAfterDecodingSSZ() error { return nil }
```
abi87, line 69: this was probably a bug and I think calling `ut.EnforceUnused()` in `ValidateAfterDecodingSSZ` is the correct thing to do (although the bug would have only showed up if we marshalled UnusedTypes by themselves, not as a part of an object

PR #2626, opened 2025-03-24
chore(licenses): fix licences part 2

abi87:
Follow up from https://github.com/berachain/beacon-kit/pull/2619
Now every go file has BUSL licence, as expected

### DIFF
--- a/consensus-types/types/attestation_data.go
+++ b/consensus-types/types/attestation_data.go
@@ -1,27 +1,22 @@
 // SPDX-License-Identifier: BUSL-1.1
 //
-// Copyright (c) 2025 Berachain Foundation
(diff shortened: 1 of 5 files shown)


consensus-types/types/attestation_data.go
```diff
@@ -1,27 +1,22 @@
 // SPDX-License-Identifier: BUSL-1.1
 //
-// Copyright (c) 2025 Berachain Foundation
+// Copyright (C) 2025, Berachain Foundation. All rights reserved.
```
abi87, line 3: https://github.com/berachain/beacon-kit/pull/2619 fixed the heading but not the rest of the licence

PR #2630, opened 2025-03-25
feat(fork): timestamp based forking (1/n)

shotes:
This PR introduces timestamp based forking, transitioning from epoch based forking. The goal of this PR is to only modify the way in which fork versions are set. It attempts to not touch the way fork version are used or accessed.

- [x] Change `ActiveForkVersionForSlot` and `ActiveForkVersionForEpoch` to `ActiveForkVersionForTimestamp`
- [x] Change `<forkVersion>Epoch` chainspec values to `<forkVersion>Time`
- [x] ALWAYS use timestamp from `BeaconBlockBody.ExecutionPayload.Timestamp` for getting `ActiveForkVersionForTimestamp`
  - [x] Ensure forkVersion for optimistic build retrieval corresponds with the forkVersion during optimistic build request.
  - [x] Special case for forkVersion while unmarshaling `BeaconBlock` in request. Verify forkVersions afterwards
- [x] Fix timestamps in simulated tests
- [x] Make e2e tests work with static fork versions
- [ ] Make fork times happen during middle of e2e test
  - Edit: e2e tests currently pass because the fork is set to have already happened before the test starts. Currently, e2e tests do not happen _across_ fork boundaries at all because we cannot set fork times dynamically in the chain spec just yet. This will be addressed in a separate PR.
- [x] Set Deneb1 fork in chainspecs
  - [x] for Mainnet
  - [x] for Bepolia (testnet)


beacon/validator/block_builder.go
```diff
@@ -73,29 +70,42 @@ func (s *Service) BuildBlockAndSidecars(
...
+	// Get the payload for the block.
+	envelope, err := s.retrieveExecutionPayload(ctx, st, parentBlockRoot, slotData)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed retrieving execution payload: %w", err)
+	}
+
+	// We introduce hard forks with the expectation that the first block proposed after the
+	// hard fork timestamp is when new rules apply. When building blocks, we set the timestamp
+	// now so we can build a block for the correct fork.
+	timestamp := envelope.GetExecutionPayload().GetTimestamp()
```
shotes, line 90: Had to re-order the computations in this function. We need to be sure that the timestamp we are feeding to the other builder functions is the exact same timestamp that is in the execution payload. Because of potential different times from the optimistic builder, we must grab the time from the payload we've grabbed from the builder.

state-transition/core/validation_deposits_test.go
```diff
@@ -94,7 +93,7 @@ func TestInvalidDeposits(t *testing.T) {
 		types.NewEth1Data(depRoot),
 		10,
 		[]*types.Deposit{invalidDeposit},
-		st.EVMInflationWithdrawal(constants.GenesisSlot+1),
+		st.EVMInflationWithdrawal(10),
```
shotes, line 96: These values of `timestamp = 10` in these tests are based on the hardcoded timestamp set to 10 a few lines above.

beacon/blockchain/execution_engine.go
```diff
@@ -47,15 +45,14 @@ func (s *Service) sendPostBlockFCU(
...
-	beaconBlk := blk.GetBeaconBlock()
 	// TODO: Switch to New().
 	req := ctypes.BuildForkchoiceUpdateRequestNoAttrs(
 		&engineprimitives.ForkchoiceStateV1{
 			HeadBlockHash:      lph.GetBlockHash(),
 			SafeBlockHash:      lph.GetParentHash(),
 			FinalizedBlockHash: lph.GetParentHash(),
 		},
-		s.chainSpec.ActiveForkVersionForSlot(beaconBlk.GetSlot()),
+		lph.GetForkVersion(),
```
shotes, line 55: don't need to recalculate forkVersion here. It is in the latestExecutionPayloadHeader.

beacon/blockchain/process_proposal.go
```diff
@@ -141,6 +153,7 @@ func (s *Service) ProcessProposal(
 	}
 
 	// Process the block
+	// NOTE: the consensusBlk.GetConsensusTime() must NOT be used to determine the fork version.
```
shotes, line 156: It's actually ok to use this timestamp to determine the fork version, since we have the check above that ensures that the fork versions are the same. Leaving the comment here though to indicate that the consensus time and block time are different, and the consensus time should only ever be used to validate the block time.

beacon/blockchain/interfaces.go
```diff
@@ -67,11 +67,11 @@ type LocalBuilder interface {
 		ctx context.Context,
 		st *statedb.StateDB,
 		slot math.Slot,
-		timestamp uint64,
+		timestamp math.U64,
 		parentBlockRoot common.Root,
 		headEth1BlockHash common.ExecutionHash,
 		finalEth1BlockHash common.ExecutionHash,
-	) (*engineprimitives.PayloadID, error)
+	) (*engineprimitives.PayloadID, common.Version, error)
```
rezbera, line 74: Does this need to return version now? Seems like it might not be necessary

beacon/blockchain/process_proposal.go
```diff
@@ -90,6 +93,15 @@ func (s *Service) ProcessProposal(
...
 
 	blk := signedBlk.GetBeaconBlock()
+
+	// The timestamp from the ABCI request can be slightly different from the timestamp
+	// put in the BeaconBlock's ExecutionPayload. We need to ensure that the fork version
+	// for these timestamps is the same, since the ABCI request timestamp must be used to
+	// determine the unmarshaling of the BeaconBlock. This may result in a failed proposal
+	// or two at the start of the fork.
+	if !version.Equals(s.chainSpec.ActiveForkVersionForTimestamp(blk.GetTimestamp()), forkVersion) {
+		return ErrVersionMismatch
```
rezbera, line 103: Can we add more info to the error, i.e. which fork versions

beacon/blockchain/process_proposal.go
```diff
@@ -90,6 +93,15 @@ func (s *Service) ProcessProposal(
 	}
 
 	blk := signedBlk.GetBeaconBlock()
+
+	// The timestamp from the ABCI request can be slightly different from the timestamp
+	// put in the BeaconBlock's ExecutionPayload. We need to ensure that the fork version
+	// for these timestamps is the same, since the ABCI request timestamp must be used to
+	// determine the unmarshaling of the BeaconBlock. This may result in a failed proposal
+	// or two at the start of the fork.
+	if !version.Equals(s.chainSpec.ActiveForkVersionForTimestamp(blk.GetTimestamp()), forkVersion) {
```
rezbera, line 102: nice

beacon/blockchain/interfaces.go
shotes: It is used during the `getPayload` call. This forkVersion must be the same forkVersion as the original payload request. Since the forkVersion may be different if we have a cached payloadID from an optimistic build and it crosses fork boundaries, we return the associated forkVersion

beacon/validator/block_builder.go
```diff
@@ -73,29 +70,42 @@ func (s *Service) BuildBlockAndSidecars(
...
+
+	// Build forkdata used for the signing root of the reveal and the sidecars
+	forkData, err := s.buildForkData(st, timestamp)
 	if err != nil {
 		return nil, nil, err
 	}
 
 	// Create a new empty block from the current state.
-	blk, err := s.getEmptyBeaconBlockForSlot(st, blkSlot)
+	blk, err := s.getEmptyBeaconBlockForSlot(st, blkSlot, timestamp)
```
rezbera, line 99: can we just pass in the forkVersion from `forkData.CurrentVersion` instead of timestamp?

beacon/validator/block_builder.go
```diff
@@ -108,7 +118,7 @@ func (s *Service) BuildBlockAndSidecars(
 	if err = s.computeAndSetStateRoot(
 		ctx,
 		slotData.GetProposerAddress(),
-		slotData.GetConsensusTime(),
+		timestamp,
```
rezbera, line 121: `computeAndSetStateRoot` 3rd arg is named `consensusTime`

rezbera: It doesn't necessarily get used in prepare proposal since we don't verify payload but still misleading

beacon/validator/block_builder.go
```diff
@@ -148,7 +158,7 @@ func (s *Service) BuildBlockAndSidecars(
 
 // getEmptyBeaconBlockForSlot creates a new empty block.
 func (s *Service) getEmptyBeaconBlockForSlot(
-	st *statedb.StateDB, requestedSlot math.Slot,
+	st *statedb.StateDB, requestedSlot math.Slot, timestamp math.U64,
```
rezbera, line 161: could we also pass in `parentBlockRoot` to skip recalcing the `parentBlockRoot`

config/spec/defaults.go
```diff
@@ -57,8 +57,8 @@ const (
 	defaultTargetSecondsPerEth1Block = 2 // Berachain specific.
 
 	// Fork-related values.
-	defaultDeneb1ForkEpoch  = 9999999999999998 // Set as a future epoch as not yet determined.
-	defaultElectraForkEpoch = 9999999999999999 // Set as a future epoch as not yet determined.
+	defaultDeneb1ForkTime  = 9999999999999998 // Set as a future epoch as not yet determined.
```
rezbera, line 60: comments outdated

config/spec/devnet.go
```diff
@@ -58,8 +67,11 @@ func DevnetChainSpecData() *chain.SpecData {
 	specData := MainnetChainSpecData()
 	specData.DepositEth1ChainID = DevnetEth1ChainID
 
-	// Deneb1 fork takes place at epoch 1.
-	specData.Deneb1ForkEpoch = devnetDeneb1ForkEpoch
+	// Deneb1 fork takes place at timestamp estimated to be around epoch 1.
+	specData.Deneb1ForkTime = devnetDeneb1ForkTime
+
+	// Electra fork takes place at timestamp estimated to be around epoch 2.
```
rezbera, line 73: comment

testing/e2e/e2e_inflation_test.go
```diff
@@ -36,79 +38,73 @@ func (s *BeaconKitE2ESuite) TestEVMInflation() {
...
-		)
+			expectedBalance = new(big.Int).Mul(
+				new(big.Int).SetUint64(inflationPerBlock*params.GWei),
+				big.NewInt(blkNum-forkSlot+1),
+			)
+		} else {
+			expectedBalance = new(big.Int).Mul(
+				new(big.Int).SetUint64(inflationPerBlock*params.GWei),
+				big.NewInt(blkNum),
+			)
```
rezbera, line 95: Can we enforce this case is hit in the test incase we change devnet spec and forget to update this

rezbera (commented):
Did a first pass, will recheck after addressed

config/spec/mainnet.go
```diff
@@ -73,8 +73,10 @@ const (
 	// default deposit contract address.
 	mainnetDepositContractAddress = defaultDepositContractAddress
 
-	// mainnetDeneb1ForkEpoch is the epoch at which the Deneb1 fork occurs.
-	mainnetDeneb1ForkEpoch = 2855
+	// mainnetDeneb1ForkTime is the timestamp at which the Deneb1 fork occurs.
+	// This is calculated based on the timestamp of the 2855th mainnet epoch, block 548160, which
+	// was used to initiate the fork when beacon-kit forked by epoch instead of by timestamp.
```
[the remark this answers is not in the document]
shotes: absolutely

testing/e2e/e2e_inflation_test.go
shotes: This case is hit on block 0 with timestamp 0. Then block 1 jumps to comet consensus time which is `time.Now()`.

config/spec/devnet.go
shotes: https://github.com/berachain/beacon-kit/pull/2630/commits/88be46d53f18d7d650f61136896c11816be50056

config/spec/defaults.go
shotes: https://github.com/berachain/beacon-kit/pull/2630/commits/88be46d53f18d7d650f61136896c11816be50056

beacon/validator/block_builder.go
shotes: https://github.com/berachain/beacon-kit/pull/2630/commits/88be46d53f18d7d650f61136896c11816be50056

beacon/validator/block_builder.go
shotes: https://github.com/berachain/beacon-kit/pull/2630/commits/88be46d53f18d7d650f61136896c11816be50056

beacon/validator/block_builder.go
shotes: https://github.com/berachain/beacon-kit/pull/2630/commits/88be46d53f18d7d650f61136896c11816be50056

beacon/blockchain/process_proposal.go
shotes: https://github.com/berachain/beacon-kit/pull/2630/commits/88be46d53f18d7d650f61136896c11816be50056

config/spec/defaults.go
```diff
@@ -57,8 +57,8 @@ const (
 	defaultTargetSecondsPerEth1Block = 2 // Berachain specific.
 
 	// Fork-related values.
-	defaultDeneb1ForkEpoch  = 9999999999999998 // Set as a future epoch as not yet determined.
```
shotes, line 60: Deleted comment. Not sure what to do with default value since it is unused. Just left it there like the other unused default values in the file.

beacon/blockchain/execution_engine.go
calbera: noting, this is safe to do because the LatestExecutionPayloadHeader has already been set for the processed block in the beacon state. so non-effectual change

chain/helpers.go
```diff
@@ -26,17 +26,12 @@ import (
...
-// ActiveForkVersionForSlot returns the active fork version for a given slot.
-func (s spec) ActiveForkVersionForSlot(slot math.Slot) common.Version {
-	return s.ActiveForkVersionForEpoch(s.SlotToEpoch(slot))
-}
-
-// ActiveForkVersionForEpoch returns the active fork version for a given epoch.
-func (s spec) ActiveForkVersionForEpoch(epoch math.Epoch) common.Version {
-	if epoch >= s.ElectraForkEpoch() {
+// ActiveForkVersionForTimestamp returns the active fork version for a given timestamp.
+func (s spec) ActiveForkVersionForTimestamp(timestamp math.U64) common.Version {
```
calbera, line 30: nit: can we change this to take in `timestamp uint64`? I see most places its called with a (u)int64 and even here its Unwrapped. so the U64 type feels unnecessarily introduced in this context.

calbera (commented):
utACK. some comments. will re-review after addressed and sync through mainnet deneb1 works.

beacon/blockchain/payload.go
```diff
@@ -74,7 +74,7 @@ func (s *Service) forceSyncUponProcess(
 			SafeBlockHash:      lph.GetParentHash(),
 			FinalizedBlockHash: lph.GetParentHash(),
 		},
-		s.chainSpec.ActiveForkVersionForSlot(slot),
+		s.chainSpec.ActiveForkVersionForTimestamp(lph.GetTimestamp()),
```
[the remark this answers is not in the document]
shotes: Agreed. I will remove the slot usage in this function entirely. This fork version for this request should be the fork version of the Head block, which is lph in this case. (also note that the fork version in the engine API call only determines which version of the API call to make, which has not changed in pectra - i.e. `eth_forkchoiceUpdatedV3` for Deneb and up.)

beacon/blockchain/process_proposal.go
```diff
@@ -90,6 +93,21 @@ func (s *Service) ProcessProposal(
...
+	// for these timestamps is the same, since the ABCI request timestamp must be used to
+	// determine the unmarshaling of the BeaconBlock. This may result in a failed proposal
+	// or two at the start of the fork.
+	blkVersion := s.chainSpec.ActiveForkVersionForTimestamp(blk.GetTimestamp())
+	if !version.Equals(blkVersion, forkVersion) {
+		err = fmt.Errorf("CometBFT version %v, BeaconBlock version %v: %w",
+			forkVersion, blkVersion,
+			ErrVersionMismatch,
+		)
+		s.logger.Warn(err.Error())
```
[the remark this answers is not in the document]
shotes: Good point. Did this to copy the implementation of the other errors. Will make this modification for all instances of the `s.logger.Warn(err.Error())` in `ProcessProposal`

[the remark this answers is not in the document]
calbera: cool with that

beacon/blockchain/process_proposal.go
```diff
@@ -149,6 +167,8 @@ func (s *Service) ProcessProposal(
 	err = s.VerifyIncomingBlock(
 		ctx,
 		consensusBlk.GetBeaconBlock(),
+		// The ConsensusTime() should only be used to validate the block timestamp. For all other
+		// timestamp needs, make sure to use the block timestamp.
```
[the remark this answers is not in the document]
shotes: It's also technically not correct, since we verify that the active fork version for both timestamps are equal to each other.

I'm adding a more descriptive comment that explains each timestamp and what they are used for earlier, where we do the check for the active fork version.

config/spec/devnet.go
```diff
@@ -38,8 +38,17 @@ const (
 	// of Gwei) that can be staked.
 	devnetMaxStakeAmount = 4000 * params.GWei
 
-	// devnetDeneb1ForkEpoch is the epoch at which the Deneb1 fork occurs.
-	devnetDeneb1ForkEpoch = 1
+	// devnetDeneb1ForkTime is the timestamp at which the Deneb1 fork occurs.
+	// This must be greater than 0.
```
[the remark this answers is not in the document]
shotes: It must be greater than 0 to support a specific e2e test. But I have just modified that e2e test so that it does not require the time to be > 0.

beacon/blockchain/payload.go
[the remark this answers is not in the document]
shotes: https://github.com/berachain/beacon-kit/pull/2630/commits/9f0ae7ed21040a1551754ad8204a0414e0493b13

chain/helpers.go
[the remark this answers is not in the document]
shotes: https://github.com/berachain/beacon-kit/pull/2630/commits/9f0ae7ed21040a1551754ad8204a0414e0493b13

beacon/blockchain/process_proposal.go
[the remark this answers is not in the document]
shotes: https://github.com/berachain/beacon-kit/pull/2630/commits/9f0ae7ed21040a1551754ad8204a0414e0493b13

beacon/blockchain/process_proposal.go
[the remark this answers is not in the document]
shotes: https://github.com/berachain/beacon-kit/pull/2630/commits/9f0ae7ed21040a1551754ad8204a0414e0493b13

config/spec/devnet.go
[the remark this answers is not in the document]
shotes: https://github.com/berachain/beacon-kit/pull/2630/commits/9f0ae7ed21040a1551754ad8204a0414e0493b13

config/spec/devnet.go
```diff
@@ -38,8 +38,17 @@ const (
...
+	// TODO(fork): Make devnet fork time take place during each devnet test.
+	// Currently the e2e tests start already on the Deneb1 fork, since time.Now()
+	// returns the current time, whereas the Deneb1ForkTime measures the seconds
+	// since the Unix Epoch.
+	devnetDeneb1ForkTime = 1
```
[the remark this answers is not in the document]
shotes: https://github.com/berachain/beacon-kit/pull/2630/commits/9f0ae7ed21040a1551754ad8204a0414e0493b13

beacon/blockchain/payload.go
```diff
@@ -64,7 +51,6 @@ func (s *Service) forceSyncUponProcess(
 		"head_eth1_hash", lph.GetBlockHash(),
 		"safe_eth1_hash", lph.GetParentHash(),
 		"finalized_eth1_hash", lph.GetParentHash(),
-		"for_slot", slot.Base10(),
```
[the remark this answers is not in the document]
shotes: https://github.com/berachain/beacon-kit/pull/2630/commits/94dfdf8ba5f63fac91d5b1e876fc71032b107d06

chain/helpers_test.go
```diff
@@ -94,35 +94,6 @@ func TestSlotToEpoch(t *testing.T) {
 	}
 }
 
-// TestActiveForkVersionForSlot tests the ActiveForkVersionForSlot method.
-func TestActiveForkVersionForSlot(t *testing.T) {
```
abi87, line 98: q: should this be replaced by a TestActiveForkVersionForTimestamp?

state-transition/core/validation_deposits_test.go
abi87: yeah, timestamp used to not be important really, but now it is. I will make a pass on a separate PR to fix it up

chain/helpers_test.go
shotes: The above function `TestActiveForkVersionForEpoch` was re-implemented to test the new timestamp functionality. I forgot to change the name of the test case though. Will push up change in the name of the test case to set it as `TestActiveForkVersionForTimestamp`

shotes: https://github.com/berachain/beacon-kit/pull/2630/commits/9d8e5d8d3d9a433b0e74d1f962ee2d2b1c62807e

beacon/blockchain/finalize_block.go
```diff
@@ -43,7 +43,9 @@ func (s *Service) FinalizeBlock(
 		req,
 		BeaconBlockTxIndex,
 		BlobSidecarsTxIndex,
-		s.chainSpec.ActiveForkVersionForSlot(math.Slot(req.Height))) // #nosec G115
+		// req.GetTime() should ALWAYS be during the same fork version as blk.GetTimestamp()
+		// due to checks during ProcessProposal.
```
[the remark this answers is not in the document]
shotes: https://github.com/berachain/beacon-kit/pull/2630/commits/d2d2c8f7cbfd8f3b5cafa26de94ca7926e63b828

rezbera:
Noting that the following tests have been completed as of yet:
1. Syncing Mainnet Node
2. Syncing Testnet Node
3. Executing the hard fork on Devnet environment under load.


The following tests are yet to be done and are prerequisite for this being merged:
4. The network hard fork is configured in genesis and executed successfully.
- a.  Hard fork with some missing validators (i.e. crashed or offline)
- b.  Hard fork with some validators misconfigured
- c. Hard fork with network delays (if possible) 
5. The network starts as a v.1.1.4 and then upgrades to 1.1.5.rc0 with a configured fork
- a. Under high load at the time of the fork (blob + tx)
- b.  Hard fork with some missing validators (i.e. crashed or offline)
- c. Hard fork with some validators misconfigured
- d. Hard fork with network delays (if possible)

beacon/blockchain/process_proposal.go
```diff
@@ -90,24 +93,41 @@ func (s *Service) ProcessProposal(
...
+	// This "consensus time" is used to constrain the timestamp set as the "block time" by the
+	// beacon-kit app, but they are not always equal in value. The "block time" is used by the
+	// beacon-kit consensus and execution layers to determine the active fork version.
+	//
+	// When unmarshaling the BeaconBlock, we do not yet have access to the "block time", so we
+	// must rely on the "consensus time" as our best estimation of the "block time" needed to
+	// determine the current fork version. Since the two timestamps could be different, we need to
+	// ensure that the fork version for these timestamps are the same. This may result in a failed
+	// proposal or two at the start of the fork.
+	blkVersion := s.chainSpec.ActiveForkVersionForTimestamp(blk.GetTimestamp())
```
abi87, line 110: general question: I wonder if instead of guessing, we should explicitly add a third transaction to the cometBFT block to communicate metadata like the fork version (for the time being) and possibly others in the next future.
It should simplify the flow I believe as:
- we would check the fork against the timestamps (which is a cheap check).
- we would correctly initialize the fork version in the blocks (which is never serialized afaik, nor inferred by its content)

abi87: My other question whether we could infer the version from the data serialized/not serialized somehow, which also would avoid the possible liveness issue.

abi87 (commented):
I am not opposed to this PR and I don't find major issues with its correctness.
The one point I am not sure about is the potential candidate issue around fork activation.
I would if we should explicitly introduced a third, metadata transaction in CometBFT block and store the fork there.
This would:
- allow initialising the blocks with the explicitly encoded version instead of guessing based on req.Height
- check the fork matched the expected one, once payload timestamp is serialized

I believe this structure solved the liveness issue as it makes impossible to mark valid blocks as invalid due to timing issues around req.Time

abi87 (approved):
Approving this with the agreement that we will come back and revisit the fork activation mechanism and evaluate different solutions, trying to remove the possibility of the liveness issue around fork activation

PR #2633, opened 2025-03-26
Fix/spelling and loglevels

camembera:
This PR tightens up the language and grammar in new warning messages

It also demotes a number of log messages from Info to Debug; I chose ones that don't mention a block#, only a block hash or other details.

There are still a good number of log messages sent at loglevel INFO for every block.   RFC on idea of demoting them all to Debug or promoting one of them to Warning so in the ordinary course of events, one line per block goes out, but importantly, if additional proposal rounds go around these should be warnings.

1. Timed out
2. Received proposal
3. Received complete proposal
4. processExecutionPayload
5. Finalizing commit of block
6. processExecutionPayload again
7. Finalized block
8. Committed state


camembera:
I don't understand why a bunch of other people's commits got roped into this

camembera:
to fill up my proposal, i'll add a change that demotes several other log entries to DEBUG except for 'committed'

PR #2634, opened 2025-03-27
[WI{chore(pectra): cleanup misc ssz and version stuff

calbera:


consensus-types/types/withdrawal_request.go
```diff
@@ -67,33 +75,61 @@ func (w *WithdrawalRequest) HashTreeRoot() common.Root {
...
+type WithdrawalRequests []*WithdrawalRequest
+
 // MarshalSSZ marshals the WithdrawalRequests object to SSZ format by encoding each deposit individually.
-func (wr *WithdrawalRequests) MarshalSSZ() ([]byte, error) {
-	return eip7685.MarshalItems[*WithdrawalRequest](*wr)
+func (wr WithdrawalRequests) MarshalSSZ() ([]byte, error) {
+	return sszutil.MarshalItemsEIP7685(wr)
+}
+
+func (wr WithdrawalRequests) DefineSSZ(codec *ssz.Codec) {
```
shotes, line 93: As far as I understand, `DefineSSZ` is only necessary to implement if we are going to use the standard ssz codec. But it seems that we are not using the standard ssz codec in favor of the special ssz methods `MarshalItemsEIP7685` and `UnmarshalItemsEIP7685`.

What is the use case for defining `DefineSSZ` here on these EIP7685 list types?

calbera: good catch, forgot to remove after testing stuff. https://github.com/berachain/beacon-kit/pull/2634/commits/f3940097044fb15c75ee7425b66785aff53f2f13

consensus-types/types/execution_requests_test.go
```diff
@@ -458,10 +449,10 @@ func TestDecodeExecutionRequests(t *testing.T) {
 			},
 		}
 		_, err = types.DecodeExecutionRequests(ebe.GetExecutionRequests())
-		require.ErrorContains(t, err, "invalid deposit requests SSZ size")
+		require.ErrorContains(t, err, "nvalid deposit requests SSZ size, got 0 expected at least 192")
```
rezbera, line 452: `nvalid`

da/store/store.go
```diff
@@ -76,8 +76,7 @@ func (s *Store) GetBlobSidecars(slot math.Slot) (types.BlobSidecars, error) {
 	sidecars := make(types.BlobSidecars, 0, len(sidecarBzs))
 	for _, sidecarBz := range sidecarBzs {
 		sidecar := new(types.BlobSidecar)
-		err = constraints.SSZUnmarshal(sidecarBz, sidecar)
-		if err != nil {
```
rezbera, line 80: surely there's a lint rule for this

execution/client/ethclient/engine.go
```diff
@@ -45,9 +45,18 @@ func (s *Client) NewPayload(
 	if version.IsBefore(forkVersion, version.Deneb()) {
 		return nil, ErrInvalidVersion
 	}
+
+	// Use V3 for Deneb versions (Deneb and Deneb1).
 	if version.Equals(forkVersion, version.Deneb()) || version.Equals(forkVersion, version.Deneb1()) {
-		return s.NewPayloadV3(ctx, req.GetExecutionPayload(), req.GetVersionedHashes(), req.GetParentBeaconBlockRoot())
+		return s.NewPayloadV3(
```
rezbera, line 51: also surely there's a lint rule

rezbera (approved):
1 nit, otherwise looks reasonable. I wonder what linters we could use to standardise some of the nit related changes

calbera:
there used to be a lll (line-length) linter at 80, which was very restrictive. We made it a lot larger a while back to allow for more read-ability, which now means more flexibility

da/store/store.go
calbera: shirley

consensus-types/types/execution_requests_test.go
calbera: fixed https://github.com/berachain/beacon-kit/pull/2634/commits/082455acd9e7fb1f4b1fbd0cc2fc0e6841979d07

PR #2635, opened 2025-03-27
fix(pectra): Fix Bug in ForceSyncUponFinalize

rezbera:
We were not correctly checking for Pectra here before.

Also refactors BuildNewPayloadRequestFromFork and removes the old builders which were only used in tests

Nilaway also does not support build tags, see https://github.com/uber-go/nilaway/issues/154. As such, we exclude it from checking test files. 
Since we added `test` build tag to the golangci checks, lots of linter checks were failing which are also addressed


beacon/blockchain/payload.go
```diff
@@ -94,11 +95,30 @@ func (s *Service) forceSyncUponFinalize(
 	// NewPayload call first to load payload into EL client.
 	executionPayload := beaconBlock.GetBody().GetExecutionPayload()
 	parentBeaconBlockRoot := beaconBlock.GetParentBlockRoot()
-	payloadReq := ctypes.BuildNewPayloadRequest(
-		executionPayload,
-		beaconBlock.GetBody().GetBlobKzgCommitments().ToVersionedHashes(),
-		&parentBeaconBlockRoot,
-	)
+	var payloadReq ctypes.NewPayloadRequest
```
shotes, line 98: Can we use `buildNewPayloadRequestFromFork` here instead of manually checking the fork?

abi87 (commented):
A couple of points:
- I was the one responsible for separating `HasValidVersionedAndBlockHashes` from the constructor. Given this type of bugs, I am not sure it was a good idea (we seem to be always call `HasValidVersionedAndBlockHashes` after we created the request, so maybe we should make the constructor always verify?
- I would drop/unexport `BuildNewPayloadRequest` which seems to be called in unit tests only now, and just have a single constructor.

rezbera:
> A couple of points:
> 
> * I was the one responsible for separating `HasValidVersionedAndBlockHashes` from the constructor. Given this type of bugs, I am not sure it was a good idea (we seem to be always call `HasValidVersionedAndBlockHashes` after we created the request, so maybe we should make the constructor always verify?
> * I would drop/unexport `BuildNewPayloadRequest` which seems to be called in unit tests only now, and just have a single constructor.

The original issue was not because `HasValidVersionedAndBlockHashes` was separated so I'm comfortable leaving it as is. I have refactored to remove `BuildNewPayloadRequest` which was the original problem

consensus-types/types/block_test.go
```diff
@@ -129,7 +52,7 @@ func TestBeaconBlockForDeneb(t *testing.T) {
 func TestBeaconBlock(t *testing.T) {
 	t.Parallel()
 	runForAllSupportedVersions(t, func(t *testing.T, v common.Version) {
-		block := generateValidBeaconBlock(t, v)
```
rezbera, line 132: We have 2 copies of the func so make a utility in `testing/utils`

testing/utils/generate.go
```diff
@@ -0,0 +1,110 @@
...
+// THIS LICENSE DOES NOT GRANT YOU ANY RIGHT IN ANY TRADEMARK OR LOGO OF
+// LICENSOR OR ITS AFFILIATES (PROVIDED THAT YOU MAY USE A TRADEMARK OR LOGO OF
+// LICENSOR AS EXPRESSLY REQUIRED BY THIS LICENSE).
+//
+// TO THE EXTENT PERMITTED BY APPLICABLE LAW, THE LICENSED WORK IS PROVIDED ON
+// AN “AS IS” BASIS. LICENSOR HEREBY DISCLAIMS ALL WARRANTIES AND CONDITIONS,
+// EXPRESS OR IMPLIED, INCLUDING (WITHOUT LIMITATION) WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, NON-INFRINGEMENT, AND
+// TITLE.
+//go:build test
```
calbera, line 20: nit can we add a line break after the LICENSE

node-api/backend/validator_test.go
```diff
@@ -35,7 +34,6 @@ import (
 	"github.com/berachain/beacon-kit/errors"
 	"github.com/berachain/beacon-kit/log/noop"
 	"github.com/berachain/beacon-kit/node-api/backend"
-	beacontypes "github.com/berachain/beacon-kit/node-api/handlers/beacon/types"
```
rezbera, line 38: same thing imported twice with different import alias

.golangci.yaml
```diff
@@ -11,6 +11,7 @@ run:
   # Timeout for analysis, e.g. 30s, 5m.
   # Default: 1m
   timeout: 3m
+  build-tags: ["test"]
```
rezbera, line 14: adding this causes a bunch of lint errors in files iwth this build tag. This PR addresses those as well

shotes:
utACK

consensus-types/types/payload_requests.go
```diff
@@ -55,34 +55,40 @@ type newPayloadRequest struct {
...
-		parentBeaconBlockRoot: parentBeaconBlockRoot,
-		executionRequests:     executionRequests,
+	if version.Equals(blk.GetForkVersion(), version.Electra()) {
+		var executionRequestsList []EncodedExecutionRequest
+		// If we're post-electra, we set execution requests.
+		var executionRequests *ExecutionRequests
+		executionRequests, err := body.GetExecutionRequests()
+		if err != nil {
+			return nil, err
+		}
```
calbera, line 78: super nit
```suggestion
		// If we're post-electra, we set execution requests.
		executionRequests, err := body.GetExecutionRequests()
		if err != nil {
			return nil, err
		}
		var executionRequestsList []EncodedExecutionRequest
```

calbera (approved):

beacon/blockchain/payload.go
```diff
@@ -93,19 +93,18 @@ func (s *Service) forceSyncUponFinalize(
 ) error {
 	// NewPayload call first to load payload into EL client.
 	executionPayload := beaconBlock.GetBody().GetExecutionPayload()
-	parentBeaconBlockRoot := beaconBlock.GetParentBlockRoot()
-	payloadReq := ctypes.BuildNewPayloadRequest(
-		executionPayload,
```
rezbera, line 98: Called `BuildNewPayloadRequest` when we need to call `BuildNewPayloadRequestFromFork`

PR #2638, opened 2025-03-30
feat(cli): Chain Spec as a Config File

rezbera:
Supports using chain spec as a config file.  Unblocks work with E2E testing and allows us to configure fork schedule without binary changes

To ensure node operators have backwards compatability, we keep the usage of environment variables and introduce a new environment variable that is used to specify that the chain spec is custom and must be loaded from a file.

We can consider removing this at some point and require purely a flag but in the interest of avoiding friction, we skip this.

To sync a mainnet node using the custom chain spec run

```sh
# This uses `cli/commands/server/types/mainnet_spec.toml`
make start-configurable

make start-geth-mainnet
```


cmd/beacond/defaults.go
```diff
@@ -35,7 +35,6 @@ func DefaultComponents() []any {
 		components.ProvideBlobProofVerifier,
 		components.ProvideChainService,
 		components.ProvideNode,
-		components.ProvideChainSpec,
```
rezbera, line 38: idk why this was being provided

rezbera, line 38: We can remove this as we supply it later

node-core/builder/builder.go
```diff
@@ -82,6 +87,7 @@ func (nb *NodeBuilder) Build(
 				logger,
 				db,
 				cmtCfg,
+				chainSpec,
```
rezbera, line 90: Supplying chainSpec here

cli/builder/options.go
```diff
@@ -54,3 +54,10 @@ func WithNodeBuilderFunc(nodeBuilderFunc servertypes.AppCreator) Opt {
 		cb.nodeBuilderFunc = nodeBuilderFunc
 	}
 }
+
+// WithChainSpecBuilderFunc sets the chainspec builder
+func WithChainSpecBuilderFunc(chainBuilderFunc servertypes.ChainSpecCreator) Opt {
+	return func(cb *CLIBuilder) {
+		cb.chainSpecBuilderFunc = chainBuilderFunc
```
rezbera, line 61: Doing a separate builder func allows us to use chainspec without building the whole node

cmd/beacond/main.go
```diff
@@ -61,15 +58,11 @@ func run() error {
 		),
 		// Set the Runtime Components to the Default.
 		clibuilder.WithComponents(
-			append(
-				clicomponents.DefaultClientComponents(),
-				// TODO: remove these, and eventually pull cfg and chainspec
-				// from built node
-				nodecomponents.ProvideChainSpec,
```
rezbera, line 68: Is now supplied as part of node building

cmd/beacond/main.go
```diff
@@ -61,15 +58,11 @@ func run() error {
...
-			),
+			clicomponents.DefaultClientComponents(),
 		),
 		// Set the NodeBuilderFunc to the NodeBuilder Build.
 		clibuilder.WithNodeBuilderFunc(nb.Build),
+		clibuilder.WithChainSpecBuilderFunc(types.CreateChainSpec),
```
rezbera, line 65: We supply the same func to the Node Builder and the CLI Builder. 

The CLI Builder does not always need the full node built when needing the chain spec, but the node builder always needs the chain spec builder

cli/commands/server/types/chainspec.go
```diff
@@ -0,0 +1,205 @@
...
+type ChainSpecCreator func(AppOptions) (chain.Spec, error)
+
+func CreateChainSpec(appOpts AppOptions) (chain.Spec, error) {
+	var (
+		chainSpec chain.Spec
+		err       error
+	)
+	switch os.Getenv(ChainSpecTypeEnvVar) {
+	case ConfigurableChainSpecType:
+		chainSpec, err = handleConfigurableChainSpec(appOpts)
```
rezbera, line 57: if `CHAIN_SPEC=configurable`, we process separately

rezbera: we could also just have it purely be flag based. im not fussed whatever team prefs

cli/commands/server/types/mainnet_spec.toml
shotes, line 1: This depends on expected use case - do we see config file usage as the primary way of supplying the chain spec going forward?

If so, can we move this to a top level directory and also add the other chain specs?

rezbera: As discussed in slack

> I think we should have the fallback in the binary for now, but allow using the file if it exists

calbera:
Noting some further changes that can/should be made after this PR:
- removing chain-spec as env var and instead reading it from the app.toml config
  - "chain-spec": possible values of "mainnet", "testnet", "devnet", "configurable"
  - "chain-spec-file": only read if value above is "configurable"
- fixing up kurtosis startup to allow configurable chain-specs via file in devnet tests (basically kurtosis should no longer need "devnet" chain spec and our e2e and actual devnets will all use "configurable" chain specs via files)

calbera:
merged as part of #2630, will open a new PR with some improvements

PR #2642, opened 2025-04-01
fix: return type in `GetPayloadV3`

barajeel:
`GetPayloadV3` was returning `nil` for a non-pointer type, causing potential runtime issues.
Now it correctly returns a pointer to `ctypes.BuiltExecutionPayloadEnv`.


rezbera:
change doesn't even compile.

PR #2644, opened 2025-04-02
feat(pectra): Timestamp based forking pectra e2e

shotes:
This sets the fork time for e2e tests to occur at 0, meaning the e2e tests will operate on the pectra fork.

PR #2645, opened 2025-04-02
feat(pectra): Add `engine_getPayloadV4`

shotes:
We MUST call `engine_getPayloadV4` when we are on the pectra fork according to [engine API spec updates](https://github.com/ethereum/execution-apis/blob/main/src/engine/prague.md#update-the-methods-of-previous-forks), even if our implementation is able to discriminate engine api responses under the hood for different forks. For some reason, geth and reth do not enforce that you use `getPayloadV4` like the specs expect. Erigon, however, enforces the usage of `getPayloadV4`, so we might as well conform to spec.

### DIFF
--- a/execution/client/ethclient/engine.go
+++ b/execution/client/ethclient/engine.go
@@ -40,11 +40,11 @@ func (s *Client) NewPayload(
 	ctx context.Context,
 	req ctypes.NewPayloadRequest,
 ) (*engineprimitives.PayloadStatusV1, error) {
+	forkVersion := req.GetForkVersion()
 	// Versions before Deneb are not supported for calling NewPayload.
-	if version.IsBefore(req.GetForkVersion(), version.Deneb()) {
+	if version.IsBefore(forkVersion, version.Deneb()) {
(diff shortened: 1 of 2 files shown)


execution/client/ethclient/engine.go
```diff
@@ -169,6 +173,19 @@ func (s *Client) GetPayloadV3(
 	return result, nil
 }
 
+// GetPayloadV4 calls the engine_getPayloadV4 method via JSON-RPC.
+func (s *Client) GetPayloadV4(
```
shotes, line 177: The `GetPayloadV3` and `GetPayloadV4` functions are identical other than the change in the constant "method name" for the call. However, I am keeping them as completely separate implementations for consistency. I think it makes sense to keep them distinctly separate.

rezbera (approved):
LGTM. Would encourage test cases if possible

shotes:
Confirmed working with e2e tests when combined with https://github.com/berachain/beacon-kit/pull/2644

calbera (approved):

PR #2647, opened 2025-04-03
[2/2] chore(ssz): merkle libraries cleanup

calbera:
- remove unused ssz lib code
- aggregate merkle lib code in 1 location
- etc.


primitives/merkle/tree.go
calbera, line 1: this `Tree` implementation is used in DA for blob sidecar verification

primitives/merkle/object_path.go
calbera, line 1: object_path is used with GeneralizedIndex (index.go) and the encoding/ssz/schema package to create SSZ structures. It's very helpful for visualizing SSZ structs as Merkle-ized trees and currently used in unit tests to guarantee conformance. Hence its kept

shotes (approved):
Wow. So much unused stuff, glad it's getting removed. LGTM

PR #2648, opened 2025-04-03
feat(node-api): get block header api

nidhi-singh02:


node-api/handlers/beacon/header.go
```diff
@@ -57,22 +59,27 @@ func (h *Handler) GetBlockHeaderByID(c handlers.Context) (any, error) {
...
 	}
 	return beacontypes.NewResponse(&beacontypes.BlockHeaderResponse{
-		Root:      header.GetBodyRoot(),
+		Root:      root, // This is root hash of entire beacon block
 		Canonical: true,
 		Header: &beacontypes.SignedBeaconBlockHeader{
 			Message:   beacontypes.BeaconBlockHeaderFromConsensus(header),
-			Signature: "", // TODO: implement
+			Signature: (&crypto.BLSSignature{}).String(), // TODO: implement
 		},
```
nidhi-singh02, line 83: Signature needs to be implemented. @berachain/core Any thoughts on this. This is the point that I brought up in the sync call.

nidhi-singh02:
Things to-do:
- E2E tests from validator related API via genesis query are failing, need to debug.
- Need to see how to get the block root at the genesis. Currently it is all zero.

node-api/handlers/beacon/header.go
```diff
@@ -57,22 +61,36 @@ func (h *Handler) GetBlockHeaderByID(c handlers.Context) (any, error) {
...
-		return nil, err
+	var header *ctypes.BeaconBlockHeader
+	var root common.Root
+
+	switch req.BlockID {
+	case utils.StateIDGenesis, "0":
+		header = h.backend.GenesisBlockHeader()
+		root = h.backend.GenesisBlockRoot()
+	default:
+		slot, errInSlot := utils.SlotFromBlockID(req.BlockID, h.backend)
```
nidhi-singh02, line 74: Here handling the special genesis case when we query using "genesis" or slot number 0.

node-api/handlers/utils/id.go
```diff
@@ -112,13 +112,14 @@ func IsTimestampIDPrefix(timestampID string) bool {
 }
 
 // slotFromStateID returns a slot number from the given state ID.
-// TODO: This pattern does not allow us to query block 0. Genesis points to block 1.
+// In case of genesis, we will return 1 as the slot number. This is to ensure
+// validator API endpoints return a valid slot number for genesis.
```
nidhi-singh02, line 116: Here keeping the genesis slot as 1 (as it was earlier, just removed the Genesis constant) so that the APIs that use this function can work. As in if we return 0, we will get the latest slot.

Will have to find a way to handle the special genesis case for all these APIs, as part of separate PR as it is not related to Block header.

nidhi-singh02:
Working on the genesis cases for validator and validator balance endpoints in a separate PR, which we will merge into this PR.

nidhi-singh02:
Here's the PR https://github.com/berachain/beacon-kit/pull/2707 for handling genesis cases for validator and validator balance endpoints which will be merged into this current PR.

abi87:
Replaced by https://github.com/berachain/beacon-kit/pull/2930 (as far as genesis state is concerned)

PR #2650, opened 2025-04-03
Fix - Workflow Error on Starting Kurtosis

gummybera:
>   Caused by: Tried pulling image 'timberio/vector:0.31.0-debian' with platform '' but failed
>   Caused by: Error response from daemon: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit

Docker anon API limit reached - authenticate with account to fix this.


calbera (commented):
utACK

.github/workflows/pipeline.yml
```diff
@@ -174,6 +174,7 @@ jobs:
           echo "deb [trusted=yes] https://apt.fury.io/kurtosis-tech/ /" | sudo tee /etc/apt/sources.list.d/kurtosis.list
           sudo apt update
           sudo apt install kurtosis-cli=$(go list -m -f '{{.Version}}' github.com/kurtosis-tech/kurtosis/api/golang | sed 's/^v//') -y
+          echo "${{ secrets.DOCKERHUB_PASSWORD }}" | docker login --username "${{ secrets.DOCKERHUB_USERNAME }}" --password-stdin
```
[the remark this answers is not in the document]
fridrik01: We should already be logged in via the previous step `Login to Docker Hub` (see above)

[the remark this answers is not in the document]
fridrik01: I made a [PR](https://github.com/berachain/beacon-kit/pull/2607) few weeks ago which checked if this `DOCKERHUB_PASSWORD` is set and then tries to login

fridrik01 (changes requested):
@gummybera we should already be logged in via the previous `Login to Docker Hub` step

.github/workflows/pipeline.yml
```diff
@@ -151,12 +151,11 @@ jobs:
           fi
         shell: bash
       - name: Login to Docker Hub
-        uses: docker/login-action@v3
+        uses: docker/login-action@v3.4.0
         if: steps.check-docker-credentials.outputs.has_credentials == 'true'
         with:
-          registry: https://index.docker.io/v1/
```
fridrik01, line 157: We should just switch to ghcr.io and use the same credentials as done here below in `Authenticate to GitHub Container Registry`

fridrik01:
> @fridrik01 thanks for the context, made a few observations/updates:
> 
> * The registry URL `https://index.docker.io/v1/` is deprecated and returns a 404. Strangely, the action reports a successful login in spite of this.
> * Image pull activity on our account correlates weakly with the actions performed on github, so a good chance that the auth wasn't succeeding.
> * Updated to a more recent release of the docker login action, and implemented an access token for auth as recommended.
> * The action will now print `docker info` before starting kurtosis, which will confirm the state of authentication.
> 
> Let's rerun the workflow with these changes, this should help move it toward resolution.

Propose we try switching to ghcr like we already to when pushing the beacon image, we can use the same credentials and I believe ghcr has higher rate limits, especially inside github actions

fridrik01:
@gummybera I tried login to ghcr in https://github.com/berachain/beacon-kit/pull/2662. I verified the login works from checking the docker config file in the workflow.

.github/workflows/pipeline.yml
```diff
@@ -174,6 +173,7 @@ jobs:
           echo "deb [trusted=yes] https://apt.fury.io/kurtosis-tech/ /" | sudo tee /etc/apt/sources.list.d/kurtosis.list
           sudo apt update
           sudo apt install kurtosis-cli=$(go list -m -f '{{.Version}}' github.com/kurtosis-tech/kurtosis/api/golang | sed 's/^v//') -y
+          docker info
```
fridrik01, line 176: I don't think `docker info` shows whether we are logged in or not (which is strange, at least when I looked at it). To check this AFAICT we need to look at the `auths` section in `~/.docker/config.json`

.github/workflows/pipeline.yml
gummybera: Not all, the action is defined for dockerhub. I was able to test this independently

.github/workflows/pipeline.yml
gummybera: `docker info` shows a masked username in Github Actions when authenticated, like so:

```
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Username: ***
 Experimental: false
```

gummybera:
@fridrik01 thanks for the feedback, replies inline.

I've been trying to reproduce this behavior with limited success. What's complicating the issue is that the deb package repo used for Kurtosis is broken - the workflows have been making do with an older image sourced from another repo. ( latest `v1.6.0`, in use `v1.4.3`)

Given this situation, the best path forward might be to manually pull the dependency images before kurtosis runs. I've included another commit on this PR to that effect. This will hold good until the Kurtosis repository is fixed and a newer package is available, at which time we can revisit the changes if the problem persists. In the meantime, these changes will prevent any possible rate limiting on Kurtosis. Does this sound good?

.github/workflows/pipeline.yml
```diff
@@ -174,6 +173,10 @@ jobs:
...
+          for img in kurtosistech/engine:1.4.3 timberio/vector:0.31.0-debian traefik:2.10.6 alpine:3.17; do
```
fridrik01, line 177: maybe add some context why we have this here?

fridrik01 (approved):
lets try it!

.github/workflows/pipeline.yml
gummybera: docker info - for debugging
the `for` loop - pulls docker images manually outside of kurtosis to eliminate any potential issues there

gummybera:
ty @fridrik01 ! Could you also approve the workflow?

gummybera:
yup can see the difference. The docker auth is active within this workflow step, but doesn't seem to work from within kurtosis. Hopefully this fixes it, we will keep an eye out.

**Before:**
```
INFO[2025-05-08T07:24:17Z] Starting Kurtosis engine from image 'kurtosistech/engine:1.4.3'... 
INFO[2025-05-08T07:24:17Z] No Kurtosis engine was found; attempting to start one... 
INFO[2025-05-08T07:24:17Z] Starting the centralized logs components...  
INFO[2025-05-08T07:24:17Z] Pulling image 'timberio/vector:0.31.0-debian' 
Error:  An error occurred running command 'start'
[stack trace]
  Caused by: Error response from daemon: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
```

**After:**
```
INFO[2025-05-08T15:33:27Z] Starting Kurtosis engine from image 'kurtosistech/engine:1.4.3'... 
INFO[2025-05-08T15:33:27Z] No Kurtosis engine was found; attempting to start one... 
INFO[2025-05-08T15:33:27Z] Starting the centralized logs components...  
INFO[2025-05-08T15:33:27Z] Centralized logs components started.         
INFO[2025-05-08T15:33:28Z] Reverse proxy started.                       
INFO[2025-05-08T15:33:29Z] Successfully started Kurtosis engine         
INFO[2025-05-08T15:33:29Z] Kurtosis engine started                      
```

.github/workflows/pipeline.yml
fridrik01: yeah I meant adding the context in the code so we wont forget what its doing

PR #2655, opened 2025-04-08
[5/n] feat(pectra): Withdrawals Support 

rezbera:
### Context
This PR introduces support for Withdrawals. We largely follow the Ethereum specification, but relevant deviations can be found in our fork of the consensus specs here https://github.com/berachain/consensus-specs/pull/1/files

The lifecycle of a deposit is as follows and can be seen in simulated tests such as `TestFullLifecycle_WithRequests_IsSuccessful`.

### Technicals:

1. User submits a Withdrawal Request, this is a normal EVM transaction. 
2. Execution Layer propagates the Withdrawal Request to the Consensus Layer as part of `GetPayload`.
3. The Block Builder will include the Withdrawal Request as part of the block - a necessary prerequisite for block validity. 

Now we dive deeper into the actual State Transition that validators will execute:
4. As part of `ProcessBlock` -> `processOperations` -> `processWithdrawalRequest`, the Withdrawal Request will be processed, there are two paths it can take:

#### FullExitRequest

A full exit request is one where the user sets the value `0` as their desired withdrawal amount. If it's a full exit request, we initiate a validator exit where the `exitEpoch` is set to the next epoch and the `withdrawableEpoch` is `exitEpoch + MinValidatorWithdrawabilityDelay`. 

When the `exitEpoch` is reached, as part of `processEpoch`, `getActiveVals` will be called which will no longer have the validator that initiated the exit, updating comet. 

Their balance will be withdrawn once `withdrawableEpoch` is reached as part of `processWithdrawals`.
   
#### PartialExitRequest

A partial exit request is one where the user sets a non-`0` value as their desired withdrawal amount. As part of `processWithdrawalRequest`, the a partial withdrawal will hit the `processPartialWithdrawal` logic path. a `PendingPartialWithdrawal` will be created and persisted in the BeaconState, where the `WithdrawableEpoch` is set to `nextEpoch + MinValidatorWithdrawabilityDelay`. 

Their balance will be withdrawn once `withdrawableEpoch` is reached as part of `processWithdrawals`, which will fetch `pendingPartialWithdrawal`s from the BeconState.

If the withdrawal put's the validators balance below `EJECTION_BALANCE`, an exit will be initiated as part of `processRegistryUpdates`. 

### Limitations:

There are some design limitatations:

- If a validator only wants to stake 250K BERA but the validator set cap has been reached, they can stake > 250K BERA, then partial withdraw back down to 250K BERA.

- If the `PENDING_PARTIAL_WITHDRAWALS_LIMIT` has been hit, a valid withdrawal will silently be rejected, which is not observable from a smart contract's perspective.


state-transition/core/state_processor_validators.go
```diff
@@ -145,16 +148,15 @@ func (sp *StateProcessor) processValidatorSetCap(st *statedb.StateDB) error {
...
-		valToEject.SetWithdrawableEpoch(nextEpoch + 1)
 		idx, err = st.ValidatorIndexByPubkey(valToEject.GetPubkey())
 		if err != nil {
 			return fmt.Errorf(
 				"validators cap, failed loading validator index: %w",
 				err,
 			)
 		}
-		if err = st.UpdateValidatorAtIndex(idx, valToEject); err != nil {
+		err = sp.InitiateValidatorExit(st, idx)
```
rezbera, line 158: using this will likely need to be gated by hard fork as the logic in `InitiateValidatorExit` is changing

shotes: added in https://github.com/berachain/beacon-kit/pull/2655/commits/d9f39bedcf786efe4cbb167b9e5d7c47c75df504

rezbera:
Going to rebase this onto https://github.com/berachain/beacon-kit/pull/2656

chain/data.go
```diff
@@ -140,8 +140,6 @@ type SpecData struct {
 	FieldElementsPerBlob uint64 `mapstructure:"field-elements-per-blob"`
 	// BytesPerBlob denotes the size of EIP-4844 blobs in bytes.
 	BytesPerBlob uint64 `mapstructure:"bytes-per-blob"`
-	// KZGCommitmentInclusionProofDepth is the depth of the KZG inclusion proof.
-	KZGCommitmentInclusionProofDepth uint64 `mapstructure:"kzg-commitment-inclusion-proof-depth"`
```
calbera, line 144: unused in chain spec

consensus-types/types/body.go
```diff
@@ -96,7 +96,7 @@ type BeaconBlockBody struct {
 	blsToExecutionChanges []*BlsToExecutionChange
 	// BlobKzgCommitments is the list of KZG commitments for the EIP-4844 blobs.
 	BlobKzgCommitments []eip4844.KZGCommitment
-	// executionRequests is introduced in electra. We keep this private so that it must go through Getter/Setter
+	// executionRequests is introduced in Electra. We keep this private so that it must go through Getter/Setter
```
calbera, line 99: just noting that this is an anti-pattern from the rest of consensus types. fields are kept private iff they are unused and satisfy `EnforceUnused()`. Unsure if we should keep this field public or allow private to imply that its not _always_ used.

config/spec/defaults.go
```diff
@@ -78,10 +78,12 @@ const (
...
 	defaultBytesPerBlob                     = 131072
-	defaultKZGCommitmentInclusionProofDepth = 17
 
 	// Berachain values.
 	defaultValidatorSetCap      = 256
 	defaultEVMInflationAddress  = "0x0000000000000000000000000000000000000000"
 	defaultEVMInflationPerBlock = 0
+
+	// Electra values.
+	defaultMinValidatorWithdrawabilityDelay = 256
```
rezbera, line 88: 256 epochs at 384 seconds per epoch is ~27 hours.

config/spec/mainnet.go
```diff
@@ -151,10 +150,13 @@ func MainnetChainSpecData() *chain.SpecData {
 		// Deneb1 values.
 		EVMInflationAddressDeneb1:  common.NewExecutionAddressFromHex(mainnetEVMInflationAddressDeneb1),
 		EVMInflationPerBlockDeneb1: mainnetEVMInflationPerBlockDeneb1,
+
+		// Electra values.
+		MinValidatorWithdrawabilityDelay: defaultMinValidatorWithdrawabilityDelay,
```
calbera, line 155: TODO: determine correct value here. As @rezbera noted: 256 epochs at 384 seconds per epoch is ~27 hours.

config/spec/defaults.go
calbera: will keep this here as this file has ETH2.0 values as defaults

state-transition/core/state_processor_validators.go
```diff
@@ -145,16 +147,15 @@ func (sp *StateProcessor) processValidatorSetCap(st *statedb.StateDB) error {
 	var idx math.ValidatorIndex
 	for li := range uint64(len(nextEpochVals)) - validatorSetCap {
 		valToEject := nextEpochVals[li]
-		valToEject.SetExitEpoch(nextEpoch)
-		valToEject.SetWithdrawableEpoch(nextEpoch + 1)
```
rezbera, line 149: Logic moved to `InitiateValidatorExit`, in line with spec

state-transition/core/state_processor_validators.go
```diff
@@ -45,28 +45,27 @@ func (sp *StateProcessor) processRegistryUpdates(st *statedb.StateDB) error {
 	}
 
 	currEpoch := sp.cs.SlotToEpoch(slot)
-	nextEpoch := currEpoch + 1
-
-	minEffectiveBalance := math.Gwei(
-		sp.cs.EjectionBalance() + sp.cs.EffectiveBalanceIncrement(),
-	)
+	activationEpoch := currEpoch + 1
+	minEffectiveBalance := math.Gwei(sp.cs.MinActivationBalance())
```
rezbera, line 49: We can change this without a hard fork as `sp.cs.MinActivationBalance()` == sp.cs.EjectionBalance() + sp.cs.EffectiveBalanceIncrement()`

state-transition/core/state_processor_validators.go
```diff
@@ -45,28 +45,27 @@ func (sp *StateProcessor) processRegistryUpdates(st *statedb.StateDB) error {
 	}
 
 	currEpoch := sp.cs.SlotToEpoch(slot)
-	nextEpoch := currEpoch + 1
-
-	minEffectiveBalance := math.Gwei(
-		sp.cs.EjectionBalance() + sp.cs.EffectiveBalanceIncrement(),
-	)
+	activationEpoch := currEpoch + 1
+	minActivationBalance := math.Gwei(sp.cs.MinActivationBalance())
```
rezbera, line 49: Does not require a hardfork as `sp.cs.EjectionBalance() + sp.cs.EffectiveBalanceIncrement() == sp.cs.MinActivationBalance()`

node-api/backend/mocks/beacon_state.mock.go
```diff
@@ -1,2224 +0,0 @@
-// Code generated by mockery v2.52.1. DO NOT EDIT.
```
rezbera, line 1: Unused AFAIK

node-core/components/interfaces.go
```diff
@@ -482,7 +482,6 @@ type (
 	// ReadOnlyWithdrawals only has read access to withdrawal methods.
 	ReadOnlyWithdrawals interface {
 		EVMInflationWithdrawal(math.Slot) *engineprimitives.Withdrawal
-		ExpectedWithdrawals() (engineprimitives.Withdrawals, error)
```
rezbera, line 485: Unused AFAIK

state-transition/core/state_processor_staking_test.go
```diff
@@ -506,19 +506,18 @@ func TestTransitionMaxWithdrawals(t *testing.T) {
...
 //
 //nolint:paralleltest // uses envars
 func TestTransitionHittingValidatorsCap_ExtraSmall(t *testing.T) {
 	cs := setupChain(t)
 	sp, st, ds, ctx, _, _ := statetransition.SetupTestState(t, cs)
 
 	var (
-		maxBalance      = math.Gwei(cs.MaxEffectiveBalance())
-		ejectionBalance = math.Gwei(cs.EjectionBalance())
-		minBalance      = ejectionBalance + math.Gwei(cs.EffectiveBalanceIncrement())
```
rezbera, line 520: This needed to be changed as the `MinBalance` on devnet adds to up to `17 ETH` which is incorrect. It should be `32 ETH`.

testing/quick/compare_test.go
```diff
@@ -40,7 +40,7 @@ var (
 	spec = zspec.Mainnet
 )
 
-// TODO(pectra): None of the iterations makes it to `HashTreeRoot` as they all get caught by the nil checks.
```
rezbera, line 43: Fixing this is not a prerequisite for pectra

state-transition/core/state_processor_validators.go
rezbera: However, on Devnet this would be a hardfork due to incorrect values - but we don't care about that since devnet is ephemeral

rezbera:
Marking this as ready to review as it's functionally complete, however, I'm still adding test cases

chain/spec.go
```diff
@@ -38,6 +38,9 @@ type BalancesSpec interface {
 	// EffectiveBalanceIncrement returns the increment of balance used in reward
 	// calculations.
 	EffectiveBalanceIncrement() uint64
+
+	// MinActivationBalance returns the maximum balance counted in rewards calculations in Gwei.
```
abi87, line 42: comment may be off?

cli/commands/server/types/mainnet_spec.toml
```diff
@@ -68,3 +68,7 @@ evm-inflation-per-block = 0
 # Deneb1 value changes
 evm-inflation-address-deneb-one = "0x656b95E550C07a9ffe548bd4085c72418Ceb1dba"
 evm-inflation-per-block-deneb-one = 5750000000
+
+# Electra values
+min-activation-balance = 250000000000000
```
abi87, line 73: super nit: consider exponential notation or inserting `_` every three zeros

config/spec/devnet.go
```diff
@@ -55,6 +55,8 @@ const (
 	// devnetEVMInflationPerBlockDeneb1 is the amount of native EVM balance (in units
 	// of Gwei) to be minted per EL block after the Deneb1 fork.
 	devnetEVMInflationPerBlockDeneb1 = 11 * params.GWei
+
+	devnetMinValidatorWithdrawabilityDelay = 32
```
abi87, line 59: these are epochs right? delay is 32 epochs?

node-core/components/interfaces.go
abi87: Also missing timestamp parameter. Either fix or drop. I vote drop

state-transition/core/state/statedb.go
```diff
@@ -72,20 +74,24 @@ func (s *StateDB) DecreaseBalance(idx math.ValidatorIndex, delta math.Gwei) erro
...
 
 // ExpectedWithdrawals as defined in the Ethereum 2.0 Specification:
-// https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#new-get_expected_withdrawals
+// https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#new-get_expected_withdrawals
 //
 // NOTE: This function is modified from the spec to allow a fixed withdrawal
 // (as the first withdrawal) used for EVM inflation.
-func (s *StateDB) ExpectedWithdrawals(timestamp math.U64) (engineprimitives.Withdrawals, error) {
+//
+//nolint:gocognit,funlen // TODO(pectra).
```
abi87, line 82: nit: probably the TODO can go?

state-transition/core/state/statedb.go
```diff
@@ -167,7 +183,75 @@ func (s *StateDB) ExpectedWithdrawals(timestamp math.U64) (engineprimitives.With
...
+	uint64, // withdrawalIndex
+	uint64, // processedPartialWithdrawals
+	error,
+) {
+	// By this point, if we're post-Electra, the fork version on the BeaconState will have been set as part of `PrepareStateForFork`.
+	// We sanity check this by enforcing the state fork and timestamp version are equal.
+	// We cannot do this check before Electra as Deneb1 did not have a beacon state fork version set.
+	stateFork, forkErr := s.GetFork()
+	if forkErr != nil {
+		return nil, 0, 0, forkErr
```
abi87, line 205: consider improving error expressiveness with something like:
`return nil, 0,0, fmt.Errorf("consumePendingPartialWithdrawals: failed retrieving fork: %w", forkErr)`
here and in the rest of the file

state-transition/core/state/statedb.go
```diff
@@ -167,7 +183,75 @@ func (s *StateDB) ExpectedWithdrawals(timestamp math.U64) (engineprimitives.With
...
+	ppWithdrawals, getErr := s.GetPendingPartialWithdrawals()
+	if getErr != nil {
+		return nil, 0, 0, getErr
+	}
+	processedPartialWithdrawals := uint64(0)
+
+	for _, withdrawal := range ppWithdrawals {
+		if withdrawal.WithdrawableEpoch > epoch || len(withdrawals) == int(constants.MaxPendingPartialsPerWithdrawalsSweep) {
+			// If the first withdrawal in the queue is not withdrawable, then all subsequent withdrawals will also be in later
+			// epochs and hence are not withdrawable, so we can break early.
```
abi87, line 219: awesome comments here and in ther rest of the file!

state-transition/core/state_processor_withdrawals.go
```diff
@@ -49,11 +53,16 @@ func (sp *StateProcessor) processWithdrawals(
...
-	expectedWithdrawals, err := st.ExpectedWithdrawals(blk.GetTimestamp())
+	expectedWithdrawals, processedPartialWithdrawalsCount, err := st.ExpectedWithdrawals(blk.GetTimestamp())
 	if err != nil {
 		return err
 	}
 
+	if version.IsBefore(blk.GetForkVersion(), version.Electra()) && processedPartialWithdrawalsCount != 0 {
+		// This should never happen as we would not add pending partial withdrawals to the beacon state before electra.
+		return fmt.Errorf("there must be no partial withdrawals before electra but found %d", processedPartialWithdrawalsCount)
+	}
```
abi87, line 64: this seems redundant to me since in `consumePendingPartialWithdrawals` (called by `ExpectedWithdrawals`) we already check that `consumePending` is called only if Electra is active. I would drop one of them (possibly this since the other errs eariler)

state-transition/core/state_processor_withdrawals.go
```diff
@@ -136,3 +163,185 @@ func (sp *StateProcessor) processWithdrawals(
...
+		// Note that we do not return error on invalid requests as it's a user error and invalid withdrawal requests are simply skipped.
+		// Since we're interacting with the KV Store, i/o errors could occur, but we expect these to be sparse enough
+		// that it won't impact consensus. We could consider returning error on codec.ErrEncoding rather than silencing here.
+		return nil
+	}
+	if validator == nil {
+		// This should never occur as we expect ErrNotFound if the validator is not found. As such, we return error here.
+		sp.logger.Warn("Validate withdrawal return nil validator", withdrawalFields(withdrawalRequest, nil)...)
+		return errors.New("unexpected nil validator")
+	}
```
abi87, line 198: this seems redundant since in `validateWithdrawal` we already err explicitly if validator is nil

state-transition/core/state_processor_withdrawals.go
```diff
@@ -136,3 +163,185 @@ func (sp *StateProcessor) processWithdrawals(
...
+	// Verify exit has not been initiated
+	if validator.GetExitEpoch() != math.Epoch(constants.FarFutureEpoch) {
+		return errors.New("withdrawal already initiated")
+	}
+	// Verify the validator has been active long enough
+	// In the spec, config.SHARD_COMMITTEE_PERIOD is added as well, but we ignore this since
+	// it's related to ETH data shards which is no longer planned.
+	if currentEpoch < validator.ActivationEpoch {
+		return errors.New("validator not active long enough")
+	}
```
abi87, line 331: IIRC validator is active only if `epoch <= validator.ActivationEpoch`. Since we already check above that the validator is active, this check seems redundant

state-transition/core/state_processor_exits.go
```diff
@@ -0,0 +1,76 @@
...
+		// Return if validator already initiated exit.
+		if validator.GetExitEpoch() != math.Epoch(constants.FarFutureEpoch) {
+			return nil
+		}
+		slot, slotErr := st.GetSlot()
+		if slotErr != nil {
+			return slotErr
+		}
+		nextEpoch := sp.cs.SlotToEpoch(slot) + 1
+		exitEpoch = nextEpoch
```
abi87, line 52: may be worth commenting that we keep on not having a cap on validators churn, both upon activation and exit

state-transition/core/state_processor_exits.go
```diff
@@ -0,0 +1,76 @@
...
+		return err
+	}
+	// By this point, the state should be correctly updated with the fork.
+	fork, err := st.GetFork()
+	if err != nil {
+		return err
+	}
+	var withdrawableEpoch, exitEpoch math.Epoch
+	if version.EqualsOrIsAfter(fork.CurrentVersion, version.Electra()) {
+		// Return if validator already initiated exit.
```
abi87, line 43: nit: I read this as "make sure we exit validators only once" i.e. never process exit twice on a validator

state-transition/core/state_processor_exits.go
```diff
@@ -0,0 +1,76 @@
...
+			return nil
+		}
+		slot, slotErr := st.GetSlot()
+		if slotErr != nil {
+			return slotErr
+		}
+		nextEpoch := sp.cs.SlotToEpoch(slot) + 1
+		exitEpoch = nextEpoch
+		// The withdrawable Epoch is `MinValidatorWithdrawabilityDelay` epoch's after `exitEpoch`.
+		withdrawableEpoch = math.Epoch(uint64(exitEpoch) + sp.cs.MinValidatorWithdrawabilityDelay())
```
abi87, line 54: I am not even sure how helpful this is. I understand why we need this, I just wonder if we really need this now?
We may defer it all to the point where we introduce slashing?

config/spec/devnet.go
rezbera: correct. clarified in comment 14fff97e5519ad8d176fe8d9d195a0dba51880cf

cli/commands/server/types/mainnet_spec.toml
rezbera: 1832c3c37d47b04d68593099d3127e0badffc590

state-transition/core/state/statedb.go
rezbera: a88b814a8ad6157e1009396be810a92823f9833c

state-transition/core/state/statedb.go
rezbera: c7f41ac1da79003507f91e514b6120db0b5ac3c3

state-transition/core/state/statedb.go
rezbera: This is where we remove items from the pendingPartialWithdrawals, where we effectively remove from the beginning like a sliding window. 

https://github.com/berachain/beacon-kit/blob/3992711351bd37d17b0757e8ce2e4bcacb7de789/state-transition/core/state_processor_withdrawals.go#L104-L120

This is where we append to the pendingPartialWithdrawals https://github.com/berachain/beacon-kit/blob/3992711351bd37d17b0757e8ce2e4bcacb7de789/state-transition/core/state_processor_withdrawals.go#L257-L277

The main thing to note from the second link is that the `PendingPartialWithdrawals` are effectively sorted by `withdrawableEpoch` as we only append to the list, and any new `pendingWithdrawals` requests will always go in a >= WithdrawableEpoch than previous ones.

state-transition/core/state_processor_withdrawals.go
rezbera: nilaway is blind to it. further comments in 34ee2a92cea28c94f409c70ef45e59793c07f275

state-transition/core/state_processor_withdrawals.go
rezbera: It's mostly defensive programming, but perhaps this one falls into redundancy. Removed 8f3515fa99e643b6b7ff19b8d3081492437cc0a7

state-transition/core/state_processor_withdrawals.go
rezbera: This is largely taken from the spec

```python
    # Verify the validator is active
    if not is_active_validator(validator, get_current_epoch(state)):
        return
    # Verify exit has not been initiated
    if validator.exit_epoch != FAR_FUTURE_EPOCH:
        return
    # Verify the validator has been active long enough
    if get_current_epoch(state) < validator.activation_epoch + config.SHARD_COMMITTEE_PERIOD:
        return

```
If we removed this check, there is a functional difference.

For example, if `currentEpoch = 5` and `activationEpoch = 5`, then the `IsActive` check would pass but the `currentEpoch < validator.ActivationEpoch` check would fail. Hence a functional difference i'd rather not introduce.

state-transition/core/state_processor_exits.go
rezbera: 39570a77eba3489d89f0a439a837fede882f2dcf

state-transition/core/state_processor_withdrawals.go
abi87: So IIUC the point here it keep the validator active at least one epoch before allowing even a partial withdrawal?
Given we're taking `config.SHARD_COMMITTEE_PERIOD` the condition is our codebase is equivalent to `if currentEpoch == validator.ActivationEpoch  {` right?

rezbera: > So IIUC the point here it keep the validator active at least one epoch before allowing even a partial withdrawal?

That's my takeaway from the spec behaviour.

Given the python below (i.e. without `SHARD_COMMITTEE_PERIOD`)

```python
get_current_epoch(state) < validator.activation_epoch
```

The equivalent would be what's in the codebase at the moment. i.e.
```go
	if currentEpoch < validator.ActivationEpoch {
		return errors.New("validator not active long enough")
	}
```

rezbera: I can't give a concrete reason for why the spec would want at least one active epoch before allowing partial withdrawals - it's not well explained in the spec so we'd have to reason about it ourselves confidently before making a difference

state-transition/core/state_processor_withdrawals.go
```diff
@@ -136,3 +155,186 @@ func (sp *StateProcessor) processWithdrawals(
...
+	balance, err := st.GetBalance(index)
+	if err != nil {
+		return err
+	}
+
+	pendingBalanceToWithdraw := ctypes.PendingPartialWithdrawals(pendingWithdrawals).PendingBalanceToWithdraw(index)
+
+	hasExcess := balance > minActivationBalance+pendingBalanceToWithdraw
+
+	if validator.HasCompoundingWithdrawalCredential() && hasSufficient && hasExcess {
```
abi87, line 249: nit: should the request fail for `!hasExcess` or `!hasSufficient` should we log an info level message to simplify debugging?

state-transition/core/state_processor_withdrawals.go
```diff
@@ -136,3 +155,186 @@ func (sp *StateProcessor) processWithdrawals(
...
+
+// verifyWithdrawalConditions checks additional conditions like active status, exit not initiated, and minimal activation period.
+func verifyWithdrawalConditions(chainSpec ChainSpec, st *state.StateDB, validator *ctypes.Validator) error {
+	slot, err := st.GetSlot()
+	if err != nil {
+		return err
+	}
+	currentEpoch := chainSpec.SlotToEpoch(slot)
+	// Verify the validator is active
+	if !validator.IsActive(currentEpoch) {
```
abi87, line 312: not for myself: this covers the case where a validator eviction happens while a withdrawal request is being submitted. It does not cover the case where the withdrawal request is already accepted and the validator is evicted before the request is processed.

state-transition/core/state/statedb.go
```diff
@@ -167,7 +184,75 @@ func (s *StateDB) ExpectedWithdrawals(timestamp math.U64) (engineprimitives.With
...
+			break
+		}
+		validator, err := s.ValidatorByIndex(withdrawal.ValidatorIndex)
+		if err != nil {
+			return nil, 0, 0, err
+		}
+		minActivationBalance := math.Gwei(s.cs.MinActivationBalance())
+		hasSufficientEffectiveBalance := validator.GetEffectiveBalance() >= minActivationBalance
```
abi87, line 228: this cover the case where a validator is evicted after a withdrawal request has been submitted but before it is processed. In this case the request is simply ignored

abi87: consider adding logging for requests ignored here. It would help debugging

chain/data.go
```diff
@@ -166,4 +164,12 @@ type SpecData struct {
 	// EVMInflationPerBlockDeneb1 is the amount of native EVM balance (in Gwei) to be
 	// minted to the EVMInflationAddressDeneb1 via a withdrawal every block in the Deneb1 fork.
 	EVMInflationPerBlockDeneb1 uint64 `mapstructure:"evm-inflation-per-block-deneb-one"`
+
+	// Electra Values
+	//
+	// MinActivationBalance [New in Electra:EIP7251] Minimum balance for a validator to become active
+	MinActivationBalance uint64 `mapstructure:"min-activation-balance"`
```
calbera, line 171: with this introduced, we no longer need EjectionBalance. Can we remove it entirely?

state-transition/core/state_processor_withdrawals.go
rezbera: 584c1a63b0f8659e8e1f18bb106f409e21bb80bd

config/spec/mainnet.go
```diff
@@ -88,14 +87,17 @@ const (
 	// mainnetEVMInflationPerBlockDeneb1 is the amount of native EVM balance (in Gwei) to be
 	// minted to the EVMInflationAddressDeneb1 via a withdrawal every block in the Deneb1 fork.
 	mainnetEVMInflationPerBlockDeneb1 = 5.75 * params.GWei
+
+	// mainnetMinValidatorWithdrawabilityDelay is the number of epochs of delay epochs of delay for a balance to be withdrawable.
+	// 256 Epochs equates to roughly ~27 hours of withdrawal delay. This gives us room to emergency fork if needed.
+	mainnetMinValidatorWithdrawabilityDelay = 256
```
calbera, line 93: nit: use default when its the same
```suggestion
	mainnetMinValidatorWithdrawabilityDelay = defaultMinValidatorWithdrawabilityDelay
```

state-transition/core/state/statedb.go
rezbera: hmm, looks like `statedb` doesn't have a logger available oddly. will have to see if I can refactor it to get logger

state-transition/core/state/statedb.go
```diff
@@ -167,7 +183,66 @@ func (s *StateDB) ExpectedWithdrawals(timestamp math.U64) (engineprimitives.With
...
+	withdrawals engineprimitives.Withdrawals,
+	withdrawalIndex uint64,
+) (
+	[]*engineprimitives.Withdrawal,
+	uint64, // withdrawalIndex
+	uint64, // processedPartialWithdrawals
+	error,
+) {
+	// By this point, if we're post-Electra, the fork version on the BeaconState will have been set as part of `PrepareStateForFork`.
+	// This will fail if the fork version is not set.
```
calbera, line 200:
```suggestion
	// This will fail if the state has not been prepared for a post-Electra fork version.
```

state-transition/core/state/statedb.go
```diff
@@ -96,17 +102,27 @@ func (s *StateDB) ExpectedWithdrawals(timestamp math.U64) (engineprimitives.With
...
 
 	totalValidators, err := s.GetTotalValidators()
 	if err != nil {
-		return nil, err
+		return nil, 0, err
+	}
+
+	// [New in Electra:EIP7251] Consume pending partial withdrawals
+	timestampFork := s.cs.ActiveForkVersionForTimestamp(timestamp)
+	if version.EqualsOrIsAfter(timestampFork, version.Electra()) {
```
calbera, line 120:
```suggestion
	forkVersion := s.cs.ActiveForkVersionForTimestamp(timestamp)
	if version.EqualsOrIsAfter(forkVersion, version.Electra()) {
```
just find timestampFork to be confusing. `forkVersion` is implied to be based on timestamp

state-transition/core/state/statedb.go
rezbera: @abi87 a potential PR to unlock this capability https://github.com/berachain/beacon-kit/pull/2705

state-transition/core/state_processor_staking.go
```diff
@@ -67,6 +68,19 @@ func (sp *StateProcessor) processOperations(
 		}
 	}
 
+	if version.EqualsOrIsAfter(blk.GetForkVersion(), version.Electra()) {
+		// After Electra, validators can request withdrawals through execution requests which must be handled.
+		requests, err := blk.GetBody().GetExecutionRequests()
+		if err != nil {
+			return err
+		}
+		for _, withdrawal := range requests.Withdrawals {
```
abi87, line 77: Note: while we don't handle deposits or attestation requests, we may want to log that they have happened just to have observability over them

state-transition/core/state_processor_validators.go
```diff
@@ -45,28 +45,34 @@ func (sp *StateProcessor) processRegistryUpdates(st *statedb.StateDB) error {
...
+		// Note: without slashing and voluntary withdrawals, there is no way
+		// for an active validator to have its balance less or equal to EjectionBalance.
+		// Even Partial Withdrawals through EIP7002 can only reduce a validator's balance to `MinActivationBalance`,
+		// which is not enough to trigger a validator exit.
+		// A Full Withdrawal through EIP7002 would initiate a validator exit directly and does not rely on `processRegistryUpdates`.
+		// As such, we do not include the logic:
+		/*
+			elif is_active_validator(validator, current_epoch) and validator.effective_balance <= config.EJECTION_BALANCE:
+							initiate_validator_exit(state, ValidatorIndex(index))  # [Modified in Electra:EIP7251]
+		*/
```
abi87, line 70: suggestion: in such a case we may log an error, just to be on the safe side. From what I have seen this case is really not possible, so I would not be against panic here, but logging an error may be enough

state-transition/core/state_processor_withdrawals.go
```diff
@@ -136,3 +155,194 @@ func (sp *StateProcessor) processWithdrawals(
...
+			withdrawalFields(withdrawalRequest, nil)...,
+		)
+		return nil
+	}
+	index, validator, err := validateWithdrawal(st, withdrawalRequest)
+	if err != nil {
+		sp.logger.Info("Failed to validate withdrawal", withdrawalFields(withdrawalRequest, err)...)
+		// Note that we do not return error on invalid requests as it's a user error and invalid withdrawal requests are simply skipped.
+		// Since we're interacting with the KV Store, i/o errors could occur, but we expect these to be sparse enough
+		// that it won't impact consensus. We could consider returning error on codec.ErrEncoding rather than silencing here.
```
abi87, line 183: I would drop this part of the comment. In order for that to happen we need a super majority of nodes to have data corruption for the very same block. I don't think we need to account for that (and a request can always be re-issued)

state-transition/core/state_processor_withdrawals.go
abi87: It's fine, we just need to be aware of this behaviour: a validator cannot stake and unstake "immediately". It will be forced to run at least a epoch cc @camembera

abi87 (approved):
LGTM

state-transition/core/state_processor_validators.go
rezbera: good idea
e241190e3e0ec78477e42170d852096f2ea3b4ca

consensus-types/types/pending_partial_withdrawal.go
```diff
@@ -147,3 +147,14 @@ func (p *PendingPartialWithdrawals) ValidateAfterDecodingSSZ() error {
 	}
 	return nil
 }
+
+// PendingBalanceToWithdraw implements get_pending_balance_to_withdraw.
```
calbera, line 151:
```suggestion
// PendingBalanceToWithdraw implements get_pending_balance_to_withdraw from the ETH2.0 spec.
// https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#new-get_pending_balance_to_withdraw
```

state-transition/core/state_processor_exits.go
```diff
@@ -0,0 +1,77 @@
...
+func (sp *StateProcessor) InitiateValidatorExit(st *statedb.StateDB, idx math.ValidatorIndex) error {
+	validator, err := st.ValidatorByIndex(idx)
+	if err != nil {
+		return err
+	}
+	// By this point, the state should be correctly updated with the fork.
+	fork, err := st.GetFork()
+	if err != nil {
+		return err
+	}
```
calbera, line 40: this is not necessarily true. Specifically this when this is called by < processValidatorSetCap < processEpoch < ProcessSlots. ProcessFork is always called _after_ ProcessSlots

calbera: would rephrase comment to something like
```suggestion
	// We will use the fork version from the state to determine how to exit the validator.
	fork, err := st.GetFork()
	if err != nil {
		return err
	}
```

calbera: noting: In the case the timestamp is between epochs (most likely), i.e. where first part of epoch is Deneb1, second part of epoch is Electra, we will process the validator exits at the end of this epoch as per the Electra rules.

And, if the timestamp of the Electra fork happens to be crossed at the first slot of a new epoch, we will process the exits of the previous epoch according to Deneb1 (this is bc ProcessSlots is called but ProcessFork hasn't been called yet). 

Correct me if I'm wrong on understanding this.

state-transition/core/state_processor_withdrawals.go
```diff
@@ -136,3 +155,190 @@ func (sp *StateProcessor) processWithdrawals(
...
+		return err
+	}
+	// If partial withdrawal queue is full, only full exits are processed
+	if len(pendingPartialWithdrawals) == constants.PendingPartialWithdrawalsLimit && !isFullExitRequest {
+		sp.logger.Warn(
+			"skipping processing of withdrawal request as partial withdrawal queue is full",
+			withdrawalFields(withdrawalRequest, nil)...,
+		)
+		return nil
+	}
```
calbera, line 177: question: this means we skip for now, but we will get to it in the next block right? since the next block will process another 8 partial withdrawals

state-transition/core/state_processor_withdrawals.go
```diff
@@ -136,3 +155,190 @@ func (sp *StateProcessor) processWithdrawals(
...
+		// As long as `processPartialWithdrawal` is called after `processSlots`, this will always return the correct slot.
+		currentSlot, getErr := st.GetSlot()
+		if getErr != nil {
+			return getErr
+		}
+		nextEpoch := sp.cs.SlotToEpoch(currentSlot) + 1
+		// Note that we do not need to set the ExitEpoch anywhere here as Partial withdrawals do not exit the validator
+		// as we enforce that the `toWithdraw` amount is always above the `minActivationBalance`.
+		// For example, if a validator's balance is already at `minActivationBalance` (250K BERA on mainnet)
+		// and they request a partial withdrawal, the `toWithdraw` amount will always be zero.
```
calbera, line 256: question: if `toWithdraw` is 0, should we just skip adding the PendingPartialWithdrawal? i.e. why waste it on one with an amount of 0 (IIUC that would effectively be a no-op when processing it). I know this isn't part of spec so maybe theres a something going on I'm missing here

state-transition/core/state_processor_exits.go
rezbera: > this is not necessarily true. Specifically this when this is called by < processValidatorSetCap < processEpoch < ProcessSlots. ProcessFork is always called after ProcessSlots

Very good point. Fixed. You're right in that it has implications around the fork boundary and worth a thought experiment. Will think on it and see if I can write a test for the described scenario

state-transition/core/state_processor_withdrawals.go
rezbera: no it will dropped

rezbera: this is the edge case the concerns me the most with regards to UX as it silently dropping without any mistake from a withdrawer

state-transition/core/state_processor_withdrawals.go
rezbera: It's a reasonable point. My take is that validator operators could be using the 0 withdrawal as a feedback loop in their operations. The only discussion find on this is here https://ethereum-magicians.org/t/eip-7002-execution-layer-triggerable-exits/14195/23 . Beyond that, no obvious reason to me why the no-op is kept

rezbera: We could defs contribute these sort of questions for future eips to eth-magicians forum

state-transition/core/state_processor_withdrawals.go
calbera: ah yeah, its never stored in the state. We should add a metric as we do for [non-withdrawable deposits](https://github.com/berachain/beacon-kit/blob/main/state-transition/core/state_processor_staking.go#L145) and [deposit stake lost](https://github.com/berachain/beacon-kit/blob/main/state-transition/core/state_processor_staking.go#L167)

rezbera: Yea this PR doesn't have any metrics atm. Will save it for follow up PRs as we go - a dashboard for the logs would suffice for now

calbera: Should be done before cutting an rc so the metric can be tracked as it would be on mainnet

calbera: Should probably also have a warning/alerting if we are approaching close to this limit, to alert externally not to create withdrawals

rezbera: Agreed - our alerting could use a lot of improvement.

state-transition/core/state_processor_withdrawals.go
calbera: This feels like a griefing vector especially if we are near the limit

state-transition/core/state_processor_exits.go
calbera: I don’t think it is an issue though as there shouldn’t be any non determinism when replaying blocks

state-transition/core/state_processor_withdrawals.go
rezbera: https://ethereum-magicians.org/t/eip-7002-execution-layer-triggerable-exits/14195/29?u=rezbera

lets see if they say anything

state-transition/core/state_processor_exits.go
calbera: Also it’s a ~1/192 chance of the latter case happening so i wouldn’t worry too much

state-transition/core/state_processor_withdrawals.go
calbera: No earlier than 3-5 business days but I’ll gladly eat my words!

state-transition/core/state_processor_withdrawals.go
calbera: https://github.com/berachain/beacon-kit/issues/2710

calbera (approved):
utACK

state-transition/core/state_processor_withdrawals.go
rezbera: For posterity - CTL https://github.com/berachain/beacon-kit/pull/2708#discussion_r2060844599

PR #2656, opened 2025-04-09
[1/n] feat(pectra): Add Pending Partial Withdrawals to BeaconState

rezbera:
This PR does the following:
1. Introduce the `PendingPartialWithdrawal` type needed for Electra and is introduces as part of the `BeaconState`. 
2. Introduce `PendingPartialWithdrawals` as the list type for `PendingPartialWithdrawal`, same patterns as `BlobSidecars`. 
3. Add `pendingPartialWithdrawals sdkcollections.Item[*ctypes.PendingPartialWithdrawals]` to the BeaconDB KV Store. Getting and Setting on this structure is gated behind fork. 
4. Make the `BeaconState` `versionable`, such that is now fork aware. This allows us to modify the SSZ for BeaconState such that it includes `[]*PendingPartialWithdrawal` after `electra`. 
5. Small refactor of the Proof API Handlers such that it does not require generics.


consensus-types/types/pending_partial_withdrawal.go
```diff
@@ -0,0 +1,122 @@
...
+	buf := make([]byte, ssz.Size(p))
+	return buf, ssz.EncodeToBytes(buf, p)
+}
+
+// HashTreeRoot computes and returns the hash tree root for the PendingPartialWithdrawal.
+func (p *PendingPartialWithdrawal) HashTreeRoot() common.Root {
+	return ssz.HashSequential(p)
+}
+
+type PendingPartialWithdrawals []*PendingPartialWithdrawal
```
rezbera, line 95: SSZ impl mostly inspired by `type BlobSidecars []*BlobSidecar`

primitives/constants/operations_per_block.go
```diff
@@ -43,4 +43,8 @@ const (
 
 	// MaxDepositRequestsPerPayload is the maximum number of deposit requests in a execution payload.
 	MaxDepositRequestsPerPayload = 8192
+
+	// PendingPartialWithdrawalsLimit is the maximum number of pending partial withdrawals.
+	// as defined in the spec https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#state-list-lengths
+	PendingPartialWithdrawalsLimit = 134_217_728
```
rezbera, line 49: I'm unsure on the expected value of this. Spec.md says `134_217_728`, but pyspec says 
```py
PENDING_PARTIAL_WITHDRAWALS_LIMIT = uint64(64)
```
Need confirms from @shotes / @calbera

consensus-types/types/state.go
```diff
@@ -149,17 +160,6 @@ func (st *BeaconState) HashTreeRoot() common.Root {
 /*                                   FastSSZ                                  */
 /* -------------------------------------------------------------------------- */
 
-func (st *BeaconState) MarshalSSZTo(
```
rezbera, line 152: Unused except in test

consensus-types/types/state_test.go
```diff
@@ -124,75 +142,74 @@ func generateRandomBytes32(count int) []common.Bytes32 {
...
 }
 
 func TestBeaconState_UnmarshalSSZ_Error(t *testing.T) {
 	t.Parallel()
-	state := &types.BeaconState{}
-	err := state.UnmarshalSSZ([]byte{0x01, 0x02, 0x03}) // Invalid data
-	require.ErrorIs(t, err, io.ErrUnexpectedEOF)
-}
-
-func TestBeaconState_MarshalSSZTo(t *testing.T) {
```
rezbera, line 166: test removed as function removed

node-api/handlers/proof/types/interfaces.go
```diff
@@ -21,24 +21,9 @@
...
 
 import (
-	ctypes "github.com/berachain/beacon-kit/consensus-types/types"
-	"github.com/berachain/beacon-kit/primitives/math"
 	fastssz "github.com/ferranbt/fastssz"
 )
 
-// BeaconState is the interface for a beacon state.
-type BeaconState[
-	BeaconStateMarshallableT any,
```
rezbera, line 31: Removed as not needed with minor refactoring (i.e. pass in BeaconStateMarshallable instead of BeaconState to functions that are going to Marshal it anyway

primitives/constants/operations_per_block.go
calbera: image
I believe minimal is used for testing, so we should stick to 2**27

consensus-types/types/pending_partial_withdrawal.go
calbera: Just pointing out, think the receiver for all these functions should be without pointer `like (p PendingPartialWithdrawals)`, because functions are read only and unnecessary to have pointer to a slice (which is already reference). Not opinionated either way though.

primitives/constants/operations_per_block.go
rezbera: nice find. didn't realise `minimal` deviated

consensus-types/types/pending_partial_withdrawal.go
rezbera: I initially thought the same but for some obscure reason, SSZ Unmarshal doesn't like it and the tests `TestPendingPartialWithdrawals_ValidValuesSSZ` all break

rezbera: @shotes do you happen to recall issues around this?

node-api/handlers/proof/merkle/execution_number_test.go
```diff
@@ -68,24 +69,30 @@ func TestExecutionNumberProof(t *testing.T) {
...
+				bbh := types.NewBeaconBlockHeader(
+					tc.slot,
+					tc.proposerIndex,
+					tc.parentBlockRoot,
+					bs.HashTreeRoot(),
+					tc.bodyRoot,
+				)
+				bsm, err := bs.GetMarshallable()
+				require.NoError(t, err)
+				if forkVersion == version.Electra() {
```
rezbera, line 88: These tests fail on pectra. Need to figure out why before merging

consensus-types/types/pending_partial_withdrawal.go
shotes: Ah yea, I do recall some obscure issue being the reason it is the way it is. Let me investigate and get back

shotes: Let's walk through the reasoning behind why it is this way for `BlobSidecars` and then correlate that to the `PendingPartialWithdrawals`:
- `ssz.Unmarshal(sidecarBz, &sidecars)` requires a pointer to the sidecars value so that it can write the unmarshaled data to it.
- Under the hood, the ssz marshaler treats a pointer to an object different than just the object. Thus, `*BlobSidecars` has a different ssz definition than `BlobSidecars`.
- Since we only define `DefineSSZ` and `SizeSSZ` for `*BlobSidecars` and NOT for `BlobSidecars`, the ssz marshaler uses our defined definition for `*BlobSidecars` and uses its own default behavior definition for `BlobSidecars`. These are not the same.
- Recall the first point, where `ssz.Unmarshal()` requires the pointer. This means that the ssz marshaler uses the definition for `*BlobSidecars`, so we MUST implement our ssz definitions on the pointer to be able to use `ssz.Unmarshal()`.

All of these points also apply to `PendingPartialWithdrawals`.

shotes: Actually I don't know if the first point stands. Testing.

rezbera: Awesome, thanks for the investigation @shotes

node-api/handlers/proof/merkle/proposer_pubkey_test.go
```diff
@@ -69,30 +70,37 @@ func TestBlockProposerPubkeyProof(t *testing.T) {
...
 
-			proof, _, err := merkle.ProveProposerPubkeyInBlock(bbh, bs)
-			require.NoError(t, err)
-			expectedProof := ReadProofFromFile(t, tc.expectedProofFile)
-			require.Equal(t, expectedProof, proof)
-		})
+				bsm, err := bs.GetMarshallable()
+				require.NoError(t, err)
+				if forkVersion == version.Electra() {
+					t.Skip("TODO(pectra): This test currently fails as ProveProposerPubkeyInBlock will panic since the g-index changes post-electra")
```
rezbera, line 97: Will be addressed in https://github.com/berachain/beacon-kit/pull/2661/files#diff-60b848098da818e613307f6633a74467ceab385f1bfe5d2277ba649a6de8ef6a

rezbera:
Tested mainnet sync upto block ~90K

node-api/handlers/proof/merkle/proposer_pubkey_test.go
calbera, line 1: would remove all changes to this file. testdata files with real proofs for deneb (existing) and electra (new) in https://github.com/berachain/beacon-kit/pull/2661

node-api/handlers/proof/merkle/proposer_pubkey_test.go
```diff
@@ -69,30 +70,37 @@ func TestBlockProposerPubkeyProof(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		t.Run(tc.name, func(t *testing.T) {
-			vals := make(types.Validators, tc.numValidators)
-			for i := range vals {
-				vals[i] = &types.Validator{}
-			}
-			vals[tc.proposerIndex] = &types.Validator{Pubkey: tc.pubKey}
+		for _, forkVersion := range version.GetSupportedVersions() {
```
calbera, line 73: need to remove the change to iterate through supported versions. These tests are based on pre-generated proofs in testdata folder, so each testcase will have its own version to specify (cannot run all fork versions on each testcase. https://github.com/berachain/beacon-kit/pull/2661 will add the new testcases

node-api/handlers/proof/merkle/proposer_pubkey_test.go
```diff
@@ -69,30 +70,37 @@ func TestBlockProposerPubkeyProof(t *testing.T) {
...
+					vals[i] = &types.Validator{}
+				}
+				vals[tc.proposerIndex] = &types.Validator{Pubkey: tc.pubKey}
 
-			bs, err := mock.NewBeaconState(
-				tc.slot, vals, 0, common.ExecutionAddress{},
-			)
-			require.NoError(t, err)
+				bs, err := mock.NewBeaconState(
+					tc.slot, vals, 0, common.ExecutionAddress{}, forkVersion,
```
calbera, line 82: this should just take in `version.Deneb()` in this PR. https://github.com/berachain/beacon-kit/pull/2661 will add diff cases for `version.Electra()` BeaconState

node-api/handlers/proof/merkle/proposer_pubkey_test.go
calbera: can be removed from this PR

node-api/handlers/proof/merkle/proposer_pubkey_test.go
calbera: resolved in https://github.com/berachain/beacon-kit/pull/2656/commits/d8a4bbe7ea3ba682ff5e048120a197763967be7d

node-api/handlers/proof/merkle/constants.go
```diff
@@ -21,6 +21,8 @@
 package merkle
 
 const (
+	// TODO(pectra): All of these values need to be evaluated for pectra.
```
calbera, line 24: addressed in #2661

storage/beacondb/registry.go
```diff
@@ -242,3 +243,28 @@ func (kv *KVStore) GetTotalActiveBalances(
 	)
 	return totalActiveBalances, err
 }
+
+// GetPendingPartialWithdrawals is equivalent to `pending_partial_withdrawals`
+// If called before electra, will return an error.
+func (kv *KVStore) GetPendingPartialWithdrawals() ([]*ctypes.PendingPartialWithdrawal, error) {
+	if err := kv.requireEqualOrAfterVersion(version.Electra()); err != nil {
```
calbera, line 250: noting: these version checks are fine as long as we set the Fork object to Electra before state_processing any Electra blocks. May need to be careful of this when proposer building first Electra block.

calbera (approved):
utACK

consensus-types/types/pending_partial_withdrawal_test.go
```diff
@@ -0,0 +1,253 @@
...
+			// Unmarshal back into original type
+			var recomputedPending types.PendingPartialWithdrawal
+			err = ssz.Unmarshal(prysmBytes, &recomputedPending)
+			require.NoError(t, err)
+			require.Equal(t, *tc.pending, recomputedPending)
+		})
+	}
+}
+
+//nolint:paralleltest // Invalid SSZ values cannot be run in parallel due to shared zeroalloc.
```
abi87, line 106: <3

consensus-types/types/state.go
abi87: IIRC correctly this is part of the interface required by `fastssz` to encode stuff.
I am not necessarily against removing it, but I am very much in favour of consolidating the ssz libraries usage at some point.

storage/beacondb/registry.go
```diff
@@ -242,3 +242,22 @@ func (kv *KVStore) GetTotalActiveBalances(
...
+		return nil, err
+	}
+	if pendingPartialWithdrawals == nil {
+		return nil, errors.New("unexpected nil pending partial withdrawals")
+	}
+	return *pendingPartialWithdrawals, err
+}
+
+// SetPendingPartialWithdrawals sets the pending partial withdrawals
+func (kv *KVStore) SetPendingPartialWithdrawals(pendingPartialWithdrawals []*ctypes.PendingPartialWithdrawal) error {
```
abi87, line 260: just to be clear: this requires that the slice is not modified once it is passed to `SetPendingPartialWithdrawals` right?

rezbera: Yes, any modification to the slice should be will be done before calling `SetPendingPartialWithdrawals`. This makes sure that business logic does not leak to the KVStore

abi87 (approved):
looks fine to me. One point I see is that it looks pending withdrawals should not be modified once passed to SetPendingWithdrawals, otherwise what we store in the kvstore may be retroactively modified too.
I would make this explicit with a comment and a unit test. Not sure if we are open to more radical ideas like deep copies or nil-ling slice when returning from Set.

rezbera:
> looks fine to me. One point I see is that it looks pending withdrawals should not be modified once passed to SetPendingWithdrawals, otherwise what we store in the kvstore may be retroactively modified too.

This is unclear to me @abi87 . There will be a common use case of 'updating' the `PendingWithdrawals`. For example:

```go
                pendingWithdrawals, _ := st.GetPendingPartialWithdrawals()
		pendingWithdrawals = append(pendingWithdrawals, ppWithdrawal)
		st.SetPendingPartialWithdrawals(pendingWithdrawals)
```

The above example is supported in its current form

abi87:
> > looks fine to me. One point I see is that it looks pending withdrawals should not be modified once passed to SetPendingWithdrawals, otherwise what we store in the kvstore may be retroactively modified too.
> 
> This is unclear to me @abi87 . There will be a common use case of 'updating' the `PendingWithdrawals`. For example:
> 
> ```go
>                 pendingWithdrawals, _ := st.GetPendingPartialWithdrawals()
> 		pendingWithdrawals = append(pendingWithdrawals, ppWithdrawal)
> 		st.SetPendingPartialWithdrawals(pendingWithdrawals)
> ```
> 
> The above example is supported in its current form

No case, I don't see need to do it but with current implementation I believe we must not do it too

calbera:
> > > looks fine to me. One point I see is that it looks pending withdrawals should not be modified once passed to SetPendingWithdrawals, otherwise what we store in the kvstore may be retroactively modified too.
> > 
> > 
> > This is unclear to me @abi87 . There will be a common use case of 'updating' the `PendingWithdrawals`. For example:
> > ```go
> >                 pendingWithdrawals, _ := st.GetPendingPartialWithdrawals()
> > 		pendingWithdrawals = append(pendingWithdrawals, ppWithdrawal)
> > 		st.SetPendingPartialWithdrawals(pendingWithdrawals)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The above example is supported in its current form
> 
> No case, I don't see need to do it but with current implementation I believe we must not do it too

I don't think this is an issue because the sdk collections will only store this value after being encoded in bytes. 

(sorry for the many quotes)

PR #2657, opened 2025-04-09
chore(transition): Always process exactly one slot in `ProcessSlots`

shotes:
```go
// ProcessSlots deviates from the ethereum consensus specs `process_slots`. The `process_slots` function must
// iterate and process slots in which the target `slot` can be several slots ahead of the `stateSlot`. This is because
// the beacon chain can miss blocks for a given slot, resulting in not processing the slot. For example, the current
// slot is 100, and no blocks were proposed for slots 101 and 102. Upon receiving a block at slot 103, the beacon state
// must "catch up" the state and trigger slot and epoch transitions that may have happened during slot 101 and 102.
//
// Beacon-kit does not allow missed slots. Each height from cometBFT will always correspond to a beacon block slot, so
// `ProcessSlots` will always be called at every slot. Thus, we will only process the state up to the next slot.
// The reasoning behind this deviation is to be explicit in this behavior and also to better support the usage of fork
// logic in the `processEpoch` function. Since we do not fork by slot but instead fork by timestamp, we must be able to
// strictly tie each call of `processSlot` and `processEpoch` to a timestamp. Since we don't have beacon blocks during
// each iteration of the slot loop, we cannot correlate each slot to a timestamp. We instead identify that we process
// only one slot. This will allow us to simply grab the latest payload header from the state and use its fork version
// wherever we may need to add fork logic in the future.
```


state-transition/core/state_processor.go
```diff
@@ -112,28 +123,30 @@ func (sp *StateProcessor) ProcessSlots(
 	if err != nil {
 		return nil, err
 	}
+	if slot == stateSlot {
+		return res, nil
+	}
```
rezbera, line 128: When would this be the case?

shotes: During block building in `BuildBlockAndSidecars()`, we first call `ProcessSlots()` on the `blkSlot` for the next block. Afterwards when we build the block, we call `computeAndSetStateRoot()` for the state root calculation. This calls the `Transition()` function which calls `ProcessSlots()` again. But since we've already called `ProcessSlots()`, this case is hit where `slot == stateSlot`. Thus `ProcessSlots()` is effectively a no-op.

rezbera (approved):
Tested with a local mainnet sync up to block ~15K. No issues

state-transition/core/state_processor.go
```diff
@@ -112,28 +123,30 @@ func (sp *StateProcessor) ProcessSlots(
...
+	if slot != stateSlot+1 {
+		return nil, fmt.Errorf("slot %d does not match expected slot %d", slot, stateSlot+1)
+	}
```
calbera, line 131: somewhat a nit: after this enforcement that the passed in param `slot` must equal `stateSlot + 1`, would prefer using the passed in param `slot` below in this function.

calbera:
utACK, I think we should definitely add unit tests around this before merging

calbera (approved):

nidhi-singh02 (approved):

PR #2660, opened 2025-04-10
chore(node-api): Remove unused proof/ endpoints

calbera:


rezbera (approved):
Approved with @calbera's guarantee that these are unused.

PR #2678, opened 2025-04-15
[3/n] feat(pectra): Prepare beacon state for fork upgrade

calbera:
- In all of BuildBlock/ProcessProposal/OptimisticBuildBlock/FinalizeBlock, `PrepareStateForFork` is called with the following invariants:
  - it is done **before** the call to `FCU` to build a new EL payload if necessary
  - it is done **before** the call to `ProcessBlock` if necessary (i.e. for running a block's state-transition via `stateProcessor.Transition`)
  - it is done **after** the call to `ProcessSlots` if necessary. If we are transitioning to a new fork version, `ProcessSlots` must be run before the transition to ensure the previous state roots is calculated correctly [here](https://github.com/berachain/beacon-kit/blob/main/state-transition/core/state_processor.go#L170). 

- Currently, `PrepareStateForFork` only has unique logic for when upgrading to `electra` since `deneb` and `deneb1` did not require any specific state updates.

- Cleans up `node-api` usage of `ProcessSlots` by removing StateProcessor from it entirely. Now processing happens as needed per endpoint.

- Logs the upgrade to electra as such:
image

- COMPLETE ✅ -- sync a node from Genesis to past Deneb1 to ensure full backwards compatibility.


state-transition/core/state_processor.go
```diff
@@ -87,8 +87,14 @@ func (sp *StateProcessor) Transition(
 		return nil, nil
 	}
 
+	// Prepare the state for the given fork version.
+	slot := blk.GetSlot()
+	if err := sp.prepareStateForFork(st, blk.GetForkVersion(), slot); err != nil {
```
calbera, line 92: q: should this call be in ProcessSlots (as it is done in Prysm)? We do however call ProcessSlots before Transition (in block_builder) or without ever calling Transition (optimistic builder, node-api backend).

state-transition/core/state_processor_forks.go
```diff
@@ -0,0 +1,87 @@
...
+		return fmt.Errorf("unsupported fork version: %s", forkVersion)
+	}
+
+	return nil
+}
+
+// upgradeToElectra upgrades the state to the Electra fork version. It is modified from the ETH 2.0
+// spec (https://ethereum.github.io/consensus-specs/specs/electra/fork/#upgrading-the-state)
+// to only upgrade the Fork struct in the BeaconState.
+func (sp *StateProcessor) upgradeToElectra(
```
rezbera, line 79: As part of Electra, we also need to set the StateDB SetPendingPartialWithdrawals to empty list, otherwise, we'll get error `"collections: not found: key 'no_key' of type SSZMarshallable` when calling `GetPendingPartialWithdrawals`

state-transition/core/state_processor.go
rezbera: Not spec defined when it's called. It's not immediately clear why we need to call `ProcessSlots` to prepare the state in block_builder.

Worst case it needs to be done in both but the outcome will be based on whatever passes test

state-transition/core/state_processor_forks.go
```diff
@@ -0,0 +1,87 @@
...
+		// which adequately prepares the BeaconState for Deneb.
+	case version.Deneb1():
+		// Do nothing. NOTE: Deneb1 is the first hard fork of Berachain.
+		// In this fork, the Fork struct on BeaconState is NOT updated.
+		// In future hard forks, the Fork struct WILL be updated.
+	case version.Electra():
+		return sp.upgradeToElectra(st, stateFork, slot)
+	default:
+		return fmt.Errorf("unsupported fork version: %s", forkVersion)
+	}
```
rezbera, line 71: Something to consider: if we're configuring a devnet where all the hardforks are enabled from genesis, then only one of the upgrades would be run. It's okay for now though

node-api/handlers/debug/state.go
```diff
@@ -37,20 +38,49 @@ func (h *Handler) GetState(c handlers.Context) (any, error) {
...
 	if err != nil {
 		return nil, err
 	}
+
+	// Get the fork version from the state.
+	fork, err := state.GetFork()
+	if err != nil {
+		return nil, err
+	}
+	var forkVersion string
```
rezbera, line 57: This `ToString`-like impl could be a helper func in `version` package

state-transition/core/state_processor.go
calbera: It doesn't need to be done in both. Summarizing why in all the cases:
- block_builder: Transition is called after the first ProcessSlots so state will be prepared before the actual state transition.
  - Verdict: This is fine ✅ 
- ProcessSlots in optimistic builder: the state is used to retrieve `ExpectedWithdrawals` and `GetRandaoMixAtIndex`.
  - Verdict: This may be an issue 🙅, specifically if the optimistic build is for a new fork version and either `ExpectedWithdrawals` or `GetRandaoMixAtIndex` requires a value set for the new fork.

calbera: TLDR of all of the above now:
- In both block builder (PrepareProposal) and optimistic builder: we now run prepareStateForFork just before the FCU call with payload args. This ensures that if building a block that crosses into a new fork version, any payload args are correctly determined from the now-prepared beacon state. ✅ 
- `ProcessSlots` is fully removed from node-api and handled as needed from within the node-api logic. ✅

storage/beacondb/registry.go
```diff
@@ -209,40 +172,6 @@ func (kv *KVStore) GetBalances() ([]uint64, error) {
 	return balances, err
 }
 
-// GetTotalActiveBalances returns the total active balances of all validatorkv.
-// TODO: unhood this and probably store this as just a value changed on writekv.
-// TODO: this shouldn't live in KVStore.
-func (kv *KVStore) GetTotalActiveBalances(
```
calbera, line 215: unused

storage/beacondb/registry.go
```diff
@@ -135,42 +134,6 @@ func (kv *KVStore) GetTotalValidators() (uint64, error) {
 	return uint64(len(validators)), nil
 }
 
-// GetValidatorsByEffectiveBalance retrieves all validators sorted by
-// effective balance from the beacon state.
-func (kv *KVStore) GetValidatorsByEffectiveBalance() (
```
calbera, line 140: unused

state-transition/core/state/statedb.go
```diff
@@ -72,31 +71,6 @@ func (s *StateDB) DecreaseBalance(idx math.ValidatorIndex, delta math.Gwei) erro
 	return s.SetBalance(idx, balance-min(balance, delta))
 }
 
-// UpdateSlashingAtIndex sets the slashing amount in the store.
-func (s *StateDB) UpdateSlashingAtIndex(index uint64, amount math.Gwei) error {
```
calbera, line 76: unused

consensus/cometbft/mod.go
calbera, line 1: empty file -- unused

node-api/backend/state.go
```diff
@@ -21,47 +21,51 @@
...
 }
 
 // GetStateFork returns the fork of the state at the given stateID.
 func (b *Backend) StateForkAtSlot(slot math.Slot) (*ctypes.Fork, error) {
-	var fork *ctypes.Fork
-	st, _, err := b.stateFromSlot(slot)
+	st, _, err := b.StateAtSlot(slot)
 	if err != nil {
-		return fork, err
+		return nil, err
```
abi87, line 68: not even sure we need these helpers?
We may retrieve the state and let client pick whatever it needs

state-transition/core/state_processor_forks.go
```diff
@@ -0,0 +1,190 @@
...
+// PrepareStateForFork prepares the state for the fork version at the given timestamp.
+//   - If this function is called for the same version as the state's current version,
+//     it will do nothing. Unless it is the genesis slot, in which case we want to
+//     prepare the state for the genesis fork version.
+//   - If this function is called for a version before the state's current version,
+//     it will return error as this is not allowed.
+//   - If this function is called for a version after the state's current version,
+//     it will upgrade the state to the new version.
+func (sp *StateProcessor) PrepareStateForFork(
+	st *statedb.StateDB, timestamp math.U64, slot math.Slot, logUpgrade bool,
```
abi87, line 43: IIUC, slot here is always the `st.GetSlot`. Even in case where we are building a block, state would have already be updated to the right slot via ProcessSlot.
Consider dropping slot from the parameters and load it from `st`. This is what we do across all state processing functions

state-transition/core/state_processor_forks.go
```diff
@@ -0,0 +1,190 @@
...
+		return err
+	}
+
+	// Return early if the given fork version is before or equal to the current state fork version.
+	forkVersion := sp.cs.ActiveForkVersionForTimestamp(timestamp)
+	if version.IsBefore(forkVersion, stateFork.CurrentVersion) {
+		return fmt.Errorf(
+			"cannot downgrade state from %s to %s", stateFork.CurrentVersion, forkVersion,
+		)
+	} else if slot > 0 && version.Equals(forkVersion, stateFork.CurrentVersion) {
```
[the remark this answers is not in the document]
calbera: We want to allow for any  fork version being genesis.

This check is required because at genesis if this ProcessFork is called, the `forkVersion` will be equal to the `stateFork.CurrentVersion`. So we need to also enforce that slot > 0 so we can skip returning early and continue below.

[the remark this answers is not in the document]
calbera: I think the idea in [4/n](https://github.com/berachain/beacon-kit/pull/2686) is to call ProcessFork from within InitializePreminedStateFromEth1 (which is how the blockchain service initializes the state for genesis).

state-transition/core/state_processor_forks.go
```diff
@@ -0,0 +1,190 @@
...
+		// Bepolia testnet. At genesis, InitializePreminedBeaconStateFromEth1 should be called,
+		// which adequately prepares the BeaconState for Deneb.
+		if slot != constants.GenesisSlot {
+			return fmt.Errorf("the Deneb fork must be at genesis slot but got %d", slot)
+		}
+
+		// Log the upgrade to Deneb if requested.
+		if logUpgrade {
+			sp.logDenebFork(timestamp)
+		}
```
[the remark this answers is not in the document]
calbera: If we are calling this on Deneb, it must be genesis. 

Since our InitializePreminedStateFromEth1 will still call ProcessFork to "warm up" the state for whichever fork is being used for genesis. I think logging when we enter a "new" fork version is still a good idea.

state-transition/core/state_processor_forks.go
```diff
@@ -0,0 +1,190 @@
...
+) error {
+	// Set the fork on BeaconState.
+	fork.PreviousVersion = fork.CurrentVersion
+	fork.CurrentVersion = version.Electra()
+	fork.Epoch = sp.cs.SlotToEpoch(slot)
+	if err := st.SetFork(fork); err != nil {
+		return err
+	}
+
+	// Initialize the pending partial withdrawals to an empty array.
```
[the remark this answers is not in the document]
calbera: yes. As in [spec](https://ethereum.github.io/consensus-specs/specs/electra/fork/#upgrading-the-state), we are "initializing" it here so that it can now be used. We want to enforce that using this value before this fork should error (and luckily the statedb/kvstore will error)

state-transition/core/state_processor.go
```diff
@@ -87,12 +87,22 @@ func (sp *StateProcessor) Transition(
...
+	validatorUpdates, err := sp.ProcessSlots(st, slot)
 	if err != nil {
 		return nil, err
 	}
 
+	// Prepare the state for the next block's fork version, logging only once during FinalizeBlock.
+	//
+	// TODO: allow the state transition context to directly indicate what stage of block processing
+	// we are in, i.e. ProcessProposal, FinalizeBlock, etc.
+	inFinalizeBlock := ctx.VerifyPayload() && !ctx.VerifyRandao()
```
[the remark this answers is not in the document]
calbera: I think this TODO makes sense to be in the code as it is still very much tech debt to only call this once. Which is currently the proposed behavior (of logging only once) enforced and allowed by having this check here. 

When we do the cleanup later on, we can verify our change if we maintain the existing (of this PR) behavior.

node-core/services/version/version.go
```diff
@@ -140,6 +140,8 @@ func (rs *ReportingService) Stop() error {
 	return nil
 }
 
+// v1.1.5-98b93c6-76-gbc067a11d
```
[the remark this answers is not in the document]
calbera: removed

node-api/backend/validator_test.go
```diff
@@ -65,16 +59,6 @@ func TestFilteredValidators(t *testing.T) {
 	require.NoError(t, err)
 	sb := storage.NewBackend(cs, nil, kvStore, depositStore, nil)
 
-	sp := core.NewStateProcessor(
```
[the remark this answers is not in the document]
calbera: It was used to ProcessSlots before (which just calculates the state root and block root of block N which doesnt get committed to state for block N). Now we do it manually.

node-api/backend/state.go
```diff
@@ -21,47 +21,51 @@
...
-		return common.Root{}, err
+		return nil, slot, fmt.Errorf("CreateQueryContext failed: %w", err)
 	}
+	st := b.sb.StateFromContext(queryCtx)
 
-	// As calculated by the beacon chain. Ideally, this logic
-	// should be abstracted by the beacon chain.
-	return st.StateRootAtIndex(resolvedSlot.Unwrap() % b.cs.SlotsPerHistoricalRoot())
+	// If using height 0 for the query context, make sure to return the latest slot.
+	if slot == 0 {
```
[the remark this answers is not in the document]
calbera: Discussed this with @nidhi-singh02 and this will be handled in her work (https://github.com/berachain/beacon-kit/pull/2648, among others) to support genesis. Specifically it is required to treat the genesis case separately and in fact if querying for genesis, this function (query context) should not be used at all.

node-api/backend/state.go
calbera: It either lives in the backend or the specific API handlers. I dont mind it living in the backend. This separates concerns as the backend manages retrieving and "computing" any necessary state. And the handlers simply route that.

state-transition/core/state_processor_forks.go
```diff
@@ -0,0 +1,190 @@
...
+
+// PrepareStateForFork prepares the state for the fork version at the given timestamp.
+//   - If this function is called for the same version as the state's current version,
+//     it will do nothing. Unless it is the genesis slot, in which case we want to
+//     prepare the state for the genesis fork version.
+//   - If this function is called for a version before the state's current version,
+//     it will return error as this is not allowed.
+//   - If this function is called for a version after the state's current version,
+//     it will upgrade the state to the new version.
+func (sp *StateProcessor) PrepareStateForFork(
```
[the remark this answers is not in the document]
calbera: Actually cannot be called in ProcessSlots for the reason that here, we do not know the timestamp (which is a necessary param for ProcessFork) of the block being built yet. 

Lower in this code block, if a payload has not already been built by this node, then when building a new payload for this slot we will in fact call ProcessFork here, but this time with our best estimate of this new block's timestamp. 

Aside: this is how all blocks are being built (even the optimistic ones) -- with a "best estimate" of the new block's timestamp. This means that when we enter a new fork version, that first block of the fork may have issues getting built and verified (but also may not). Pointed out when fork by timestamp was introduced.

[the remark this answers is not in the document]
calbera: renamed to ProcessFork in https://github.com/berachain/beacon-kit/pull/2678/commits/6408e427da6c39cd3f83aa345b62a03ed030ad78

state-transition/core/state_processor_forks.go
```diff
@@ -0,0 +1,190 @@
...
+
+		// Log the upgrade to Deneb1 if requested.
+		if logUpgrade {
+			sp.logDeneb1Fork(stateFork.CurrentVersion, timestamp, slot)
+		}
+	case version.Electra():
+		return sp.upgradeToElectra(st, stateFork, timestamp, slot, logUpgrade)
+	default:
+		return fmt.Errorf("unsupported fork version: %s", forkVersion)
+	}
```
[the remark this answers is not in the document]
calbera: https://github.com/berachain/beacon-kit/pull/2678/commits/6408e427da6c39cd3f83aa345b62a03ed030ad78

state-transition/core/state_processor_forks.go
```diff
@@ -0,0 +1,190 @@
...
+		// Do nothing to the state. NOTE: Deneb1 is the first hard fork of  Berachain mainnet and
+		// Bepolia testnet. In this fork, the Fork struct on BeaconState is NOT updated. In
+		// future hard forks, the Fork struct should be updated.
+
+		// Log the upgrade to Deneb1 if requested.
+		if logUpgrade {
+			sp.logDeneb1Fork(stateFork.CurrentVersion, timestamp, slot)
+		}
+	case version.Electra():
+		return sp.upgradeToElectra(st, stateFork, timestamp, slot, logUpgrade)
```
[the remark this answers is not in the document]
calbera: https://github.com/berachain/beacon-kit/pull/2678/commits/6408e427da6c39cd3f83aa345b62a03ed030ad78

state-transition/core/state_processor_forks.go
calbera: https://github.com/berachain/beacon-kit/pull/2678/commits/6408e427da6c39cd3f83aa345b62a03ed030ad78

state-transition/core/state_processor_forks.go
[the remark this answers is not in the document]
calbera: Am going to move the slot > 0 check + any explicit handling of the genesis case in ProcessFork to [4/n](https://github.com/berachain/beacon-kit/pull/2686), where this is being solved.

state-transition/core/state_processor_forks.go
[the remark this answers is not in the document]
calbera: Regardless of going to genesis with Deneb, Electra, etc I think that fork information should still be logged

node-api/backend/state.go
calbera: https://github.com/berachain/beacon-kit/pull/2678/commits/21d9b307d4f9283fb739b8761fa02d5f203673ca

beacon/blockchain/interfaces.go
```diff
@@ -86,6 +86,10 @@ type StateProcessor interface {
 		*ctypes.ExecutionPayloadHeader,
 		common.Version,
 	) (transition.ValidatorUpdates, error)
+	// ProcessFork prepares the state for the fork version at the given timestamp.
+	ProcessFork(
+		st *statedb.StateDB, timestamp math.U64, logUpgrade bool,
```
rezbera, line 91: this `logUpgrade` feels odd. Usually log filtering should be done by the logger through log levels

rezbera: Not a blocker tho

calbera: This is whether or not we want to log at all. It will always be `INFO` level. If we remove this we will see the new fork version information being logged 3-4 times

rezbera: Ahkk

beacon/validator/block_builder.go
```diff
@@ -248,15 +240,25 @@ func (s *Service) retrieveExecutionPayload(
 		return nil, err
 	}
 
+	// We must prepare the state for the fork version of the new block being built to handle
+	// the case where the new block is on a new fork version. Although we do not have the
+	// confirmed timestamp by the EL, we will assume it to be `nextPayloadTimestamp` to decide
```
rezbera, line 245: `Although we do not have the
	// confirmed timestamp by the EL`

As discussed, the EL must use the timestamp provided in FCU Attributed

See here where the timestamp is applied to header
https://github.com/ethereum/go-ethereum/blob/71f9f1e684affbd0ab87f21202d247786c9ce371/miner/worker.go#L179-L186

calbera: Correct. The comment is still true that we are not confirmed bc we are not 100% confirmed that the payload will build validly until we get it back.

rezbera (approved):
LGTM  - not 100% on the API specific changes but assuming that issues would have been caught in E2E

state-transition/core/state_processor_forks.go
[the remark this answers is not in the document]
calbera: Just thinking on this more. We can't 100% know the block time (EPH time) for [this call to ProcessSlots](https://github.com/berachain/beacon-kit/blob/fork-struct-upgrade/beacon/validator/block_builder.go#L69). See:

If block `N` was finalized with
- cons time: `a`
- EPH time: `b`

Block `N+1` is optimistically built with an EPH time of `c`, where `c = max(a + 1, b + 1)`

So when PrepareProposal is called for block `N+1` with a cons time of `t`, our best prediction of block `N+1`'s EPH time would be `nextPayloadTimestamp` ([which is calculated here](https://github.com/berachain/beacon-kit/blob/fork-struct-upgrade/beacon/validator/block_builder.go#L247-L251)): `d = max(t, b + 1)`. If we are not using optimistic build we know the timestamp of `N + 1` is `d` with 100% confidence. However, if using the optimistic build for `N + 1`, these cases are possible:
  - `c = b + 1` if `b + 1 > a + 1`. Then `d = c = b + 1` if and only if `t <= b + 1`.
  - `c = a + 1` if `a + 1 > b + 1`. Then `d = c = a + 1` if and only if `t >= b + 1` AND `t = a + 1`.
  
So therefore because of optimistic build we can't know the timestamp when we call `ProcessSlots`. QED

beacon/validator/block_builder.go
```diff
@@ -214,15 +214,11 @@ func (s *Service) retrieveExecutionPayload(
...
 	//
 	// Get the payload for the block.
-	envelope, err := s.localPayloadBuilder.RetrievePayload(
-		ctx,
-		slotData.GetSlot(),
-		parentBlockRoot,
-	)
+	slot := slotData.GetSlot()
+	envelope, err := s.localPayloadBuilder.RetrievePayload(ctx, slot, parentBlockRoot)
 	if err == nil {
```
abi87, line 222: @calbera question: shouldn't we carry out `ProcessFork` as well, using `envelope.GetExecutionPayload().GetTimestamp()`?
`ProcessFork` does modify the state, but with the code as is, we are doing it some of the times but not others.

beacon/validator/block_builder.go
abi87: I don't really understand the comment `Although ...`.
Here we are building a new block, so it is understandable that its timestamp is not finalised. IIUC the purpose of this snipped is to make the fork coherent with this timestamp (and as Rez pointed EL is bound to use that timestamp).

state-transition/core/state_processor_forks.go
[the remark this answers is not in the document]
abi87: I am not convinced by this.
I am the one who introduced the idea of "estimating" the timestamp with optimistic block building but I find this confusing now.
When we build a block we are free to pick whatever timestamp we want, as long as it satisfies the `payloadtime.Verify` check. There is really no estimated nor real time, anything passing verification is fine.
If you accept this, then we do know the timestamp to be fed to `ProcessFork` when we call `ProcessSlots`; only the code structure as is it's making it way harder than needed to see that. Specifically on block building:
- either we have optimistic build off, in which case `nextPayloadTimestamp` is really calculated deterministically from `slotData` and from a state quantity, `LatestExecutionPayloadHeader`, which is not modified by `ProcessSlots`
- or we have optimistic build on, in which case timestamp has been already decided and we just need to retrieve it via `getPayload` and we must use that for `ProcessFork` (which we are not doing it currently and it's a bug I think)

I am going to give it a try and refactor this in a separate PR, which I cannot anticipate rn if it's going to be easy or not

abi87 (commented):
LGTM other than a missing `ProcessFork`

beacon/validator/block_builder.go
calbera: True but in this specific location, we are not 100% confirmed the EL will indeed build a valid block. Only once we know that can we also know the exact beacon block timestamp

beacon/validator/block_builder.go
calbera: Not necessary in this case because 
- here the optimistically built envelope is returned
- back into control flow in BuildBlockAndSidecars
- then comuteAndSetStateRoot is called which then invokes state processor Transition function, which then calls in the following order:
  - ProcessSlots (no-op in this case since ProcessSlots was called before even retrieving the envelope)
  - ProcessFork
  - ProcessBlock

abi87: But then why is `ProcessFork` needed at all here? It does not seem to me that `s.localPayloadBuilder.RequestPayloadSync` need the state to be prepared with `ProcessFork`

calbera: In the case there is no optimistically built payload, we will not return the envelope here. Continuing in this function, before we send an FCU to EL to build this new payload we should process the fork in case this new payload is entering a new fork version. Before the FCU call, there are a couple reads from the beacon state which may be fork version dependent.

abi87: You're right

abi87 (approved):
Code seems correct to me. I am trying to consolidate `ProcessSlots` and `ProcessFork` but it's a larger endeavour, so pushing it to a different PR

PR #2684, opened 2025-04-15
Test electra timestamp

fridrik01:


fridrik01:
Replaced with https://github.com/berachain/beacon-kit/pull/2685

PR #2685, opened 2025-04-15
Make sure deneb is the genesis fork version regardless of timestamp

fridrik01:
This PR makes sure we set Deneb as the genesis version in `forceSyncUponProcess`. 

It seems that after landing https://github.com/berachain/beacon-kit/pull/2644 which sets the `ElectraForkTime` to 0 then we started having issues producing blocks after genesis, resulting in increasing rounds.

After debugging with @abi87 then we tracked it down to `ActiveForkVersionForTimestamp` being passed a old timestamp from the `forceSyncUponProcess` which matches the one in the [eth-genesis](https://github.com/berachain/infra/blob/devnet-core/helm-charts/shared-resources/templates/bera-eth-genesis.yaml) (1699397542) causing it to return Deneb version (which should be correct, however then it can't produce blocks).

I don't understand yet why it works in kurtosis devnet as it sets the devnet spec timestamp to 0 and the [eth-genesis.json](https://github.com/berachain/beacon-kit/blob/92944b63f906d22957c4c9388ba2a90eb25ee303/testing/files/eth-genesis.json#L44) is also 0. It means that ActiveForkVersionForTimestamp should return Electra which is the same we have in devnet.

### DIFF
--- a/beacon/blockchain/payload.go
+++ b/beacon/blockchain/payload.go
@@ -29,6 +29,7 @@ import (
 	engineerrors "github.com/berachain/beacon-kit/engine-primitives/errors"
 	"github.com/berachain/beacon-kit/errors"
 	"github.com/berachain/beacon-kit/primitives/math"
+	"github.com/berachain/beacon-kit/primitives/version"
 	statedb "github.com/berachain/beacon-kit/state-transition/core/state"
 )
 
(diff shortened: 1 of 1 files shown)


beacon/blockchain/payload.go
```diff
@@ -61,7 +68,7 @@ func (s *Service) forceSyncUponProcess(
 			SafeBlockHash:      lph.GetParentHash(),
 			FinalizedBlockHash: lph.GetParentHash(),
 		},
-		s.chainSpec.ActiveForkVersionForTimestamp(lph.GetTimestamp()),
+		forkVersion,
```
calbera, line 71: just noting that this change likely does not matter as this fork version is only used to validate that we are **not before Deneb** [here](https://github.com/berachain/beacon-kit/blob/main/execution/client/ethclient/engine.go#L121-L124). So if this is Deneb, Deneb1, Electra, it will not error on our side.

calbera: Also will not affect the FCU call to the EL either

abi87: Totally agree. Should we then drop the fork check?
It is very weirdly placed to me, considering EL is not aware of CL forks.
Also we may just move it to relevant `ValidateAfterDecodingSSZ` or to other constructors.
cc @rezbera, @fridrik01

fridrik01:
Closed as the root problem that I needed to do a regenesis when changing the timestamp, also https://github.com/berachain/beacon-kit/pull/2686 should improve the genesis version handling

PR #2686, opened 2025-04-15
[4/n] feat(genesis): Remove version.Genesis and allow for different fork version in Genesis

rezbera:
One of the issues that has arisen as part of introducing pectra support is the realisation that our system is not well-designed to have multiple forks enabled in genesis.

To account for this, we have to stop treating `Deneb()` as 'special' Genesis version, and rather calculate the actual genesis version based on the genesis timestamp and the fork timestamps. 

The simplest path forward is to add the genesis timestamp to the chain spec. There is an operation risk introduced, where we must ensure that the chain spec genesis timestamps aligns with the actual timestamps in the CL and EL genesis files. 

For mainnet and testnet, this won't be a problem given that they are static values, however, in devnet, if we ever change the genesis timestamp, we will need to ensure the update is correctly reflected in the devnet chain spec. I don't forsee this being a currently, the devnet genesis timestamp is set to zero, and is should not be frequently updated.


config/spec/devnet.go
```diff
@@ -38,6 +38,9 @@ const (
 	// of Gwei) that can be staked.
 	devnetMaxStakeAmount = 4000 * params.GWei
 
+	// devnetGenesisTime is the timestamp of devnet genesis.
+	devnetGenesisTime = 0
```
rezbera, line 42: The simplest approach has been to add the genesis time to the chain spec as the chainspec is widely available, whereas the genesis file is not always available

cli/commands/server/types/mainnet_spec.toml
```diff
@@ -34,6 +34,7 @@ eth1-follow-distance = 1
 target-seconds-per-eth1-block = 2
 
 # Fork-related values
+genesis-time = 1737381600
 deneb-one-fork-epoch = 2855
 electra-fork-epoch = 9999999999999999
```
calbera, line 39: these values should be changed

rezbera:
`make start` fails due to beacond init. Need a similar solution to simulated tests custom `DefaultGenesis`.

config/spec/testnet.go
```diff
@@ -29,6 +29,9 @@ func TestnetChainSpecData() *chain.SpecData {
 	// Testnet uses chain ID of 80069.
 	specData.DepositEth1ChainID = TestnetEth1ChainID
 
+	// Timestamp of the genesis block of Bepolia testnet.
+	specData.GenesisTime = 1739976735
```
calbera, line 33: This timestamp corresponds to GMT:  "Wed Feb 19 2025 14:52:15 GMT+0000". THis is consistent with eth-genesis.json. 

Just pointing out that beacon genesis.json has genesis time as "2025-02-20T20:18:20.017462Z"

rezbera: Did we stuff up by having diff timestamps in cl gen and el gen?

state-transition/core/state_processor_genesis.go
```diff
@@ -49,6 +58,9 @@ func (sp *StateProcessor) InitializePreminedBeaconStateFromEth1(
 	if err := st.SetFork(fork); err != nil {
 		return nil, err
 	}
+	if err := sp.PrepareStateForFork(st, execPayloadHeader.GetTimestamp(), 0); err != nil {
```
rezbera, line 61: Note that we call `PrepareStateForFork`. `PrepareStateForFork` has a special case for when the slot is `0`.

state-transition/core/state_processor_genesis.go
```diff
@@ -41,6 +41,15 @@ func (sp *StateProcessor) InitializePreminedBeaconStateFromEth1(
 	execPayloadHeader *ctypes.ExecutionPayloadHeader,
 	genesisVersion common.Version,
 ) (transition.ValidatorUpdates, error) {
+	if !version.Equals(genesisVersion, execPayloadHeader.GetForkVersion()) {
```
calbera, line 44: this check can be removed once GetForkVersion removed from ExecPayloadheader

state-transition/core/state_processor_genesis.go
```diff
@@ -41,6 +41,15 @@ func (sp *StateProcessor) InitializePreminedBeaconStateFromEth1(
...
+		return nil, fmt.Errorf("fork mismatch between genesisVersion %s and execPayloadHeader %s",
+			genesisVersion, execPayloadHeader.GetForkVersion())
+	}
+	if !version.Equals(sp.cs.GenesisForkVersion(), genesisVersion) {
+		return nil, fmt.Errorf("fork mismatch between chain spec genesis version %s and genesisVersion %s",
```
calbera, line 49:
```suggestion
		return nil, fmt.Errorf("fork mismatch between chain spec genesis version %s and CL genesis file fork version %s",
```

state-transition/core/state_processor_genesis.go
calbera: https://github.com/berachain/beacon-kit/pull/2686/commits/ef596d8bb815244b3806025d004b768d1f72ace0

storage/beacondb/eth1.go
```diff
@@ -22,34 +22,17 @@ package beacondb
...
+// SetLatestExecutionPayloadHeader sets the latest execution payload header.
 func (kv *KVStore) SetLatestExecutionPayloadHeader(
 	payloadHeader *ctypes.ExecutionPayloadHeader,
 ) error {
-	version := payloadHeader.GetForkVersion()
-	if err := kv.latestExecutionPayloadVersion.Set(
-		kv.ctx, version.ToUint32(),
-	); err != nil {
-		return err
-	}
```
rezbera, line 51: Concern: AppHashs

calbera: https://github.com/berachain/beacon-kit/pull/2686/commits/03278e1fd0106ae7ba320aa659d12e817189d7c7

config/spec/testnet.go
calbera: I dont believe so because the cl "genesis_time" field is kinda useless

calbera:
Noting that https://github.com/berachain/beacon-kit/issues/2692 is still not resolved

state-transition/core/state_processor_genesis.go
```diff
@@ -49,6 +56,9 @@ func (sp *StateProcessor) InitializePreminedBeaconStateFromEth1(
 	if err := st.SetFork(fork); err != nil {
 		return nil, err
 	}
+	if err := sp.PrepareStateForFork(st, execPayloadHeader.GetTimestamp(), 0, true); err != nil {
```
calbera, line 59: undecided if `logUpgrade` should be true or false here

calbera:
Noting an identified prerequisite to merging this PR: 
- Regenesis on a devnet (with electra at genesis) correctly configured 
- Resyncing a node from. genesis on the same devnet

state-transition/core/state_processor_genesis.go
```diff
@@ -83,30 +81,15 @@ func (sp *StateProcessor) InitializePreminedBeaconStateFromEth1(
...
-			return nil, err
-		}
-	}
-
-	// Before processing deposits, set the eth1 deposit index to 0.
-	if err := st.SetEth1DepositIndex(constants.FirstDepositIndex); err != nil {
+	if err := sp.seedRandaoMix(
+		st,
+		execPayloadHeader.GetBlockHash(),
+	); err != nil {
```
rezbera, line 87: minor refactors of logic to avoid so I can remove `//nolint:gocognit,funlen // todo fix.`

state-transition/core/state_processor_genesis.go
```diff
@@ -151,6 +129,57 @@ func (sp *StateProcessor) InitializePreminedBeaconStateFromEth1(
...
+		return err
+	}
+	for _, dep := range deposits {
+		if err := sp.processDeposit(st, dep); err != nil {
+			return err
+		}
+	}
+	return sp.processGenesisActivation(st)
+}
```
rezbera, line 181: These 3 new funcs are no logic changes, just refactors so the main func can remove `//nolint:gocognit,funlen // todo fix.`

rezbera:
> Noting that #2692 is still not resolved

Should be resolved now from my testing - someone else confirm on their machine.

calbera (approved):

beacon/blockchain/execution_engine.go
```diff
@@ -43,16 +43,14 @@ func (s *Service) sendPostBlockFCU(
...
-	// EL of the new head.
-	// TODO: Switch to New().
+	// Send a forkchoice update without payload attributes to notify EL of the new head.
 	req := ctypes.BuildForkchoiceUpdateRequestNoAttrs(
 		&engineprimitives.ForkchoiceStateV1{
 			HeadBlockHash:      lph.GetBlockHash(),
 			SafeBlockHash:      lph.GetParentHash(),
 			FinalizedBlockHash: lph.GetParentHash(),
 		},
-		lph.GetForkVersion(),
```
abi87, line 55: @rezbera why is this necessary (if it is)?
I know we marked `GetForkVersion` for `ExecutionPayloadHeader` as unnecessary, but it seems to me it still is necessary (e.g. in kvStore for `Set/GetLatestExecutionPayloadHeader`.
As such can't we keep it?

cli/commands/genesis/payload.go
```diff
@@ -106,17 +106,20 @@ func AddExecutionPayload(chainSpec ChainSpec, elGenesisPath string, config *cmtc
 	); err != nil {
 		return errors.Wrap(err, "failed to unmarshal beacon state")
 	}
-
 	// Inject the execution payload.
-	genesisInfo.ExecutionPayloadHeader, err =
+	eph, err :=
```
abi87, line 110: super nit: there is an extra new line here

beacon/blockchain/execution_engine.go
rezbera: @calbera wanted to remove this so I'll leave this with him

cli/commands/genesis/payload.go
```diff
@@ -106,17 +106,20 @@ func AddExecutionPayload(chainSpec ChainSpec, elGenesisPath string, config *cmtc
...
 		executableDataToExecutionPayloadHeader(
-			version.Genesis(),
+			chainSpec.GenesisForkVersion(),
 			payload,
 			chainSpec.MaxWithdrawalsPerPayload(),
 		)
 	if err != nil {
 		return errors.Wrap(err, "failed to unmarshal beacon state")
 	}
+	if eph == nil {
```
abi87, line 119: nit: error above seems off, can you fix it?

cli/commands/genesis/payload.go
```diff
@@ -140,9 +143,9 @@ func executableDataToExecutionPayloadHeader(
 	// todo: re-enable when codec supports.
 	_ uint64,
 ) (*types.ExecutionPayloadHeader, error) {
-	var executionPayloadHeader *types.ExecutionPayloadHeader
+	eph := &types.ExecutionPayloadHeader{}
 	switch forkVersion {
-	case version.Genesis():
+	case version.Deneb(), version.Deneb1(), version.Electra():
```
abi87, line 148: nit: I am a fan of switches but here we may consider just using an if. We could err immediately if version is not in the range of allowed ones and reduce indentation after it

cli/commands/genesis/payload.go
rezbera: c8d30feb6fe7f641c7536202217fe1876159960a

cli/commands/genesis/payload.go
rezbera: cb3c0a8e5de7f955fb188a748845e3534281e1f4

consensus-types/types/payload_header.go
```diff
@@ -42,6 +42,12 @@ var (
 
 // ExecutionPayloadHeader represents the payload header of an execution block.
 type ExecutionPayloadHeader struct {
+	// NOTE: This version is not required but left in for backwards compatibility.
```
abi87, line 45: I am not sure I agree this is not required. It seems to be we need it in the KvStore where we don't have access to chain specs

rezbera: Will also leave this to @calbera

state-transition/core/state_processor_payload_test.go
```diff
@@ -53,6 +52,8 @@ func TestPayloadTimestampVerification(t *testing.T) {
 
 	// process genesis before any other block
 	genesisTime := time.Now().Truncate(time.Second)
+	genesisFork := cs.ActiveForkVersionForTimestamp(math.U64(genesisTime.Unix()))
```
abi87, line 55: q: isn't GenesisTime now part of the specs?

abi87, line 55: why is not genesis time taken from the specs?

abi87 (commented):
nits

consensus-types/types/payload_header.go
rezbera: It can't be dropped in the kv store because it would cause an appHash. We would need to drop it from the kvstore as part of a hard fork

state-transition/core/state_processor_payload_test.go
rezbera: We cannot as the genesis time starts from 0.  The first test does
`payloadTime: consensusBlkTime.Add(-10 * time.Second),` which causes an underflow

beacon/blockchain/execution_engine.go
calbera: Keeping `GetForkVersion` on ExecutionPayloadHeader is **only** necessary for backwards compatibility. Hence the comment [here](https://github.com/berachain/beacon-kit/pull/2686#discussion_r2053727218).

i.e. we only need this function so that there are no apphashes (because we have a state value specifically for the fork version of the latest exec payload header in the beacon state) for a node syncing from genesis. The point being this was introduced when there was never a need for it, since the exec payload header does not change across fork versions. 

Made this change here to enforce non-usage of a function that is only left in for backwards compatibility.

consensus-types/types/payload_header.go
calbera: I would’ve loved to just drop these two fields ([here](https://github.com/berachain/beacon-kit/blob/93779da823b34e6beb5387ad01387949181493e6/storage/beacondb/kvstore.go#L61) and [here](https://github.com/berachain/beacon-kit/blob/93779da823b34e6beb5387ad01387949181493e6/storage/beacondb/kvstore.go#L64)) entirely but this change would require the kvstore to become fork version aware as well to maintain backwards compatibility. That is a refactor I’m also open to but just in this PR

calbera: Hence the comment in the [kvstore](https://github.com/berachain/beacon-kit/pull/2686/files#diff-a25cb339e5fa03deaa65a426b600a38779c4074db0cf13dc6ec390911e41801aR33) here as well, explaining why it can’t be removed simply.

state-transition/core/state_processor_staking_test.go
```diff
@@ -100,14 +99,7 @@ func TestTransitionUpdateValidators(t *testing.T) {
 	}
 
 	depRoot := append(genDeposits, blkDeposit).HashTreeRoot()
-	blk1 := buildNextBlock(
-		t,
```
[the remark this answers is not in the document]
rezbera: no preference. it was automatically put into 1 line by my idea when i added extra param 

a98e73b8e1fcd3bdad9b33a7cdc4b09397c4f740

calbera:
re: the ExecutionPayloadHeader fork version, putting up a proposal to clean up tech debt  [here](https://github.com/berachain/beacon-kit/pull/2701). cc @abi87 @rezbera

PR #2693, opened 2025-04-17
feat(node-api): node syncing api 

nidhi-singh02:


consensus/cometbft/service/abci.go
```diff
@@ -198,3 +199,8 @@ func (*Service) CheckTx(
 ) (*abci.CheckTxResponse, error) {
 	return &abci.CheckTxResponse{}, nil
 }
+
+// GetCometNode returns the concrete CometBFT node.
+func (s *Service) GetCometNode() *node.Node {
+	return s.node
```
abi87, line 205: nit: should we rename this `consensusInstance` or something like that?

node-api/handlers/node/sync.go
```diff
@@ -0,0 +1,66 @@
...
+	blockStore := node.BlockStore()
+	if blockStore == nil {
+		return nil, errNilBlockStore
+	}
+
+	latestHeight := blockStore.Height()
+	baseHeight := blockStore.Base()
+
+	response := nodetypes.SyncingData{
+		HeadSlot:     latestHeight,
```
abi87, line 52: I would comment here that this relays on the fact that there is a one to one matching among comet blocks and payloads. Should this ever change in the future, this would not be correct anymore

node-api/handlers/node/sync.go
```diff
@@ -0,0 +1,66 @@
...
+		IsOptimistic: true,
+		ELOffline:    false,
```
abi87, line 54: not sure how do we set these data by only looking at CometBFT.
In fact all of these data seems to me related to beaconKit state only.
Even the isSyncing quantity can be stored in `consensus/cometbft/service object by only looking at the SyncingToHeight attribute in FinalizeBlockRequest.

abi87 (changes requested):
It seems to me the content of this API should be served by looking at BeaconKit data only, not querying cometBFT

PR #2695, opened 2025-04-18
WIP: Consolidate process fork

abi87:


node-core/components/interfaces.go
```diff
@@ -482,7 +481,7 @@ type (
 	// ReadOnlyWithdrawals only has read access to withdrawal methods.
 	ReadOnlyWithdrawals interface {
 		EVMInflationWithdrawal(math.Slot) *engineprimitives.Withdrawal
-		ExpectedWithdrawals() (engineprimitives.Withdrawals, error)
```
abi87, line 485: interface we forgot to update

node-core/components/interfaces.go
```diff
@@ -48,7 +48,6 @@ type (
 	AttributesFactory interface {
 		BuildPayloadAttributes(
 			st *statedb.StateDB,
-			slot math.Slot,
```
abi87, line 51: slot is always picked from state (or from the block after ProcessSlots has been called, which makes state slots equal to block state).
So we can drop this parameter

payload/attributes/interfaces.go
```diff
@@ -30,3 +31,10 @@ type ChainSpec interface {
 	EpochsPerHistoricalVector() uint64
 	SlotToEpoch(slot math.Slot) math.Epoch
 }
+
+// ReadOnlyBeaconState is the interface for a read-only beacon state.
+type ReadOnlyBeaconState interface {
```
abi87, line 36: just an attempt at showing that state is only read when retrieving payloads

abi87:
This was an attempt, not good enough to be merged

PR #2701, opened 2025-04-23
chore(types): remove ExecutionPayloadHeader version

calbera:
A non-critical proposal to clean up tech debt. This PR can be merged now but chosen to be kept as a simple refactor instead of a hard fork by changing [this line](https://github.com/berachain/beacon-kit/pull/2701/files#diff-a25cb339e5fa03deaa65a426b600a38779c4074db0cf13dc6ec390911e41801aR43)


cli/commands/genesis/payload.go
```diff
@@ -69,137 +67,107 @@ func AddExecutionPayload(chainSpec ChainSpec, elGenesisPath string, config *cmtc
...
 	data *gethprimitives.ExecutableData,
-	// todo: re-enable when codec supports.
-	_ uint64,
 ) (*types.ExecutionPayloadHeader, error) {
-	eph := &types.ExecutionPayloadHeader{}
-
-	// We do not support fork versions before Deneb and after Electra.
-	if version.IsAfter(forkVersion, version.Electra()) ||
-		version.IsBefore(forkVersion, version.Deneb()) {
-		return nil, types.ErrForkVersionNotSupported
```
calbera, line 150: noting that as neither the geth `ExecutableData` nor the bkit `ExecutionPayloadHeader` objects are versioned, it doesn't make sense for the CLI command to be opinionated to versions

storage/encoding/ssz.go
```diff
@@ -62,45 +61,3 @@ func (SSZValueCodec[T]) Stringify(value T) string {
 func (SSZValueCodec[T]) ValueType() string {
 	return "SSZMarshallable"
 }
-
-// SSZVersionedValueCodec provides methods to encode and decode SSZ values for a specific version.
-type SSZVersionedValueCodec[T constraints.SSZMarshallable] struct {
```
calbera, line 67: no longer used, was only used by the ExecPayloadHeader, which didn't even need to be versioned.

storage/beacondb/eth1.go
```diff
@@ -22,38 +22,32 @@ package beacondb
...
 	*ctypes.ExecutionPayloadHeader, error,
 ) {
-	// NOTE: unmarshalling this struct is NOT affected by it's own fork version. The versioned
-	// codec is left in for backwards compatibility.
-	forkVersion, err := kv.latestExecutionPayloadVersion.Get(kv.ctx)
-	if err != nil {
-		return nil, err
-	}
-	kv.latestExecutionPayloadCodec.SetActiveForkVersion(bytes.FromUint32(forkVersion))
+	// NOTE: unmarshalling this struct is NOT affected by it's own fork version.
```
calbera, line 34:
```suggestion
```

calbera:
Confirming backwards compat with mainnet:
- sync started at height ~860,000
- currently at ~880,000 (so far so good)
- will update in a few hours...

storage/beacondb/eth1.go
```diff
@@ -22,38 +22,32 @@ package beacondb
...
 ) error {
-	// NOTE: marshalling this struct is NOT affected by it's own fork version. The versioned
-	// codec is left in for backwards compatibility.
-	version := payloadHeader.GetForkVersion()
-	if err := kv.latestExecutionPayloadVersion.Set(
-		kv.ctx, version.ToUint32(),
-	); err != nil {
-		return err
+	// NOTE: marshalling this struct is NOT affected by it's own fork version.
+	if version.IsBefore(forkVersion, version.Electra()) {
```
calbera, line 43: Can delay the hard fork by making this `version.Electra1()` or something else in the future

calbera: This refactored code is still backwards compatible

rezbera:
Agree with this change for a later fork. Will review properly after all bectra related work is done

storage/beacondb/eth1.go
abi87: super nit: consider inverting test to simplfiy code. Something like

```golang
if !version.IsBefore(forkVersion, version.Electra()) {
  // post Electra we simplified marshalling and we do not need to store version
  return nil
}
...
```

storage/beacondb/eth1.go
```diff
@@ -22,38 +22,32 @@ package beacondb
...
-	}
-	kv.latestExecutionPayloadCodec.SetActiveForkVersion(bytes.FromUint32(forkVersion))
 	return kv.latestExecutionPayloadHeader.Get(kv.ctx)
 }
 
 // SetLatestExecutionPayloadHeader sets the latest execution payload header in
 // the BeaconStore.
 func (kv *KVStore) SetLatestExecutionPayloadHeader(
-	payloadHeader *ctypes.ExecutionPayloadHeader,
+	payloadHeader *ctypes.ExecutionPayloadHeader, forkVersion common.Version,
```
abi87, line 40: spitballing here: an alternative could be to have a `SetLatestExecutionPayloadHeaderPreElectra` and a `SetLatestExecutionPayloadHeader`.
From Pectra on we would call only the latter, but clients of this method would have to decide which method to call.
Probably does not make much of a different, just pointing it out

abi87 (commented):
Content is fine by me. I agree it's probably slightly best to include this change in a sequent fork instead of Electra, but not by much.

PR #2702, opened 2025-04-23
fix(state-processor): log deneb1 only once

calbera:
Fixes a bug where Deneb1 was being logged every time a Deneb1 block was processed.

### DIFF
--- a/state-transition/core/state_processor_forks.go
+++ b/state-transition/core/state_processor_forks.go
@@ -22,7 +22,6 @@ package core
 
 import (
 	"fmt"
-	"sync"
 
 	"github.com/berachain/beacon-kit/consensus-types/types"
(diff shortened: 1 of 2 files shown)


state-transition/core/state_processor_forks.go
```diff
@@ -136,7 +135,7 @@ func (sp *StateProcessor) logDeneb1Fork(
 	// Since state fork is not updating to Deneb1, every block observes Deneb1 as "new fork" during
 	// Deneb1. Hence, we must wrap this in a OnceFunc to ensure it is logged only the first time
 	// we process a Deneb1 block.
-	sync.OnceFunc(func() {
```
calbera, line 139: the OnceFunc was being re-triggered every time this function is called and never saved for re-use.

rezbera (approved):

PR #2705, opened 2025-04-24
refactor(pectra): Move ExpectedWithdrawals to StateProcessor

rezbera:
This will allow us to access logging and metrics for behaviour in `StateProcessor.

### DIFF
--- a/state-transition/core/state_processor_withdrawals.go
+++ b/state-transition/core/state_processor_withdrawals.go
@@ -22,9 +22,12 @@ package core
 
 import (
 	ctypes "github.com/berachain/beacon-kit/consensus-types/types"
+	engineprimitives "github.com/berachain/beacon-kit/engine-primitives/engine-primitives"
 	"github.com/berachain/beacon-kit/errors"
+	"github.com/berachain/beacon-kit/primitives/common"
 	"github.com/berachain/beacon-kit/primitives/constants"
 	"github.com/berachain/beacon-kit/primitives/math"
+	"github.com/berachain/beacon-kit/primitives/version"
 	"github.com/berachain/beacon-kit/state-transition/core/state"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/ethereum/go-ethereum/params"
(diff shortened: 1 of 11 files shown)


state-transition/core/state_processor_withdrawals.go
```diff
@@ -156,6 +161,181 @@ func (sp *StateProcessor) processWithdrawals(
...
+			withdrawalIndex++
+		} else {
+			sp.logger.Info("consumePendingPartialWithdrawals: validator cannot withdraw partial balance",
+				"validator_pubkey", validator.GetPubkey().String(),
+				"balance", balance.Base10(),
+				"exit_epoch", validator.ExitEpoch.Base10(),
+				"has_sufficient_effective_balance", hasSufficientEffectiveBalance,
+				"has_excess_balance", hasExcessBalance,
+			)
+		}
```
rezbera, line 331: New logging capability

abi87:
> This will allow us to access logging and metrics for behaviour in `ExpectedWithdrawals`. If there's a simpler way to do this, happy to adjust

can't we just pass the logger in the `beacond/KvStore`?  It looks the simplest change to me.
I was considering  moving `ExpectedWithdrawal` to state processor but I do not like that either

rezbera:
> > This will allow us to access logging and metrics for behaviour in `ExpectedWithdrawals`. If there's a simpler way to do this, happy to adjust
> 
> can't we just pass the logger in the `beacond/KvStore`? It looks the simplest change to me. I was considering moving `ExpectedWithdrawal` to state processor but I do not like that either

will close this and try that

abi87:
> > > This will allow us to access logging and metrics for behaviour in `ExpectedWithdrawals`. If there's a simpler way to do this, happy to adjust
> > 
> > 
> > can't we just pass the logger in the `beacond/KvStore`? It looks the simplest change to me. I was considering moving `ExpectedWithdrawal` to state processor but I do not like that either
> 
> will close this and try that

we will need to pass the logger to KvStore upon initialization and also export it to allow use by stateDB, but it's the least invasive change I can see now

PR #2706, opened 2025-04-24
chore(chain-spec): Move chain spec to app.toml

calbera:
Removes usage of env var `CHAIN_SPEC` :))


camembera:
I like this, but would like a way to get a sample/template/current chainspec file generated, ie what's sent to `make start-custom`

camembera (approved):

calbera:
> I like this, but would like a way to get a sample/template/current chainspec file generated, ie what's sent to `make start-custom`

https://github.com/berachain/beacon-kit/blob/main/testing/networks/80087/spec.toml

PR #2707, opened 2025-04-24
Handle genesis case for validator/validator balances endpoints

nidhi-singh02:
All the five validator endpoints were not handling the genesis case correctly as genesis always used to point to slot 1.


beacon/blockchain/init_chain.go
```diff
@@ -71,33 +72,27 @@ func (s *Service) ProcessGenesisData(
...
 	genesisBlockRoot := genesisHeader.HashTreeRoot()
 
-	// Get the genesis validators root from the state.
-	genesisValidatorsRoot, err := genesisState.GetGenesisValidatorsRoot()
-	if err != nil {
-		return nil, nil, common.Root{}, common.Root{}, err
-	}
-
-	return validatorUpdates, genesisHeader, genesisValidatorsRoot, genesisBlockRoot, nil
+	return validatorUpdates, genesisHeader, genesisBlockRoot, genesisState, nil
```
calbera, line 97: rather than returning all 3 just return the genesisState because the others can be derived from state.

nidhi-singh02: yeah that's the plan to derive header and blockroot from genesis state. Before that refactor I wanted to test few cases, seems not getting correct balances from the genesisState due to which e2e tests are breaking.

nidhi-singh02: addressed, just returning genesisState and deriving values from state in `initChainer`

nidhi-singh02:
E2E tests are failing. In the case of validator balance APIs, when I am querying via "genesis" I am getting balances as 4000 BERA, however when I query via slot "1" we get 32 BERA as balance.

node-api/handlers/utils/id.go
```diff
@@ -41,6 +41,9 @@ var ErrNoSlotForStateRoot = errors.New("slot not found at state root")
 func SlotFromStateID[StorageBackendT interface {
 	GetSlotByStateRoot(root common.Root) (math.Slot, error)
 }](stateID string, storage StorageBackendT) (math.Slot, error) {
+	if stateID == StateIDGenesis {
+		return 0, errors.New("genesis slot not supported")
+	}
```
calbera, line 46: why?

node-api/backend/backend.go
```diff
@@ -43,11 +44,12 @@ type Backend struct {
...
-	genesisValidatorsRoot atomic.Pointer[common.Root]
-	genesisBlockRoot      atomic.Pointer[common.Root]
-	genesisTime           atomic.Pointer[math.U64]
-	genesisForkVersion    atomic.Pointer[common.Version]
+	genesisHeader      atomic.Pointer[ctypes.BeaconBlockHeader]
+	genesisBlockRoot   atomic.Pointer[common.Root]
+	genesisValidators  atomic.Pointer[[]*ctypes.Validator]
+	genesisTime        atomic.Pointer[math.U64]
+	genesisForkVersion atomic.Pointer[common.Version]
+	genesisState       atomic.Pointer[statedb.StateDB]
```
calbera, line 52: all of this (except `genesisTime`) may be consolidated into just genesisState, since everything else can be derived from genesisTime or genesisState

node-api/handlers/utils/id.go
nidhi-singh02: I assumed from our discussion - refactor the slotFromStateID func to alert the caller that if the input id is “genesis”. That's why added this check.  Did I miss something here ?

node-api/backend/backend.go
nidhi-singh02: done, deriving others from genesisTime or genesisState

node-api/handlers/utils/id.go
```diff
@@ -112,14 +112,10 @@ func IsTimestampIDPrefix(timestampID string) bool {
...
 
 // slotFromStateID returns a slot number from the given state ID.
-// In case of genesis, we will return 1 as the slot number. This is to ensure
-// validator API endpoints return a valid slot number for genesis.
 func slotFromStateID(id string) (math.Slot, error) {
 	switch id {
 	case StateIDFinalized, StateIDJustified, StateIDHead:
 		return Head, nil
-	case StateIDGenesis:
-		return 1, nil
```
abi87, line 122: This is the core issue:
- BeaconKit does not commit genesis state at height zero. Instead it keeps it in memory and commit it along with block 1 changes, at height 1.
- This explains the mapping above but it is not correct as any validator added with block 1 but not in genesis would appear to be in genesis.

PR #2709, opened 2025-04-24
refactor(pectra): clean up code of new state processing

calbera:
mainly code simplification for my own readability -- no logic changes


state-transition/core/state_processor_exits.go
```diff
@@ -29,49 +29,43 @@ import (
...
 func (sp *StateProcessor) InitiateValidatorExit(st *statedb.StateDB, idx math.ValidatorIndex) error {
+	// Return if the validator already initiated an exit, making sure to only exit validators once.
 	validator, err := st.ValidatorByIndex(idx)
 	if err != nil {
 		return err
 	}
-	// By this point, the state should be correctly updated with the fork.
+	if validator.GetExitEpoch() != constants.FarFutureEpoch {
+		return nil
+	}
```
calbera, line 39: noting this check is in spec (pre-electra) and will not affect backwards compat for us, so moving to outside of fork version check

calbera: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#initiate-validator-exit

calbera: removing this as it changes logic. dont want this PR to change logic.

state-transition/core/state_processor_exits.go
```diff
@@ -27,51 +27,48 @@ import (
...
-		// It did not add the `MinValidatorWithdrawabilityDelay`.
-		slot, slotErr := st.GetSlot()
-		if slotErr != nil {
-			return slotErr
-		}
-		nextEpoch := sp.cs.SlotToEpoch(slot) + 1
-		exitEpoch = nextEpoch
-		// The withdrawable Epoch is the next epoch after `exitEpoch`.
-		withdrawableEpoch = nextEpoch + 1
+		withdrawableEpoch = exitEpoch + sp.cs.MinValidatorWithdrawabilityDelay()
```
rezbera, line 67: much cleaner

rezbera (dismissed):
awesome

chain/spec.go
```diff
@@ -30,29 +30,30 @@ import (
...
+	// Gwei
+	MinActivationBalance() math.Gwei
 }
 
 type HysteresisSpec interface {
 	// HysteresisQuotient returns the quotient used in effective balance
 	// calculations to create hysteresis. This provides resistance to small
 	// balance changes triggering effective balance updates.
-	HysteresisQuotient() uint64
+	HysteresisQuotient() math.U64
```
rezbera, line 48: any reason for changing these specific ones to `math.U64` but leaving others? e.g.

https://github.com/berachain/beacon-kit/blob/bcd8e8ffb64d75b6f33f685ec9f77aa7ec5fd589/chain/spec.go#L109-L117

calbera: Changed it for how they are used. Things that are referring to a count of things I still feel can be uint64. Those used in computations with math.U64 (or aliases of) I feel should be returned as such; saves annoying conversions in the logic part of code.

state-transition/core/state_processor_payload.go
```diff
@@ -188,13 +188,13 @@ func (sp *StateProcessor) validateStatefulPayload(
...
 	// Verify RANDAO
-	slot, err := st.GetSlot()
+	epoch, err := st.GetEpoch()
 	if err != nil {
 		return err
 	}
 
 	expectedMix, err := st.GetRandaoMixAtIndex(
-		sp.cs.SlotToEpoch(slot).Unwrap() % sp.cs.EpochsPerHistoricalVector(),
+		epoch.Unwrap() % sp.cs.EpochsPerHistoricalVector(), // TODO: chain spec type.
```
rezbera, line 197: not sure what this TODO os asking for

state-transition/core/state_processor_withdrawals.go
```diff
@@ -126,21 +126,19 @@ func (sp *StateProcessor) processWithdrawals(
...
 	var nextValidatorIndex math.ValidatorIndex
-
-	// #nosec G115 -- won't overflow in practice.
-	if numWithdrawals == int(sp.cs.MaxWithdrawalsPerPayload()) {
+	if uint64(numWithdrawals) == sp.cs.MaxWithdrawalsPerPayload() {
 		// Next sweep starts after the latest withdrawal's validator index.
-		nextValidatorIndex = (expectedWithdrawals[numWithdrawals-1].
-			GetValidatorIndex() + 1) % math.ValidatorIndex(totalValidators)
+		nextValidatorIndex = expectedWithdrawals[numWithdrawals-1].GetValidatorIndex() + 1
+		nextValidatorIndex %= totalValidators
```
rezbera, line 132: preferred this in 1 operation. (i find the `%=` syntax confronting)

calbera: it follows the same pattern used a few lines below [here](https://github.com/berachain/beacon-kit/blob/refactor-val-exits/state-transition/core/state_processor_withdrawals.go#L141)

state-transition/core/state_processor_payload.go
calbera: removed in https://github.com/berachain/beacon-kit/pull/2709/commits/1d4ceac3eec9f23784f8d0b95a6836aaa09f49b3

state-transition/core/state_processor_withdrawals.go
calbera: dont mind switching back though https://github.com/berachain/beacon-kit/pull/2709/commits/9146a7c30213957986cbddce70c2e5ef269ddb98

state-transition/core/state_processor_withdrawals.go
```diff
@@ -228,33 +227,32 @@ func (sp *StateProcessor) processPartialWithdrawal(
...
-		// as we enforce that the `toWithdraw` amount is always above the `minActivationBalance`.
-		// For example, if a validator's balance is already at `minActivationBalance` (250K BERA on mainnet)
-		// and they request a partial withdrawal, the `toWithdraw` amount will always be zero.
-		withdrawableEpoch := math.Epoch(uint64(nextEpoch) + sp.cs.MinValidatorWithdrawabilityDelay())
+		// Note that we do not need to set the ExitEpoch anywhere here as Partial withdrawals do not
+		// exit the validator as we enforce that the `toWithdraw` amount is always above the
+		// `minActivationBalance`. For example, if a validator's balance is already at
+		// `minActivationBalance` (250K BERA on mainnet) and they request a partial withdrawal, the
+		// `toWithdraw` amount will always be zero.
+		withdrawableEpoch := currentEpoch + 1 + sp.cs.MinValidatorWithdrawabilityDelay()
```
rezbera, line 255: without the explicit `nextEpoch`, should add a comment that we add `1` so we start from the nextEpoch to align with spec. Otherwise could be conflated for magic number

calbera: good point, more clear with nextEpoch defined

chain/spec.go
rezbera: It's net improvement at the moment since clearly a reduction in conversion required, but a newcomer would definitely be confused above when to use which - the usage should be implied based on the type. Nothing worth changing for this PR though

state-transition/core/state_processor_withdrawals.go
```diff
@@ -159,22 +155,26 @@ func (sp *StateProcessor) processWithdrawals(
...
-	isFullExitRequest := amount == constants.FullExitRequestAmount
+	isFullExitRequest := withdrawalRequest.Amount == constants.FullExitRequestAmount
 	pendingPartialWithdrawals, err := st.GetPendingPartialWithdrawals()
 	if err != nil {
 		return err
 	}
+
 	// If partial withdrawal queue is full, only full exits are processed
 	if len(pendingPartialWithdrawals) == constants.PendingPartialWithdrawalsLimit && !isFullExitRequest {
+		// TODO: add metrics here.
```
rezbera, line 170: nit - use `TODO(pectra)` as that lets us filter old noise vs new

chain/spec.go
calbera: We can continue with these changes then. Another example is many of these are "max"s, meaning they get compared to `len(someSlice)`, meaning they should be stored and returned as `int`

state-transition/core/state_processor_withdrawals.go
calbera: https://github.com/berachain/beacon-kit/pull/2709/commits/26a4e557eb7d5d0cb214b4260a711607e0dd7e95

rezbera (approved):
LGTM🚀

PR #2716, opened 2025-04-28
chore(pectra): make gIndex calculations consistent with BeaconState

calbera:
Ensures that the Gindex calculations use the correct BeaconState schema for Deneb/Electra versions such that the GIndex values point to the correct data.


rezbera (approved):
Is there no impact to the test by making this change? No issues with this since no changes to the types

nidhi-singh02 (approved):
lgtm

abi87:
> Is there no impact to the test by making this change? No issues with this since no changes to the non-test types

Agree, although I wonder if it would be possible to consolidate these size data in order to update them automagically when we change prod code

calbera:
> Agree, although I wonder if it would be possible to consolidate these size data in order to update them automagically when we change prod code

This should be done throughout. Even in the [BeaconState file](https://github.com/berachain/beacon-kit/blob/main/consensus-types/types/state.go#L134), we are using hardcoded numbers

PR #2718, opened 2025-04-29
chore(pectra): Partial withdrawal failure metrics

calbera:
Need to add to prometheus collectors / grafana dashboards. 

New metric keys:
`beacon_kit.state.partial_withdrawal_request_dropped `: When the withdrawal queue is full, we drop any validators' partial withdrawal requests.

`beacon_kit.state.partial_withdrawal_failed`: When the validator's partial withdrawal request doesn't go through because the validators is _not_ partially withdrawable.

### DIFF
--- a/state-transition/core/metrics.go
+++ b/state-transition/core/metrics.go
@@ -38,24 +38,6 @@ func newStateProcessorMetrics(
 	}
 }
 
-func (s *stateProcessorMetrics) gaugeTimestamps(
-	payloadTimestamp uint64,
-	consensusTimestamp uint64,
-) {
-	// the diff can be positive or negative depending on whether the payload
-	// timestamp is ahead or behind the consensus timestamp
-	diff := int64(payloadTimestamp) - int64(consensusTimestamp) // #nosec G115
-	s.sink.SetGauge("beacon_kit.state.payload_consensus_timestamp_diff", diff)
-}
-
-func (s *stateProcessorMetrics) incrementValidatorNotWithdrawable() {
-	s.sink.IncrementCounter("beacon_kit.state.validator_not_withdrawable")
-}
-
-func (s *stateProcessorMetrics) incrementDepositStakeLost() {
-	s.sink.IncrementCounter("beacon_kit.state.deposit_stake_lost")
-}
-
 func (s *stateProcessorMetrics) gaugeBlockGasUsed(
 	blockNumber math.U64,
 	txGasUsed math.U64,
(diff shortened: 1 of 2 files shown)


state-transition/core/metrics.go
```diff
@@ -75,3 +57,33 @@ func (s *stateProcessorMetrics) gaugeBlockGasUsed(
...
+	s.sink.SetGauge("beacon_kit.state.payload_consensus_timestamp_diff", diff)
+}
+
+func (s *stateProcessorMetrics) incrementDepositStakeLost() {
+	s.sink.IncrementCounter("beacon_kit.state.deposit_stake_lost")
+}
+
+func (s *stateProcessorMetrics) incrementPartialWithdrawalRequestDropped() {
+	s.sink.IncrementCounter("beacon_kit.state.partial_withdrawal_request_dropped")
+}
```
rezbera, line 81: consider renaming to suggest that it's due to the partial withdrawal queue being full as that's the only case this is used

rezbera (approved):

camembera (approved):

state-transition/core/state_processor_withdrawals.go
```diff
@@ -101,6 +101,7 @@ func (sp *StateProcessor) processWithdrawals(
 			return getErr
 		}
 		updatedWithdrawals := ppWithdrawals[processedPartialWithdrawalsCount:]
+		sp.metrics.gaugePendingPartialWithdrawalsCount(len(updatedWithdrawals))
```
abi87, line 104: nit: we can use `processedPartialWithdrawalsCount` instead of `len(updatedWithdrawals)`

state-transition/core/metrics.go
```diff
@@ -75,3 +57,33 @@ func (s *stateProcessorMetrics) gaugeBlockGasUsed(
...
+
+func (s *stateProcessorMetrics) incrementPartialWithdrawalFailed() {
```
abi87, line 83: how about `incrementInvalidPartialWithdrawal`s?
I feel if people reads failed they would assume the chain is malfunctioning instead of understanding that there is something wrong with the withdrawal?

state-transition/core/metrics.go
abi87: I have not problem with dropped, seems pretty clear to me once you knowt there is a queue.
Maybe we should rename `gaugePendingPartialWithdrawalsCount` to something like `gaugePartialWithdrawalsEnqueued`

abi87: Also not sure we should explicitly track those withdrawals successfully carried out?
Not strictly needed as I believe that over a certain time window:
- end_withdrawals_enqueued - start_withdrawals_enqueued gives withdrawals arrived in the window
- end_withdrawals_dropped - start_withdrawals_dropped gives withdrawals dropped in the window
- their difference are the withdrawal successfully processed

abi87 (commented):
nits

PR #2719, opened 2025-04-29
chore(pectra): Add Logger to StateDB so we can log in ExpectedWithdrawals

rezbera:


node-core/components/backend.go
```diff
@@ -52,5 +54,6 @@ func ProvideStorageBackend(
 		in.BeaconStore,
 		in.DepositStore,
 		in.BlockStore,
+		in.Logger,
```
abi87, line 57: consider `in.Logger.With("backend", "storage")` to align this with what we do with other services.

abi87 (approved):
only a nit

PR #2721, opened 2025-04-30
chore(pectra): StateDB metrics

calbera:
New metric key:
- `beacon_kit.statedb.partial_withdrawal_request_invalid`: When the validator's partial withdrawal request doesn't go through because the validators is _not_ partially withdrawable. Mainly for the case when a validator does multiple partial withdrawals in a row, resulting in partial withdrawals that cumulatively add up to more than their balance.

### DIFF
--- a/state-transition/core/metrics.go
+++ b/state-transition/core/metrics.go
@@ -37,10 +37,6 @@ func newStateProcessorMetrics(sink TelemetrySink) *stateProcessorMetrics {
 }
 
 func (s *stateProcessorMetrics) gaugeBlockGasUsed(blockNumber, txGasUsed, blobGasUsed math.U64) {
-	if s.sink == nil {
-		return
-	}
-
 	blockNumberStr := blockNumber.Base10()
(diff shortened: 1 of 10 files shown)


state-transition/core/metrics.go
```diff
@@ -37,6 +37,10 @@ func newStateProcessorMetrics(sink TelemetrySink) *stateProcessorMetrics {
 }
 
 func (s *stateProcessorMetrics) gaugeBlockGasUsed(blockNumber, txGasUsed, blobGasUsed math.U64) {
+	if s.sink == nil {
+		return
+	}
```
rezbera, line 42: When is `s.sink` nil? Given `TelemetrySink` is an interface, we should just have a noop implementation if the intent is to ignore metric calls

rezbera (commented):
allg except the 
```
	if s.sink == nil {
		return
	}
```
should remove these if we can as it not the best pattern

rezbera (approved):

PR #2730, opened 2025-05-05
implement abci_query hander

benluelo:


consensus/cometbft/service/abci.go
```diff
@@ -143,11 +147,88 @@ func (s *Service) Commit(
 // NOOP methods
 //
 
-func (Service) Query(
-	context.Context,
-	*abci.QueryRequest,
+func (s *Service) Query(
```
calbera, line 150: can we add a NOTE: comment where this has been partially copied from ([here](https://github.com/cosmos/cosmos-sdk/blob/960d44842b9e313cbe762068a67a894ac82060ab/baseapp/abci.go#L168) I believe)

consensus/cometbft/service/abci.go
```diff
@@ -143,11 +147,88 @@ func (s *Service) Commit(
...
+	}
+
+	switch path[0] {
+	case "store":
+		// "/store" prefix for store queries
+		req.Path = "/" + strings.Join(path[1:], "/")
+		cms := s.sm.GetCommitMultiStore()
+		queryable, ok := cms.(storetypes.Queryable)
+		if !ok {
+			panic(ok)
```
calbera, line 182: might be better to return error here like [cosmos](https://github.com/cosmos/cosmos-sdk/blob/960d44842b9e313cbe762068a67a894ac82060ab/baseapp/abci.go#L1102) rather than panic

consensus/cometbft/service/abci.go
```diff
@@ -143,11 +147,88 @@ func (s *Service) Commit(
...
+		Log:       log,
+	}
+}
+
+// NOTE: Copied from here: https://github.com/cosmos/cosmos-sdk/blob/960d44842b9e313cbe762068a67a894ac82060ab/baseapp/abci.go#L1153-L1165
+//
+// SplitABCIQueryPath splits a string path using the delimiter '/'.
+//
+// e.g. "this/is/funny" becomes []string{"this", "is", "funny"}
+func SplitABCIQueryPath(requestPath string) []string {
```
calbera, line 223: nit: this could be private

consensus/cometbft/service/abci.go
```diff
@@ -143,11 +147,88 @@ func (s *Service) Commit(
...
+		sdkReq := storetypes.RequestQuery(*req)
+		resp, err := queryable.Query(&sdkReq)
+		if err != nil {
+			return queryResult(err), nil
+		}
+		resp.Height = req.Height
+
+		abciResp := abci.QueryResponse(*resp)
+
+		return &abciResp, nil
```
calbera, line 194: because we don't use named returns and the panic recovery will reassign the `resp` pointer, we should probably assign `resp` here and not return

consensus/cometbft/service/abci.go
```diff
@@ -143,11 +147,88 @@ func (s *Service) Commit(
...
+	}()
+
+	// when a client did not provide a query height, manually inject the latest
+	if req.Height == 0 {
+		req.Height = s.LastBlockHeight()
+	}
+
+	path := SplitABCIQueryPath(req.Path)
+	if len(path) == 0 {
+		return queryResult(errorsmod.Wrap(sdkerrors.ErrUnknownRequest, "no query path provided")), nil
```
calbera, line 172: for the panic recovery, should also assign `resp` before then returning `resp`

consensus/cometbft/service/abci.go
benluelo: ah this was just from our testing, forgot to copy the actual error handling in. will fix 👍

consensus/cometbft/service/abci.go
```diff
@@ -139,17 +143,97 @@ func (s *Service) Commit(
...
+			return queryResult(err), nil
+		}
+		queryResp.Height = req.Height
+
+		resp := abci.QueryResponse(*queryResp)
+
+		return &resp, nil
+
+	default:
+		resp = queryResult(errorsmod.Wrap(sdkerrors.ErrUnknownRequest, "unknown query path"))
```
abi87, line 196: consider refactoring this with something like
```golang
if path[0] != "store" {
   return err
}
...
```
to reduce indentation

benluelo: i had kept it like this to be a bit more inline with the original implementation, but this PR is superseded by #2737 anyways 👍

PR #2722, opened 2025-05-01
chore(pectra): Add the 7002 pre-deploy contract in local test genesis files

calbera:
Since the current devnet (uses testing/files) and kurtosis (uses devnet) start from Electra, adding the 7002 pre-deploy contract in those networks' eth genesis files.

PR #2733, opened 2025-05-06
fix(ci): Add Execution Request Predeploys for CI

rezbera:
Erigon in their latest release has started enforcing that execution request predeploys are present. This is the correct behaviour which other clients do not enforce


calbera:
Can we add these predeploys in all of the genesis files covered in https://github.com/berachain/beacon-kit/pull/2722

calbera (approved):

rezbera:
nethermind is encountering issues when the genesis file has predeploys at those specific addresses. Still investigating

rezbera:
Update there is something unusual about the nethermind image. When i clone nethermind locally and build the docker image via docker build -t nethermind:local . and use that in make start-nethermind, it works fine, but the remote image doesn't work as it fails due to state root mismatch.

rezbera:
For now, to keep CI unblocked, I'm going to remove nethermind from genesis files. @nidhi-singh02 will work in parallel to add it back

PR #2736, opened 2025-05-06
chore(block-biulding): minor cleanup

abi87:
Just trying to simplify the flow, mostly by avoding passing around as parameters data that can be retrieved from state.
This eliminates the possibility that we pass values that are incoherent with the state.

### DIFF
--- a/beacon/blockchain/payload.go
+++ b/beacon/blockchain/payload.go
@@ -195,8 +195,6 @@ func (s *Service) rebuildPayloadForRejectedBlock(
 	if _, _, err = s.localBuilder.RequestPayloadAsync(
 		ctx,
 		st,
-		// We are rebuilding for the current slot.
-		stateSlot,
 		nextPayloadTimestamp,
 		// We set the parent root to the previous block root. The HashTreeRoot
 		// of the header is the same as the HashTreeRoot of the block.
@@ -270,7 +268,6 @@ func (s *Service) optimisticPayloadBuild(
 	payload := blk.GetBody().GetExecutionPayload()
 	if _, _, err = s.localBuilder.RequestPayloadAsync(
 		ctx, st,
-		slot,
 		nextPayloadTimestamp,
 		// The previous block root is simply the root of the block we just
 		// processed.
--- a/payload/attributes/interfaces.go
+++ b/payload/attributes/interfaces.go
@@ -21,6 +21,8 @@
 package attributes
 
 import (
+	"github.com/berachain/beacon-kit/consensus-types/types"
+	engineprimitives "github.com/berachain/beacon-kit/engine-primitives/engine-primitives"
 	"github.com/berachain/beacon-kit/primitives/common"
 	"github.com/berachain/beacon-kit/primitives/math"
 )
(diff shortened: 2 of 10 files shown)


payload/attributes/interfaces.go
```diff
@@ -30,3 +32,11 @@ type ChainSpec interface {
 	EpochsPerHistoricalVector() uint64
 	SlotToEpoch(slot math.Slot) math.Epoch
 }
+
+// ReadOnlyBeaconState is the interface for a read-only beacon state.
+type ReadOnlyBeaconState interface {
+	GetSlot() (math.Slot, error)
+	ExpectedWithdrawals( /*timestamp*/ math.U64) (engineprimitives.Withdrawals, uint64, error)
```
calbera, line 39: no harm doing this I believe
```suggestion
	ExpectedWithdrawals(timestamp math.U64) (engineprimitives.Withdrawals, uint64, error)
```

payload/builder/payload.go
```diff
@@ -27,26 +27,34 @@ import (
...
-	statedb "github.com/berachain/beacon-kit/state-transition/core/state"
 )
 
 // RequestPayloadAsync builds a payload for the given slot and
 // returns the payload ID.
 func (pb *PayloadBuilder) RequestPayloadAsync(
 	ctx context.Context,
-	st *statedb.StateDB,
-	slot math.Slot,
+	st attributes.ReadOnlyBeaconState,
```
calbera, line 39: imo this package should define ReadOnlyBeaconState by itself and not use the interface defined in a sibling package

payload/attributes/interfaces.go
abi87: Done

payload/builder/payload.go
abi87: Agree, cleaned up interfaces

beacon/blockchain/payload.go
```diff
@@ -161,19 +161,15 @@ func (s *Service) rebuildPayloadForRejectedBlock(
...
-	stateSlot, err := st.GetSlot()
-	if err != nil {
-		return err
-	}
-
-	// Set the previous state root on the header.
 	latestHeader.SetStateRoot(st.HashTreeRoot())
+	if err = st.SetLatestBlockHeader(latestHeader); err != nil {
+		return fmt.Errorf("rebuildPayloadForRejectedBlock, failed updating latest block header: %w", err)
+	}
```
abi87, line 172: doing this makes sure we can call `GetLatestBlockHeader` later on without having to pass `latestHeader.HashTreeRoot()` explicitly

beacon/validator/config.go
```diff
@@ -26,7 +26,7 @@ const (
 
 	// defaultEnableOptimisticPayloadBuilds is the default
 	// for enabling the optimistic payload builder.
-	defaultEnableOptimisticPayloadBuilds = false
```
abi87, line 29: just a temp thing to test. Will revert as soon as tests pass

beacon/blockchain/payload.go
```diff
@@ -270,11 +266,7 @@ func (s *Service) optimisticPayloadBuild(
 	payload := blk.GetBody().GetExecutionPayload()
 	if _, _, err = s.localBuilder.RequestPayloadAsync(
 		ctx, st,
-		slot,
 		nextPayloadTimestamp,
-		// The previous block root is simply the root of the block we just
-		// processed.
-		blk.HashTreeRoot(),
```
abi87, line 277: In case you are wondering, we don't need to set latest header in state here because we call `ProcessSlots` which does it already

abi87:
Replaced by https://github.com/berachain/beacon-kit/pull/2819

PR #2737, opened 2025-05-06
feat(comet): Implement ABCI query endpoint

calbera:
Only `store/` proofs are relevant in beacon-kit. Implementation mostly taken from https://github.com/cosmos/cosmos-sdk/blob/960d44842b9e313cbe762068a67a894ac82060ab/baseapp/abci.go#L168. Adding e2e tests for comet APIs. 

Thanks to @benluelo and @aeryz for getting this started in https://github.com/berachain/beacon-kit/pull/2730.


consensus/cometbft/service/abci_utils.go
```diff
@@ -0,0 +1,91 @@
...
+	queryable, ok := s.sm.GetCommitMultiStore().(storetypes.Queryable)
+	if !ok {
+		return queryResult(
+			errorsmod.Wrap(sdkerrors.ErrUnknownRequest, "multi-store does not support queries"),
+		)
+	}
+
+	req.Path = "/" + strings.Join(path[1:], "/") // req.Path == "/beacon"
+
+	if req.Height <= 1 && req.Prove {
```
abi87, line 44: I understand this is copied from cosmos-sdk, but I wonder if this should be `s.initialHeight`?

PR #2740, opened 2025-05-07
node-api: partial pending withdrawals

nidhi-singh02:
https://ethereum.github.io/beacon-APIs/?urls.primaryName=v3.1.0#/Beacon/getPendingPartialWithdrawals along with E2E tests


node-api/backend/withdrawal.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func (b *Backend) PendingPartialWithdrawalsAtSlot(slot math.Slot) ([]*types.PendingPartialWithdrawalData, error) {
+	// Get the state at the given slot.
+	st, _, err := b.StateAtSlot(slot)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get state from slot %d", slot)
+	}
+
+	cTypePartialWithdrawals, err := st.GetPendingPartialWithdrawals()
```
abi87, line 38: @nidhi-singh02: note that this can be nil pre-Electra. However I think this is safe as we only loop over it, which is safe to do with nil slices

node-api/handlers/beacon/types/response.go
```diff
@@ -141,3 +142,28 @@ type Sidecar struct {
 type SidecarsResponse struct {
 	Data []*Sidecar `json:"data"`
 }
+
+// PendingPartialWithdrawalsResponse has a version field to indicate the fork version.
+// https://ethereum.github.io/beacon-APIs/#/Beacon/getPendingPartialWithdrawals
+type PendingPartialWithdrawalsResponse struct {
+	Version string `json:"version"`
```
abi87, line 149: isn't it kind of weird that we only have version over this specific response? Why others do no need Version?

node-api/handlers/beacon/withdrawal.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil, err
+	}
+
+	partialWithdrawals, err := h.backend.PendingPartialWithdrawalsAtSlot(slot)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get the genesis fork version.
+	genesisForkVersion, err := h.backend.GenesisForkVersion()
```
abi87, line 48: why we return the genesis fork version and not the latest version here?

node-api/handlers/beacon/types/response.go
nidhi-singh02: I see version in few other APIs as well like https://ethereum.github.io/beacon-APIs/#/Beacon/getPendingDeposits, seems it got introduced in newer APIs. One reason could be because the API will error out if state is fetched before electra.

node-api/backend/withdrawal.go
nidhi-singh02: As we discussed, I see a comment on GetPendingPartialWithdrawals saying it will error if called before electra. The beacon API spec expected behaviour is to error out if the state is retrieved before pectra. 

If it errors out, then we will return early before the loop, so we should be good. 
We need to test this case out.

abi87: Fixed

node-api/handlers/beacon/types/response.go
```diff
@@ -141,3 +142,28 @@ type Sidecar struct {
...
+	GenericResponse
+}
+
+type PendingPartialWithdrawalData struct {
+	ValidatorIndex  uint64 `json:"validator_index,string"`
+	Amount          uint64 `json:"amount,string"`
+	WithdrawalEpoch string `json:"withdrawal_epoch"`
```
calbera, line 156: Why only the first 2 have "string" json tag and the last one is type string? It seems in https://ethereum.github.io/beacon-APIs/#/Beacon/getPendingPartialWithdrawals that all three fields are decimal numbers returned as string. Just feel all 3 fields should be consistent

node-api/handlers/beacon/withdrawal.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	slot, err := utils.SlotFromStateID(req.StateID, h.backend)
+	if err != nil {
+		return nil, err
+	}
+
+	st, _, err := h.backend.StateAtSlot(slot)
```
calbera, line 46: this needs to reassign the `slot` variable (in case the requested slot is 0) to the second return value from `StateAtSlot`

calbera (changes requested):
small change but then mostly LGTM

node-api/handlers/beacon/withdrawal.go
nidhi-singh02: addressed, using `st, resolvedSlot, err := h.backend.StateAtSlot(slot)` where resolved slot is being passed to get the partial withdrawals.

node-api/handlers/beacon/types/response.go
nidhi-singh02: fair point, addressed, keeping everything as uint64 and json tag as "string"

calbera (dismissed):
the logic LGTM. let's get in a e2e test for this endpoint (should work as devnet starts out with electra). Ideally we submit a withdrawal tx from one of the validator credentials;  there may be other blockers getting this to work in e2e -- if this is the case just test that the endpoint returns empty successfully.

nidhi-singh02:
> the logic LGTM. let's get in a e2e test for this endpoint (should work as devnet starts out with electra). Ideally we submit a withdrawal tx from one of the validator credentials; there may be other blockers getting this to work in e2e -- if this is the case just test that the endpoint returns empty successfully.

e2e tests are here https://github.com/berachain/beacon-kit/pull/2760 , we submit a transaction check the pending partial withdrawals before it is empty and after the transaction can see one pending request.

node-api/handlers/beacon/withdrawal.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// Get the fork version.
+	forkVersion, err := st.GetFork()
+	if err != nil {
+		return nil, err
+	}
+
+	if version.IsBefore(forkVersion.CurrentVersion, version.Electra()) {
+		return nil, fmt.Errorf("%w: Electra fork not active yet", types.ErrInvalidRequest)
+	}
```
fridrik01, line 59: nit: I noticed we are calling `StateAtSlot` twice in the happy path here. Not a big deal but consider moving this fork check to happen inside `PendingPartialWithdrawalsAtSlot` and then we can remove the extra `StateAtSlot` call here

nidhi-singh02: Right. I have kept fetching the fork as in, felt PendingPartialWithdrawals should return only the pending partial withdrawals as it is part of `data` in the response and fork version is not part of it.
Instead now we are passing state into `PendingPartialWithdrawalsAtState` and the call to `StateAtSlot` is removed from inside this func.

nidhi-singh02:
E2E tests has been merged into this PR.

calbera (approved):

PR #2746, opened 2025-05-07
chore(block-building): minor cleanup

abi87:


beacon/blockchain/payload.go
```diff
@@ -263,17 +250,9 @@ func (s *Service) optimisticPayloadBuild(
 	}
 
 	// We then trigger a request for the next payload.
-	payload := blk.GetBody().GetExecutionPayload()
```
abi87, line 266: Right before this we run `stateprocessor.Transition` which actually store the payload header to state (see `processExecutionPayload`). So it's fine dropping payload here and load latest execution payload header from state in `RequestPayloadAsync`

payload/builder/payload.go
```diff
@@ -66,16 +64,21 @@ func (pb *PayloadBuilder) RequestPayloadAsync(
...
+
+	forkVersion := pb.chainSpec.ActiveForkVersionForTimestamp(timestamp)
 	req := ctypes.BuildForkchoiceUpdateRequest(
 		&engineprimitives.ForkchoiceStateV1{
-			HeadBlockHash:      headEth1BlockHash,
-			SafeBlockHash:      finalEth1BlockHash,
-			FinalizedBlockHash: finalEth1BlockHash,
+			HeadBlockHash:      lph.GetBlockHash(),
+			SafeBlockHash:      lph.GetParentHash(),
+			FinalizedBlockHash: lph.GetParentHash(),
```
abi87, line 81: this arrangement just fit, but it may change in the future:
- upon optimistic block building, blockHash it the block just verified (but not finalized), parent is already final
- upon pessimistic block building, blockHash is technically finalized but we never marked it as final (because of FinalizeBlock below)
- upon FinalizeBlock, blockHash is technically final but again we mark it as head

There is a possible future where we change this and mark blocks as final as soon as possible. If we ever do that, we could add a single extra parameter to handle the case. Until then, the changes in this PR should simplify code comprehension.

abi87:
Replaced by https://github.com/berachain/beacon-kit/pull/2819

PR #2748, opened 2025-05-07
api(withdrawals): Return code 400 if the state is prior to Electra

fridrik01:
To be merged in: https://github.com/berachain/beacon-kit/pull/2740

According to the [pending partial withdrawal spec](https://ethereum.github.io/beacon-APIs/?urls.primaryName=v3.1.0#/Beacon/getPendingPartialWithdrawals) we `Should return 400 if the state retrieved is prior to Electra.`

### Test plan

Set the electra fork time in the following files to happen in the future:

```
config/spec/devnet.go // <-- change devnetElectraForkTime
testing/files/eth-genesis.json // <-- pragueTime change pragueTime
```

Ran `make start` and `make start-geth` and queried the pending_partial_withdrawals before the fork timestamp

```
curl -s -X GET "http://localhost:3500/eth/v1/beacon/states/head/pending_partial_withdrawals"|jq
{
  "code": 400,
  "message": "invalid request: Electra fork not active yet"
}
```

After the fork timestamp:

```
curl -s -X GET "http://localhost:3500/eth/v1/beacon/states/head/pending_partial_withdrawals"|jq
{
  "version": "electra",
  "execution_optimistic": false,
  "finalized": true,
  "data": []
}
```

### DIFF
--- a/node-api/handlers/beacon/withdrawal.go
+++ b/node-api/handlers/beacon/withdrawal.go
@@ -21,9 +21,13 @@
 package beacon
 
 import (
+	"fmt"
+
 	"github.com/berachain/beacon-kit/node-api/handlers"
 	beacontypes "github.com/berachain/beacon-kit/node-api/handlers/beacon/types"
+	"github.com/berachain/beacon-kit/node-api/handlers/types"
 	"github.com/berachain/beacon-kit/node-api/handlers/utils"
+	"github.com/berachain/beacon-kit/primitives/version"
 )
 
 func (h *Handler) GetPendingPartialWithdrawals(c handlers.Context) (any, error) {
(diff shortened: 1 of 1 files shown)


node-api/handlers/beacon/withdrawal.go
```diff
@@ -39,18 +41,22 @@ func (h *Handler) GetPendingPartialWithdrawals(c handlers.Context) (any, error)
...
+	// Get the fork version.
+	forkVersion, err := st.GetFork()
 	if err != nil {
 		return nil, err
 	}
 
-	// Get the fork version.
-	forkVersion, err := st.GetFork()
+	if version.IsBefore(forkVersion.CurrentVersion, version.Electra()) {
+		return nil, types.ErrInvalidRequest
```
abi87, line 56: nit: should we extend the error with something like: `fmt.Errorf("%w: Electra fork not active yet", types.ErrInvalidRequest)`

PR #2754, opened 2025-05-09
refactor: replace map iteration with maps.Copy

yetyear:
Optimize code using a more modern writing style.


yetyear:
Thanks for your review!

PR #2760, opened 2025-05-09
e2e tests for partial pending withdrawals

nidhi-singh02:
In the E2E tests, we check pending partial withdrawals before the transaction, which is empty (As expected).
After submitting the transaction, can see pending partial withdrawals have 1 response.


testing/e2e/e2e_withdrawal_test.go
```diff
@@ -0,0 +1,263 @@
...
+	var blsPubkey crypto.BLSPubkey
+	for _, validator := range validatorsResp.Data {
+		credentials := validator.Validator.WithdrawalCredentials
+		if len(credentials) >= 2 && credentials[0] == 0x01 {
+			validatorIndex = fmt.Sprintf("%d", validator.Index)
+			// Convert the phase0.BLSPubKey to our crypto.BLSPubkey type
+			copy(blsPubkey[:], validator.Validator.PublicKey[:])
+			return validatorIndex, blsPubkey, nil
+		}
+	}
```
nidhi-singh02, line 151: it seems when users create validators through the deposit CLI, it creates withdrawal credentials with the 0x01 format by default (execution addresses). 
However I see test with non-ETH1 withdrawal credentials. Hence kept this check to ensure we get validator with execution withdrawal credentials for testing here.

testing/e2e/e2e_withdrawal_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+const (
+	// DefaultWithdrawalAmount is the default withdrawal amount in Gwei for testing
+	DefaultWithdrawalAmount = 1_000_000_000 // 1 BERA (10^9 Gwei)
+
+	// WithdrawalTxGasLimit is the gas limit for withdrawal transactions
+	WithdrawalTxGasLimit = 500_000
+
+	// BlocksToWaitAfterWithdrawal is the number of blocks to wait after a withdrawal
+	BlocksToWaitAfterWithdrawal = 3
```
abi87, line 57: nit: if this is related to some chain spec constant, can you reuse it instead of hard-coding values? It would help test maintenance IMO

testing/e2e/e2e_withdrawal_test.go
```diff
@@ -0,0 +1,263 @@
...
+// rpcWrapper wraps an rpc.Client and implements the rpcClient interface required by EIP7002.
+// EIP7002 requires a Call method that returns an error.
+type rpcWrapper struct {
+	*rpc.Client
+}
+
+// Call implements the rpcClient interface.
+func (r *rpcWrapper) Call(ctx context.Context, target any, method string, params ...any) error {
+	return r.Client.CallContext(ctx, target, method, params...)
+}
```
abi87, line 69: can you clarify why we need this wrapper? It does not seem to add anything really

testing/e2e/e2e_withdrawal_test.go
abi87: Fine, but is there a way to reuse the checks we have in production code instead of rewriting our own tests? Again in the interest of tests maintenability

nidhi-singh02: Can you pls elaborate, what checks are you referring to here ?

testing/e2e/e2e_withdrawal_test.go
nidhi-singh02: these are not chain spec constant. These are test-specific constants used in the E2E withdrawal test. They are used for testing purposes only.

testing/e2e/e2e_withdrawal_test.go
nidhi-singh02: Refer https://github.com/berachain/beacon-kit/blob/a3365685f34a51f8197d389fcb572bfbe936b079/execution/requests/eip7002/interfaces.go#L25

testing/e2e/e2e_withdrawal_test.go
```diff
@@ -0,0 +1,263 @@
...
+	if resp == nil {
+		return nil, errors.New("received nil response")
+	}
+
+	return resp, nil
+}
+
+// checkPendingPartialWithdrawals checks if there are pending partial withdrawals
+// and returns the list of pending withdrawals if any
+func (s *BeaconKitE2ESuite) checkPendingPartialWithdrawals(stateID string) ([]types.PendingPartialWithdrawalData, error) {
```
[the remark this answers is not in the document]
nidhi-singh02: got it, merged GET API call into this func and renamed it to `getPendingPartialWithdrawals`. Addressed in https://github.com/berachain/beacon-kit/pull/2760/commits/13198c6610261f272bc230e453926c5431b21932

testing/e2e/e2e_withdrawal_test.go
nidhi-singh02: Using method from consensus type  `IsValidEth1WithdrawalCredentials` for checking withdrawal creds. Let me know if you see any other checks that should be added.

testing/e2e/e2e_withdrawal_test.go
```diff
@@ -0,0 +1,262 @@
...
+	chainID, err := s.JSONRPCBalancer().ChainID(ctx)
+	s.Require().NoError(err)
+	signer := gethcore.NewPragueSigner(chainID)
+
+	// Get the sender's nonce
+	var nonce hexutil.Uint64
+	err = rpcClient.CallContext(ctx, &nonce, "eth_getTransactionCount",
+		common.HexToAddress("0x20f33ce90a13a4b5e7697e3544c3083b8f8a51d4"), "latest",
+	)
+	s.Require().NoError(err)
```
calbera, line 211: The eth1 EOA wallet used to sign the tx (in this case `0x20f33ce90a13a4b5e7697e3544c3083b8f8a51d4`) **needs** to be the exact address derived from the validator's own withdrawal credentials. Since we start the kurtosis devnet for all validators with the same credentials, I would comment [that](https://github.com/berachain/beacon-kit/blob/main/kurtosis/src/nodes/consensus/beacond/launcher.star#L73) here why its okay to use this wallet.

testing/e2e/e2e_withdrawal_test.go
```diff
@@ -0,0 +1,262 @@
...
+	fee, err := eip7002.GetWithdrawalFee(ctx, rpcWrapper)
+	s.Require().NoError(err)
+	s.T().Logf("Withdrawal fee: %s wei", fee.String())
+
+	// Create the withdrawal transaction data
+	withdrawalTxData, err := eip7002.CreateWithdrawalRequestData(blsPubkey, withdrawalAmount)
+	s.Require().NoError(err)
+
+	// Use a pre-loaded key that has funds
+	privateKey, err := ethcrypto.HexToECDSA("fffdbb37105441e14b0ee6330d855d8504ff39e705c3afa8f859ac9865f99306")
```
calbera, line 199: would explicitly say here in comments: not just any pre-loaded key with funds, but the private key **specifically of** the [credentials wallet](https://github.com/berachain/beacon-kit/blob/main/kurtosis/src/nodes/consensus/beacond/launcher.star#L73)

testing/e2e/e2e_withdrawal_test.go
nidhi-singh02: addressed. Added comments

testing/e2e/e2e_withdrawal_test.go
nidhi-singh02: makes sense, added comments.

nidhi-singh02:
Addressed the comments from Cal. Merging it.

PR #2766, opened 2025-05-12
chore(blockchain): adding UTs around RequestPayload calls

abi87:
Added unit tests around optimistic building of blocks.
Specifically we test the state attributes which contribute to shape the next block attributes for both accepted and rejected block case

### DIFF
--- a/payload/builder/interfaces.go
+++ b/payload/builder/interfaces.go
@@ -43,6 +43,9 @@ type ReadOnlyBeaconState interface {
 	ExpectedWithdrawals(timestamp math.U64) (engineprimitives.Withdrawals, uint64, error)
 	GetRandaoMixAtIndex(index uint64) (common.Bytes32, error)
 	GetLatestBlockHeader() (*ctypes.BeaconBlockHeader, error)
+
+	GetLatestExecutionPayloadHeader() (*ctypes.ExecutionPayloadHeader, error) // needed for UTs
+	EVMInflationWithdrawal(math.Slot) *engineprimitives.Withdrawal            // needed for UTs
 }
 
 // AttributesFactory is the interface for the attributes factory.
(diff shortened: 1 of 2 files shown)


beacon/blockchain/service.go
```diff
@@ -71,7 +71,6 @@ func NewService(
 	storageBackend StorageBackend,
 	blobProcessor BlobProcessor,
 	depositContract deposit.Contract,
-	eth1FollowDistance math.U64,
```
abi87, line 74: nit: this comes from chainSpec and it marginally simplifies UTs

state-transition/core/interfaces.go
```diff
@@ -49,6 +50,10 @@ type ReadOnlyContext interface {
 
 // ExecutionEngine is the interface for the execution engine.
 type ExecutionEngine interface {
+	NotifyForkchoiceUpdate( // added to simplify mocks
```
abi87, line 53: I am reusing the mock coming from this transactions in `blockchain` package, which requires `NotifyForkchoiceUpdate` method too.
This is counter to what we decided so far about interfaces, but doing it just to reduce Loc changed. Happy to have a different solution once I have written all the tests I care to have

payload/builder/interfaces.go
```diff
@@ -37,10 +37,15 @@ type PayloadCache interface {
 }
 
 type ReadOnlyBeaconState interface {
+	common.SSZRootable // needed for UTs
```
abi87, line 40: added a few methods that are needed in UTs

state-transition/core/state_processor_genesis.go
```diff
@@ -60,23 +60,7 @@ func (sp *StateProcessor) InitializeBeaconStateFromEth1(
 		return nil, err
 	}
 
-	versionable := ctypes.NewVersionable(genesisVersion)
```
abi87, line 63: Moved to `func GenesisBlockHeader` to be able to reuse it in UTs

abi87:
Soon to be replaced by https://github.com/berachain/beacon-kit/pull/2768

PR #2768, opened 2025-05-13
chore(blockchain): Added unit tests for optimistic block building

abi87:
The unit tests I wish I had before refactoring the optimistic payload building code flow


beacon/blockchain/payload_test.go
```diff
@@ -0,0 +1,450 @@
...
+		Body: &ctypes.BeaconBlockBody{
+			ExecutionPayload: &ctypes.ExecutionPayload{
+				Timestamp: math.U64(cs.GenesisTime() + 1),
+			},
+		},
+	}
+
+	// register async call to block building
+	var wg sync.WaitGroup          // useful to make test wait on async checks
+	var ch = make(chan struct{})   // useful to serialize build block goroutine and avoid data races
```
abi87, line 138: without this, UT seems to be having data races, but this is concerning.
The whole PR increases access to the state during the block building, but state is already access currently.
Do we already have a data race that this refactoring highlights?
Or does the data race happens in this refactoring?
Or is the data race detector wrong (like this can only happen in tests, not in prod code)?

beacon/blockchain/payload_test.go
```diff
@@ -0,0 +1,450 @@
...
+			Credentials: ctypes.NewCredentialsFromExecutionAddress(
+				common.ExecutionAddress{0x01},
+			),
+			Index: uint64(0),
+		},
+	}
+	genBytes, err := json.Marshal(genesisData)
+	require.NoError(t, err)
+	_, err = chain.ProcessGenesisData(ctx.ConsensusCtx(), genBytes)
+	require.NoError(t, err)
```
fridrik01, line 247: nit: consider refactoring this block since its almost exact copy from `TestOptimisticBlockBuildingRejectedBlockStateChecks`

fridrik01 (approved):

fridrik01 (approved):

PR #2777, opened 2025-05-19
feat(withdrawals): Allow for Disabling Withdrawals [DO NOT MERGE]

rezzmah:
This PR sets up the relevant boilerplate such that withdrawals can be disabled if needed

DO NOT MERGE till needed


chain/spec.go
```diff
@@ -469,3 +473,18 @@ func (s spec) EVMInflationPerBlock(timestamp math.U64) math.Gwei {
...
+func (s spec) WithdrawalsEnabled(_ math.U64) bool {
+	return true
+	/*
+		Example:
+			 fv := s.ActiveForkVersionForTimestamp(timestamp)
+			 switch fv {
+			 case version.Electra1():
+				 return false
+			 default:
+			 	return true
```
abi87, line 488: I get the idea but not sure this would be backward, if we activate Pectra and some withdrawals are made before this hard fork. We'd have to go by slot I think? Or timestamp

abi87 (commented):
I would expect a proper hard fork, with a (list of) timestamps specified in specs for when withdrawals should be activate/disactivated right?

config/spec/defaults.go
```diff
@@ -56,7 +56,10 @@ const (
 	defaultTargetSecondsPerEth1Block = 2 // Berachain specific.
 
 	// Fork-related values.
-	defaultElectraForkTime = 9999999999999999 // Set as a future timestamp as not yet determined.
+	defaultFarFutureTimestamp                = 9999999999999999 // a future timestamp as not yet determined.
+	defaultElectraForkTime                   = defaultFarFutureTimestamp
+	defaultElectraDisableWithdrawalsForkTime = defaultFarFutureTimestamp
```
rezbera, line 61: To enable and disable on ETH Mainnet, update these values to a relevant timestamp

testing/networks/80094/spec.toml
```diff
@@ -35,6 +35,8 @@ target-seconds-per-eth1-block = 2
 genesis-time = 1_737_381_600
 deneb-one-fork-time = 1_738_415_507
 electra-fork-time = 9_999_999_999_999_999
+electra-disable-withdrawals-fork-time = 9_999_999_999_999_999
+electra-enable-withdrawals-fork-time = 9_999_999_999_999_999
```
rezbera, line 39: Also update these these value to configure the fork on mainnet

config/spec/defaults.go
rezbera: And this 
https://github.com/berachain/beacon-kit/pull/2777/files#r2098730113

rezbera:
> I would expect a proper hard fork, with a (list of) timestamps specified in specs for when withdrawals should be activate/disactivated right?

I have added this @abi87 @calbera

chain/data.go
```diff
@@ -95,6 +95,10 @@ type SpecData struct {
 	Deneb1ForkTime uint64 `mapstructure:"deneb-one-fork-time"`
 	// ElectraForkTime is the time at which the Electra fork is activated.
 	ElectraForkTime uint64 `mapstructure:"electra-fork-time"`
+	// ElectraDisableWithdrawalsForkTime is the time at which withdrawals were first disabled (if disabled).
+	ElectraDisableWithdrawalsForkTime uint64 `mapstructure:"electra-disable-withdrawals-fork-time"`
+	// ElectraEnableWithdrawalsForkTime is the time at which withdrawals were enabled after disabling
+	ElectraEnableWithdrawalsForkTime uint64 `mapstructure:"electra-enable-withdrawals-fork-time"`
```
abi87, line 101: we should enforce somewhere that Disable Time <=  Enable Time

chain/helpers.go
```diff
@@ -38,6 +44,16 @@ func (s spec) ActiveForkVersionForTimestamp(timestamp math.U64) common.Version {
...
 
+// WithdrawalsEnabled is a switch that can be used to freeze withdrawals in an emergency scenario.
+// An exception is made for the EVM inflation withdrawal which is always active.
+func (s spec) WithdrawalsEnabled(timestamp math.U64) bool {
+	time := timestamp.Unwrap()
+	if time >= s.ElectraDisableWithdrawalsForkTime() && time < s.ElectraEnableWithdrawalsForkTime() {
+		return false
+	}
+	return true
+}
```
abi87, line 55: nit: we are really defining a window where withdrawals are disabled, so we may consider a negative logic here and have a WithdrawalDisabled method

chain/data.go
abi87: Also wonder if we should make scaffolding for multiple periods of withdrawal pause?

abi87: Also consider renaming this to `DisableWithdrawalsRequestsWindowStart` and ``DisableWithdrawalsRequestsWindowEnd`.
I do not believe this is strictly related to Electra.
Also we should stress this is a disabling window

state-transition/core/state_processor_staking.go
```diff
@@ -67,8 +67,10 @@ func (sp *StateProcessor) processOperations(
 		}
 	}
 
+	// After Electra, validators can request withdrawals through execution requests which must be handled.
+	// If withdrawals are enabled, process the withdrawals. Otherwise, the withdrawal requests are ignored
+	// to prevent withdrawals from excessively queuing up.
```
abi87, line 72: is this true? It seems to me we just let withdrawals be enqueued rn.
I am not against dropping requests but I don't think we are doing this now

abi87 (changes requested):
Some points:
- request to rename disabling window
- question about disabling not just servicing of withdrawals but enqueing of requests as well (it's in a comment but not in the code)

Finally I wonder if we should prepare the specs to allow for multiple stopping window, but making two lists of Disabling window start and disabling window ends

state-transition/core/state_processor_staking.go
rezbera: good catch, i removed the logic but forgot to update comment

chain/data.go
rezbera: > we should enforce somewhere that Disable Time <= Enable Time

This should be done as a larger ordering enforcement of forks. e.g. geth https://github.com/ethereum/go-ethereum/blob/516451dc3a514c7c122f28864ea76742a027b858/params/config.go#L702

rezbera: > Also, in terms of operations we just need to specify the start window at first, right? We can leave the end to farFuture and reset that once the window is closed

correct

rezbera:
> Some points:
> 
> * request to rename disabling window
> * question about disabling not just servicing of withdrawals but enqueing of requests as well (it's in a comment but not in the code)
> 
> Finally I wonder if we should prepare the specs to allow for multiple stopping window, but making two lists of Disabling window start and disabling window ends

Addressed 1 and 2.

With regards to 3, we can always retroactively refactor the process if we end up doing a first withdrawal fork

chain/data.go
calbera: Enforcement of fork timestamps for now can be done in the `(s *spec) validate() error` function we already have. Good way to make use of that.

calbera (commented):
General question. Do these 2 chain spec values imply we can "re-use" them to do a withdrawals freeze in the future? I.e. say we recognize an issue at t=100. Then we say we will disable at t=150 and enable at t=250. Now later on if t=1000, and we need to freeze again, we could just set the disable=1100 and enable to 1200.

Is this the general thinking?

rezbera:
> General question. Do these 2 chain spec values imply we can "re-use" them to do a withdrawals freeze in the future? I.e. say we recognize an issue at t=100. Then we say we will disable at t=150 and enable at t=250. Now later on if t=1000, and we need to freeze again, we could just set the disable=1100 and enable to 1200.
> 
> Is this the general thinking?

No, there is no intention to re-use these specific variables. As you're aware, that would cause issues with syncing. If we reach the unfortunate position where we need multiple freezes, we can refactor the fork retroactively similar to the way we did forking by timestamp

chain/data.go
rezbera: https://github.com/berachain/beacon-kit/pull/2790

PR #2778, opened 2025-05-20
chore(storage): Consolidate deposits storage 1/3

abi87:
Just an initial refactoring to:

- Move current deposit store implementation to v1 folder
- use `deposit.StoreManager` interface across the codebase instead of the v1 implementation
- Improve state-transition package unit tests to retrieve deposits root from GetDepositsByIndex call


storage/deposit/store.go
```diff
@@ -22,144 +22,107 @@ package deposit
...
+}
 
-	// closeFunc is a closure that closes the underlying database
-	// used by store to ensure that all writes are flushed to disk.
-	// We guarantee that closeFunc is called at maximum only once.
-	closeFunc CloseFunc
-	once      sync.Once
+type StoreManager interface {
+	Store
+}
```
abi87, line 50: for the time being `StoreManager` interface seems redundant. However as soon as we will introduce deposit store v2, it will be useful to handle data migration logic. So proactivelly using it in the rest of the codebase in stead of Store

storage/deposit/v1/store.go
```diff
@@ -0,0 +1,162 @@
...
+type CloseFunc func() error
+
+// NewStore creates a new deposit store.
+func NewStore(
+	baseDB dbm.DB,
+	logger log.Logger,
+) *KVStore {
+	spdbV1 := depositstorecommon.NewSynced(baseDB)
+	kvsp := NewKVStoreProvider(spdbV1)
+	closeFunc := spdbV1.Close
```
abi87, line 65: Moved these here from ProvideDepositStore file

storage/deposit/v1/store.go
```diff
@@ -0,0 +1,162 @@
...
+
+// KVStore is a simple KV store based implementation that assumes
+// the deposit indexes are tracked outside of the kv store.
+type KVStore struct {
+	store sdkcollections.Map[uint64, *ctypes.Deposit]
+
+	// closeFunc is a closure that closes the underlying database
+	// used by store to ensure that all writes are flushed to disk.
+	// We guarantee that closeFunc is called at maximum only once.
+	closeFunc CloseFunc
```
abi87, line 48: moved mutex to generalStore since we only access this store via StoreManager interface

state-transition/core/state_processor_payload_test.go
```diff
@@ -142,11 +143,19 @@ func TestPayloadTimestampVerification(t *testing.T) {
 				WithVerifyResult(false).
 				WithMeterGas(false)
 
+			var depRoot common.Root
```
abi87, line 146: technically this is what happens in prod:
- we finalize a block and pull its deposits
- we expect those deposits to be included and processed in the next block

I started these UTs before clearly understanding this

storage/deposit/store.go
```diff
@@ -22,144 +22,107 @@ package deposit
...
-			"EnqueueDeposit", "enqueued", len(deposits),
-			"start", deposits[0].GetIndex(), "end", deposits[len(deposits)-1].GetIndex(),
-		)
-	}
-	return nil
+func (gs *generalStore) Close() error {
+	gs.mu.Lock()
+	defer gs.mu.Unlock()
+
+	return gs.storeV1.Close()
```
fridrik01, line 115: Should we not always just close one db at a time? Like when migrating from V1 to V2 I think we should run close on V1 during the migration (to make sure all fds are flushed), and not in this method... So here we should consider checking currentVersion and closing the relevant implementation only.

fridrik01 (commented):
Looks good to me, just had some questions regarding how we handle cleanup when migrating between versions

fridrik01 (approved):
lgtm

storage/deposit/store.go
```diff
@@ -22,144 +22,104 @@ package deposit
...
-	if _, err := schemaBuilder.Build(); err != nil {
-		panic(errors.Wrap(err, "failed building KVStore schema"))
-	}
-	return res
+// We have changed in time the way we stored deposits. generalStore is meant to offer
+// a single way to access deposits and to handle the data migration among versions when needed
+type generalStore struct {
+	// mu protects storex for concurrent access
+	mu             sync.RWMutex
+	currentVersion uint8
```
rezbera, line 64: I expect current version would be updated by hard fork in future PR?

storage/deposit/store.go
```diff
@@ -22,144 +22,104 @@ package deposit
...
+	logger         log.Logger
 }
 
-// Close closes the store by calling the closeFunc. It ensures that the
-// closeFunc is called at most once.
-func (kv *KVStore) Close() error {
-	var err error
-	kv.once.Do(func() { err = kv.closeFunc() })
-	return err
+func NewStore(dbV1 dbm.DB, logger log.Logger) StoreManager {
```
rezbera, line 69: Consider making `generalStore` public and returning here instead, i.e. `Accept interfaces, return structs.`

Not a hard blocker, just feels odd that v1 store is public and generalStore is not since no one should be using v1 store directly (except tests which is the likely reason here). I expect general store will need to be public as well to test it

storage/deposit/v1/store.go
```diff
@@ -0,0 +1,162 @@
...
+		endIdx   = startIndex + depRange
+	)
+
+	for i := startIndex; i < endIdx; i++ {
+		deposit, err := kv.store.Get(ctx, i)
+		switch {
+		case err == nil:
+			deposits = append(deposits, deposit)
+		case errors.Is(err, sdkcollections.ErrNotFound):
+			return deposits, deposits.HashTreeRoot(), nil
```
rezbera, line 116: This is expected to occur when we don't fill up the block with maximum deposits right?, i.e. a normal happy path

rezbera (approved):
Some tests look like they are failing so will leave review of test changes but all the runtime logic changes LGTM

state-transition/core/validation_deposits.go
```diff
@@ -122,7 +122,7 @@ func ValidateNonGenesisDeposits(
 	}
 
 	// Finally check that the historical deposits root matches locally what's on the beacon block.
-	if !localDeposits.HashTreeRoot().Equals(blkDepositRoot) {
+	if !depRoot.Equals(blkDepositRoot) {
```
calbera, line 125: nit naming for clarity:
```suggestion
	if !localDepositRoot.Equals(blkDepositRoot) {
```

storage/deposit/store.go
```diff
@@ -22,144 +22,107 @@ package deposit
...
-	"github.com/berachain/beacon-kit/storage/encoding"
+	"github.com/berachain/beacon-kit/primitives/common"
+	depositstorev1 "github.com/berachain/beacon-kit/storage/deposit/v1"
+	dbm "github.com/cosmos/cosmos-db"
 )
 
-const KeyDepositPrefix = "deposit"
+const (
+	unset uint8 = 0
+	V1    uint8 = 1
```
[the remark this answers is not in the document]
abi87: Done

storage/deposit/store.go
abi87: The idea for now is that store can introspect and find out what the version is.
This is possible because we always have at least a validator i.e. a deposit.
There is also a simple way to check whether the migration is incomplete (although it is coded to happen atomically over a single block).
So current version will stay unexported but can be duly set

calbera (approved):

PR #2779, opened 2025-05-20
chore(storage): Consolidate deposits storage 2/3

abi87:
Introduce deposit store v2 which leverate a IAVL store to efficiently calculate deposits root.
While v1 needs to load all historical deposits and hash them together to get said root, which is a O(N) operation, the merkleDB allows doing it with O( log N) complexity (what it takes to recalculate the root of an append only tree), at the cost of somewhat larger storage (in a complete binary tree, there are as many intermediate nodes as leaves, so say double storage).
Data migration code is added, which allow a coordinate transfer of deposits from storeV1 to storeV2 but code is not hooked up to production code.


calbera:
closing in favor of #2794

PR #2780, opened 2025-05-21
fix(test): Fix Deposit Contract Predeploy Code

rezbera:
The Deposit Contract Predeploy Code in several tests is different to that of testnet and mainnet. 

This issue was discovered as I surprised to find I could deposit 1 BERA minimum, when the minimum should have been 10_000 BERA.


testing/e2e/e2e_staking_test.go
```diff
@@ -65,9 +65,7 @@ func (s *BeaconKitE2ESuite) TestDepositRobustness() {
 	weiPerGwei := big.NewInt(1e9)
 
 	// This value is determined by the MIN_DEPOSIT_AMOUNT_IN_GWEI variable from the deposit contract.
-	//
-	// TODO: fix the genesis file to use the correct deposit contract.
```
nidhi-singh02, line 69: haha! we had a TODO, nice catch.

nidhi-singh02 (approved):
lgtm!

PR #2782, opened 2025-05-21
bug fix: undefined validation function 'hex'

nidhi-singh02:


abi87 (changes requested):
Please add tests to show that crash was there and this PR solves it

nidhi-singh02:
> Please add tests to show that crash was there and this PR solves it

Added E2E tests.

node-api/handlers/beacon/header.go
```diff
@@ -21,35 +21,111 @@
...
 	}
-	slot, err := math.U64FromString(req.Slot)
+
+	// Determine the slot based on the query params.
+	slot, err := h.determineSlot(&req)
 	if err != nil {
 		return nil, err
 	}
+
 	header, err := h.backend.BlockHeaderAtSlot(slot)
```
fridrik01, line 106: According to spec we should return a 404 if the block was not found. Right now we return code 500 here.

storage/block/store.go
```diff
@@ -121,3 +130,14 @@ func (kv *KVStore[BeaconBlockT]) GetSlotByStateRoot(
...
 	return slot, nil
 }
+
+// GetSlotByParentRoot retrieves the slot by a given parent root from the store.
+func (kv *KVStore[BeaconBlockT]) GetSlotByParentRoot(
+	root common.Root,
+) (math.Slot, error) {
+	slot, ok := kv.parentRoots.Peek(root)
+	if !ok {
+		return 0, fmt.Errorf("slot not found at parent root: %s", root)
```
fridrik01, line 140: IIUC, this could also mean that beacond was restarted as this is stored in memory only

node-api/handlers/beacon/header.go
nidhi-singh02: My bad, I referred the incorrect endpoint. 
https://ethereum.github.io/beacon-APIs/?urls.primaryName=v3.1.0#/Beacon/getBlockHeaders This is the correct endpoint, it doesn't have 404 response code.

Updated the description of the PR.

storage/block/store.go
nidhi-singh02: I see. Should we change the error message in that case ? Or this looks good.

PR #2783, opened 2025-05-22
chore(deposits): Add Metric and Log for Excess Balance Withdrawal

rezbera:


state-transition/core/state/metrics.go
```diff
@@ -23,3 +23,9 @@ package state
 func (s *StateDB) incrementPartialWithdrawalRequestInvalid() {
 	s.telemetrySink.IncrementCounter("beacon_kit.statedb.partial_withdrawal_request_invalid")
 }
+
+// incrementExcessValidatorStakeWithdrawal increments the telemetry counter when a withdrawal is created
+// because a validator's stake went over the MaxEffectiveBalance.
+func (s *StateDB) incrementExcessValidatorStakeWithdrawal() {
```
abi87, line 29: nit: should we stick with current naming and call this `incrementPartialWithdrawal`?

calbera:
```suggestion
func (s *StateDB) incrementExcessStakePartialWithdrawal() {
```
- "validator" is already implied
- adding "partial" to indicate partial/not full

calbera (approved):
nit but otherwise LGTM

PR #2790, opened 2025-05-28
chore(forks): Enforce Fork Ordering 

rezzmah:
As discussed here https://github.com/berachain/beacon-kit/pull/2777#discussion_r2103472349

Just like Geth, we should enforce fork ordering.


calbera (approved):
Good stuff

PR #2793, opened 2025-05-28
chore: Stable block time pt2

abi87:
- [x] appeased linter
- [x] move block delay params to config to ease up testing


consensus/cometbft/service/delay/config.go
```diff
@@ -0,0 +1,101 @@
...
+	// 0, then `timeout_commit` from the config will be used, which is not what
+	// we want since we're trying to control the block time.
+	noDelay = 1 * time.Microsecond
+)
+
+type ConfigGetter interface {
+	SbtMaxDelayBetweenBlocks() time.Duration
+	SbtTargetBlockTime() time.Duration
+	SbtConstBlockDelay() time.Duration
+	SbtConsensusParamUpdate() int64
```
fridrik01: Should this be called `SbtConsensusUpdateHeight` for consistency?

config/spec/devnet.go
```diff
@@ -68,6 +68,9 @@ func DevnetChainSpecData() *chain.SpecData {
 	specData := MainnetChainSpecData()
 	specData.DepositEth1ChainID = DevnetEth1ChainID
 
+	specData.Config.ConsensusUpdateHeight = 1
+	specData.Config.ConsensusEnableHeight = 1
```
fridrik01: IIRC this didn't work in devnet as the ConsensusEnableHeight needed to be 2

consensus/cometbft/service/commit.go
```diff
@@ -47,6 +47,12 @@ func (s *Service) commit(
 
 	s.finalizeBlockState = nil
 
+	if s.blockDelay != nil {
+		if err := s.sm.SaveBlockDelay(s.blockDelay.ToBytes()); err != nil {
```
fridrik01, line 51: Thinking out loud here, but do we need to set this each time we commit?

consensus/cometbft/service/configs.go
```diff
@@ -121,6 +121,9 @@ func DefaultConsensusParams(consensusKeyAlgo string) *cmttypes.ConsensusParams {
 	res.Synchrony.Precision = precision
 	res.Synchrony.MessageDelay = messageDelay
 
+	// Activete Stable Block Time (SBT) by default.
```
fridrik01:
```suggestion
	// Activate Stable Block Time (SBT) by default.
```

consensus/cometbft/service/delay/config.go
```diff
@@ -0,0 +1,101 @@
...
+	SbtConsensusEnableHeight() int64
+}
+
+type Config struct {
+	MaxDelay              time.Duration `mapstructure:"max-delay"`
+	TargetBlockTime       time.Duration `mapstructure:"target-block-time"`
+	ConstBlockDelay       time.Duration `mapstructure:"const-block-delay"`
```
fridrik01: Did we consider removing the `ConstBlockDelay` from the Config struct and instead use the existing `TimeoutCommit` until the SBTEnableHeight is reached?

consensus/cometbft/service/delay/delay.go
```diff
@@ -0,0 +1,136 @@
...
+	"encoding/binary"
+	"fmt"
+	"time"
+)
+
+// BlockDelay is used to calculate the delay before proposing the next block.
+//
+// The general formula is:
+//
+//	NextBlockTime = InitialTime + TargetBlockTime * (CurrentHeight - InitialHeight)
```
fridrik01, line 34: maybe add here a comment in case the block delay reaches MaxDelay this struct will be reset with a new initial time/height/previoustime

consensus/cometbft/service/delay/delay.go
```diff
@@ -0,0 +1,136 @@
...
+type BlockDelay struct {
```
fridrik01, line 37: Consider using non exported fields and adding a Constructor instead since we want to set all of the fields at once.

consensus/cometbft/service/delay/config.go
```diff
@@ -0,0 +1,101 @@
...
+type ConfigGetter interface {
+	SbtMaxDelayBetweenBlocks() time.Duration
+	SbtTargetBlockTime() time.Duration
+	SbtConstBlockDelay() time.Duration
+	SbtConsensusParamUpdate() int64
+	SbtConsensusEnableHeight() int64
+}
+
+type Config struct {
+	MaxDelay              time.Duration `mapstructure:"max-delay"`
```
fridrik01: Should we call this `MaxDelayBetweenBlock` for consistency?

consensus/cometbft/service/delay/delay.go
```diff
@@ -0,0 +1,136 @@
...
+		d.InitialHeight = curBlockHeight
+		d.PreviousBlockTime = curBlockTime
+		return cfg.SbtTargetBlockTime()
+	}
+
+	d.PreviousBlockTime = curBlockTime
+
+	t := d.InitialTime.Add(cfg.SbtTargetBlockTime() * time.Duration(curBlockHeight-d.InitialHeight))
+	if curBlockTime.Before(t) {
+		return t.Sub(curBlockTime)
```
fridrik01: This may be more readable as `t.Sub` requires mental calculation to understand it's computing a delay. Also added a comment to make it clearer that this delay can be relatively long 

```suggestion                
                // if we are behind on block production (due to network restart, byzantine validators proposing too fast) we make up for it with a single large delay (maximum up to SbtMaxDelayBetweenBlocks)
                next_block_delay := t.Sub(curBlockTime)
                
		return next_block_delay 
```

consensus/cometbft/service/delay/delay.go
```diff
@@ -0,0 +1,136 @@
...
+
+// FromBytes converts the bytes to a blockDelay.
+//
+// Expected format:
+//
+//	InitialTime (int64) | InitialHeight (int64) | PreviousBlockTime (int64)
+//	(little endian)
+//
+// It returns ErrBlockDelayFromBytes if it fails to read from the buffer.
+func FromBytes(bz []byte) (*BlockDelay, error) {
```
fridrik01, line 80: Can we move the FromBytes/ToBytes to a separate bytes/serialization file or at least move `FromBytes` to come after `Next` here below

consensus/cometbft/service/delay/config.go
```diff
@@ -0,0 +1,101 @@
...
+	targetBlockTime = 2 * time.Second
+
+	// Delay to use before the upgrade to SBT.
+	constBlockDelay = 500 * time.Millisecond
+
+	// Height to enable SBT. Changes will be applied in the next block after this
+	// (10_000_001).
+	sbtEnableHeight = math.MaxInt64
+
+	sbtConsensusParamUpdate = math.MaxInt64
```
fridrik01: Would be nice to have a comment here that `sbtConsensusParamUpdate` basically decides the height where we set the sbtEnableHeight which in turn initializes the BlockDelay struct which will then be used to calculate block delay in the next height.

consensus/cometbft/service/commit.go
abi87: Yes, if not a restart may cause the node to provide a different value than other nodes (e.g. if delay reference is reset)

consensus/cometbft/service/delay/config.go
abi87: It's actually the other way around. TimeoutCommit is removed from comet configs (already merged in cometBFT repo fork) and a warning is issued if that config is set.
We don't want to make that delay configurable. We just freeze it to what its default was and move to the new system

consensus/cometbft/service/delay/config.go
abi87: Renamed to `MaxBlockDelay`

consensus/cometbft/service/delay/delay.go
abi87: Moved Next up

consensus/cometbft/service/delay/delay.go
abi87: The comment seems off to me, we don't cap the maximum delay. Restructured the code as indicated

chain/data.go
```diff
@@ -20,11 +20,16 @@
...
-import "github.com/berachain/beacon-kit/primitives/common"
+import (
+	"github.com/berachain/beacon-kit/consensus/cometbft/service/delay"
+	"github.com/berachain/beacon-kit/primitives/common"
+)
 
 // SpecData is the underlying data structure for chain-specific parameters. All fields with a
 // `mapstructure` tag are required.
 type SpecData struct {
+	delay.Config `mapstructure:"block-delay-configuration"`
```
calbera, line 31: doesn't feel correct that the cometbft/service config is imported here and in the chain spec. Can we move this config to the comet app config?

consensus/cometbft/service/delay/delay.go
abi87: We have tests running on a different namespace (`something-something_test) and we need to assert on these fields e.g. after Next is called. Hence we need them exported

chain/data.go
abi87: I see that stylistically this config is handled differently from the others, still I think this arrangement is the best:

- embedding `delay.Config` here is just a way to cut down code duplication.
- This is not really a cometBFT only config (which is actually only parsed by CometBFT code). The majority logic of next block delay is an app thing and only one of the activation heights is strictly speaking a consensus quantity.

abi87:
Replaced by https://github.com/berachain/beacon-kit/pull/2422

PR #2794, opened 2025-05-28
feat(deposits): Migration to EIP-6110 style processing

### LINKED ISSUE #2672: EIP-6110: Supply validator deposits on chain
Brings https://eips.ethereum.org/EIPS/eip-6110 to BeaconKit

calbera:
### 3 requirements

1. Given an [8192 limit of deposits per block](https://github.com/berachain/beacon-kit/blob/main/consensus-types/types/deposit_request.go#L58), we need to guarantee that an [EL doesn't include more than 8192](https://github.com/ethereum/go-ethereum/blob/master/core/state_processor.go#L319) deposits in 1 block. 
    - This is guaranteed with the current gas limit of 30 million.
    - 1 deposit on Berachain costs at least 45,000 gas ([sample deposit](https://berascan.com/tx/0xb590bc542527bfb1fd0f4d902ca08248196c2a0facb9eb6d2af83750e9b394a2#eventlog))
    - 8192 deposits would cost 45,000 * 8192 = 368,640.000 gas
    - As long as gas per block stays under 368 million gas, we are good ✅ 
      - If we do increase gas, we can also increase the maximum amount of 8192 accordingly.

2. Given the way deposits are taken from EL deposit contract events and then processed in CL, _as is_ we will lose deposits from the 2 blocks right before the Fulu fork (which starts using 6110) is activated.
    - Currently the way the system works:
      - in finalize_block N, we store any new EL deposits from EL block N-1
      - in build_block N+1, we include these new EL deposits from EL block N-1
    - With 6110:
      - in build_block N+1, the EL will now include the deposits events that are in EL block N+1 itself
    - So if Fulu activates at block N+1 (previously for this block N+1 we would have included deposits from EL block N-1). Now with new fork, we will just include for block N+1 the deposits from EL block N+1. 
      - We added a special case around the fork activation that retrieves the previous EL blocks' deposits to specifically include them ✅ 
        - in block N's finalizeBlock we retrieve N-1's deposits (in postFinalizeBlockOps which is the standard procedure before Fulu)
        - in block N+1's building or verifying procedures, we catchup and retrieve block N's deposits
        - in verifying block N+1, there is no deposit limit per block as to exhaust any outstanding deposits still in the queue. Note block N+1 will also include the 6110 deposit requests for N+1's deposits, which by itself follows the 8192 deposit limit.
      
3. Forking on the EL to change the deposit event signature to Berachain's: `0x68af751683498a9f9be59fe8b0d52a64dd155255d85cdb29fea30b1e3f891d46` ([sample deposit](https://berascan.com/tx/0xb590bc542527bfb1fd0f4d902ca08248196c2a0facb9eb6d2af83750e9b394a2#eventlog))
    - Change(s) in In reth [here](https://github.com/berachain/bera-reth/pull/226) ✅ 
    
### TODOs:
- [x] Catchup deposits in 1 of process proposal / finalize block
- [x] write validation logic in processOperations for first block of Electra2
- [x] Electra2 fork test
- [x] Unit tests for deposit requests.
- [x] Simulation tests for deposit requests.

### Remaining Testing 
On a live devnet/testnet, ensure that timeouts do not occur when building first Fulu block as the build and verification procedures for this block now include an extra JSON RPC call to the EL. **Look for PrepareProposal/ProcessProposal/FinalizeBlock timeouts.**


state-transition/core/state_processor_staking.go
```diff
@@ -37,50 +37,63 @@ func (sp *StateProcessor) processOperations(
...
+			return withdrawErr
 		}
 	}
 
-	if version.EqualsOrIsAfter(blk.GetForkVersion(), version.Electra()) {
-		// After Electra, validators can request withdrawals through execution requests which must be handled.
-		requests, err := blk.GetBody().GetExecutionRequests()
-		if err != nil {
+	// EIP-6110 Deposits.
+	for _, dep := range requests.Deposits {
```
calbera, line 90: the biggest change is here: Post this fork, we will just accept the deposits from here as we have instant finality. No deposits queue or epoch processing is necessary here; real-time instant deposits very easily done

state-transition/core/state_processor_forks.go
```diff
@@ -94,6 +94,15 @@ func (sp *StateProcessor) ProcessFork(
...
 		}
+	case version.Electra1():
+		if err = sp.upgradeToElectra1(st, stateFork, slot); err != nil {
+			return err
+		}
+
+		// Log the upgrade to Electra1 if requested.
+		if logUpgrade {
+			sp.logElectra1Fork(stateFork.PreviousVersion, timestamp, slot)
+		}
```
rezbera, line 105: This will likely break if electra1 is set at genesis because `upgradeToElectra` will be skipped, which is required for this line

```go
	if err := st.SetPendingPartialWithdrawals([]*types.PendingPartialWithdrawal{}); err != nil {
		return err
	}
```

rezbera:
Supportive of this approach. Especially since its <200 line diff and gets rid of the CL querying the EL which isn't a great pattern.

Agreed that the Block Gas Limit is a good enough protection from hitting the deposit queue. It's unlikely we'll be x10'ing the gas limit arbitrarily in the coming months as that's not a bottleneck.

As you noted, the EL change is trivial. 

There is a point to be noted that we deviate from the spec by removing the deposit queue altogether, which is present in the spec. In the spec, deposit processing is done per epoch, whereas we do it per block. This allows us to avoid any BeaconState changes.  IMO the processing overhead of deposits is low enough that it doesn't need to be batched into an epoch (don't have empirical data to support this tho).

Big fan

calbera:
> There is a point to be noted that we deviate from the spec by removing the deposit queue altogether, which is present in the spec. In the spec, deposit processing is done per epoch, whereas we do it per block. This allows us to avoid any BeaconState changes. IMO the processing overhead of deposits is low enough that it doesn't need to be batched into an epoch (don't have empirical data to support this tho).

It does "deviate from spec" but not any more than we already are "deviated from the spec". Currently also deposits are processed block by block and the state transition operation for each is not too expensive (signatures are not verified on repeat -- most -- deposits). This change only affects the "lag" at which deposits are processed but not the frequency of processing itself.

beacon/blockchain/process_proposal.go
```diff
@@ -364,6 +364,14 @@ func (s *Service) VerifyIncomingBlock(
 		go s.handleOptimisticPayloadBuild(ctx, nextBlockData)
 	}
 
+	// if version.Equals(
```
calbera: add condition: and this block's timestamp == version.Electra1

which enforces that we only do this once, on the transition from electra1 -> electra2

beacon/blockchain/process_proposal.go
```diff
@@ -364,6 +364,14 @@ func (s *Service) VerifyIncomingBlock(
...
+	// 	s.chainSpec.ActiveForkVersionForTimestamp(nextBlockData.Timestamp), 
+	// 	version.Electra2(),
+	// ) {
+	// 	// If the next block is on the Electra2 fork, we need to fetch the deposits from the EL.
+	// 	s.depositFetcher(ctx, nextBlockData.Number)
+	// }
+
```
calbera: then add check for verifying the first block of electra2 that all the deposits are included. if >16 still in queue they should be checked and included (`if depositStore has deposits in queue then process them separately`)

state-transition/core/state_processor_forks.go
```diff
@@ -268,3 +279,55 @@ func (sp *StateProcessor) logElectra1Fork(
...
+// logFuluFork logs information about the Fulu fork.
+func (sp *StateProcessor) logFuluFork(
+	previousVersion common.Version, timestamp math.U64, slot math.Slot,
+) {
+	sp.logger.Info(fmt.Sprintf(`
+
+
+	⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️⏭️
+
+	+ ✅  welcome to the Fulu (0x05020000) fork! 🎉
```
bar-bera: nit:
```suggestion
	+ ✅  welcome to the Fulu (0x06000000) fork! 🎉
```

beacon/blockchain/service.go
```diff
@@ -105,10 +97,7 @@ func (s *Service) Name() string {
 }
 
 // Start starts the blockchain service.
-func (s *Service) Start(ctx context.Context) error {
-	// Catchup deposits for failed blocks. TODO: remove.
-	go s.depositCatchupFetcher(ctx)
```
bar-bera, line 110: my curiosity, why was this TODO there? I mean retrying for failing blocks seems a nice to have to avoid a validator being stuck on errors on `ReadDeposits`

state-transition/core/state_processor_forks.go
```diff
@@ -268,3 +279,55 @@ func (sp *StateProcessor) logElectra1Fork(
 		sp.cs.SlotToEpoch(slot).Unwrap(),
 	))
 }
+
+// upgradeToFulu upgrades the state to the Fulu fork version. It is modified from the ETH
+// 2.0 spec (https://ethereum.github.io/consensus-specs/specs/electra/fork/#upgrading-the-state) to:
+//   - update the Fork struct in the BeaconState
+//   - initialize the pending partial withdrawals to an empty array (if not already initialized)
+//   - catchup deposits from the EL to migrate to EIP-6110 style deposit requests
```
bar-bera:
```suggestion
```

beacon/deposits/deposits.go
```diff
@@ -0,0 +1,191 @@
...
+			"depositFetcher, nothing to fetch",
+			"block num", blockNum,
+			"eth1FollowDistance", eth1FollowDistance,
+		)
+		return
+	}
+	blockToFetch := blockNum - eth1FollowDistance
+	deposits, err := depositContract.ReadDeposits(ctx, blockToFetch)
+	if err != nil {
+		logger.Error("Failed to read deposits", "block", blockNum, "error", err)
```
bar-bera, line 116: here could early return since deposits would be nil in case (and would otherwise log misleading messages like `nothing to fetch`)

beacon/deposits/deposits.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil
+	}
+
+	deposits, err := depositContract.ReadDeposits(ctx, lph.GetNumber())
+	if err != nil {
+		return err
+	}
+	if len(deposits) == 0 {
+		logger.Info("Deposits catchup for Fulu, nothing to fetch")
+		return nil
```
bar-bera: minor: `LastBlockNumber` guard is not set here. not a deal to re-do the deposit fetch a bunch of times since it's just happening in the fork block, but still

beacon/blockchain/service.go
calbera: The TODO is there since catching up deposits is useless. By design, if you miss deposits of a block, you will fail to verify/process/finalize that block and your node will get stuck. So catching up at a later time is useless. Retrying then and there is the only way

beacon/deposits/deposits.go
calbera: We only set last block number if the deposits are successfully stored into the deposit store, which is by design. If the deposits are successfully read but not stored then we do want to retry. 

The main purpose of this guard is to prevent the fetch from happening twice. i.e. nodes who are keeping up with the chain will run this code in prepare/process proposal. Then they shouldn't re-fetch it in finalize block. However nodes that are catching up/syncing will not run prepare/process proposal so they need to run this code in finalizeblock.

bar-bera: cmiiw, considering the happy path, proposer correctly read deposits in `PrepareProposal`, but there aren't any (most probably), then `len == 0` -> do not set the guard and fetch again in the next phases

calbera: I see your point now, gotcha. fixed [here](https://github.com/berachain/beacon-kit/pull/2794/commits/40b66d5a35f9eca5a05ae46d397a2bdc1b918ad7)

beacon/deposits/deposits.go
```diff
@@ -0,0 +1,195 @@
...
+			"depositFetcher, nothing to fetch",
+			"block num", blockNum,
+			"eth1FollowDistance", eth1FollowDistance,
+		)
+		return
+	}
+	blockToFetch := blockNum - eth1FollowDistance
+	deposits, err := depositContract.ReadDeposits(ctx, blockToFetch)
+	if err != nil {
+		logger.Error("Failed to read deposits", "block", blockNum, "error", err)
```
fridrik01, line 116: is it safe to drop the error and only log, we had a retry mechanism before in depositCatchupFetcher?

execution/deposit/contract.go
```diff
@@ -38,6 +39,10 @@ import (
 type WrappedDepositContract struct {
 	// DepositContractFilterer is a pointer to the codegen ABI binding.
 	deposit.DepositContractFilterer
+
+	// lastBlockNumber is the last block number that was successfully consumed from
+	// the deposit contract.
+	lastBlockNumber atomic.Pointer[math.U64]
```
fridrik01: nit: use `atomic.Uint64`

beacon/deposits/deposits.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	prevBlockForkVersion := chainSpec.ActiveForkVersionForTimestamp(lph.GetTimestamp())
+	isFirstFuluBlock := version.Equals(prevBlockForkVersion, version.Electra1()) &&
+		version.Equals(blk.GetForkVersion(), version.Fulu())
+	if !isFirstFuluBlock {
+		return nil
+	}
+
+	// If we already fetched deposits for this block, we don't need to do it again.
+	if depositContract.LastBlockNumber() == lph.GetNumber() {
```
fridrik01, line 63: is this to guard against calling this several times from different places if isFirstFuluBlock is true?

fridrik01 (commented):
did a quick pass, overall looks good but will need to do another

fridrik01:
@calbera we should test this soon on devnet with load testing deposits

beacon/deposits/deposits.go
calbera: yes, this doesnt change any behavior. That retry mechanism was non-effectual, long overdue cleaning of tech debt

beacon/deposits/deposits.go
calbera: Yes this guards only to prevent catchup. We will return from this function early if we are not in the first fulu block. If we are in first fulu block we need to run this fetch once amongst prepare/processProposal and finalizeBlock

execution/deposit/contract.go
```diff
@@ -60,17 +65,22 @@ func NewWrappedDepositContract(
 	}, nil
 }
 
+// LastBlockNumber returns the last block number that was successfully
+// consumed from the deposit contract.
+func (dc *WrappedDepositContract) LastBlockNumber() math.U64 {
+	return math.U64(dc.lastBlockNumber.Load())
```
fridrik01, line 71: @calbera something I noticed: `lastBlockNumber` zero-value doesn't distinguish "fetched at block 0" from "never fetched." Should be harmless in practice since IIUC genesis deposits bypass contract events, but wanted to call it out anyway.

primitives/version/versions.go
```diff
@@ -44,8 +44,8 @@ var (
 	electra = common.Version{0x05, 0x00, 0x00, 0x00}
 	// electra1 is the first hardfork of Electra on Berachain mainnet.
 	electra1 = common.Version{0x05, 0x01, 0x00, 0x00}
-	// TBD if used but kept as an example.
-	electra2 = common.Version{0x05, 0x02, 0x00, 0x00}
+	// fulu is the first version of the Fulu hardfork on Berachain mainnet.
```
fridrik01, line 47: Should we not add `fulu` into the `supportedVersions` in `primitives/version/supported.go` so `GetSupportedVersions` will include fulu?

chain/helpers_test.go
```diff
@@ -38,6 +38,7 @@ var spec, _ = chain.NewSpec(
 		Deneb1ForkTime:                   9 * 32 * 2,
 		ElectraForkTime:                  10 * 32 * 2,
 		Electra1ForkTime:                 11 * 32 * 2,
+		FuluForkTime:                     12 * 32 * 2,
```
fridrik01, line 41: Should we not add Fulu in TestActiveForkVersionForTimestamp?

beacon/deposits/deposits.go
```diff
@@ -0,0 +1,195 @@
...
+	"github.com/berachain/beacon-kit/primitives/constants"
+	"github.com/berachain/beacon-kit/primitives/math"
+	"github.com/berachain/beacon-kit/primitives/version"
+	statedb "github.com/berachain/beacon-kit/state-transition/core/state"
+	depositstore "github.com/berachain/beacon-kit/storage/deposit"
+)
+
+// If on the first block of Fulu, catchup the previous block's deposits. Between
+// Prepare/ProcessProposal and FinalizeBlock, this only needs to be done once.
+func CatchupFuluDeposits(
```
fridrik01, line 41: nit: unit tests for these would be nice (though they should be  covered in simulation/integration tests)

fridrik01 (approved):
mostly just nits

beacon/deposits/deposits.go
calbera: I didn't add unit tests for this file since most of the logic already existed in our processproposal / finalizeblock flows, so refactoring it to here should change no behavior, which can be enforced by syncing mainnet, and other e2e tests

execution/deposit/contract.go
calbera: yeah thats true, but for the purpose of deposit contract, its functionally equivalent to say that we "never fetched" deposits if the last block fetched was 0 since 0 block will not have any deposits. We start at 1 onwards.

PR #2795, opened 2025-05-29
dockerfile for e2e framework

greg-szabo:
Create a dockerfile that works with the e2e framework.

This is an override of the beacond Dockerfile that sets up some configuration for the e2e framework. It does not build baecond by itself. It has minor modifications to `entrypoint.sh` so it can run in an automated environment. The changes are backwards-compatible.


greg-szabo:
Clsong this. I will open a PR once all the features are together for a better e2e testing experience.

PR #2808, opened 2025-06-11
[DO NOT MERGE] chore(proofs): Merkle G-Indexes for Validator Withdrawal Credentials

calbera:


node-api/handlers/proof/merkle/testdata/many_validators_validator_credentials_proof.json
calbera, line 1: For this proof:
- leaf [validatorCredentials.Root()]: `0x0100000000000000000000000102030000000000000000000000000000000000`
- validatorIndex: `95` -> GIndex: `95 * 8 + 6350779162034177` => `6350779162034937`
- root [beaconBlockRoot]: `0xd661072c9985c0191fe24dc73241aa76328816d234e50229c5d5ef46dab83848`

node-api/handlers/proof/merkle/validator_credentials.go
```diff
@@ -0,0 +1,150 @@
+package merkle
```
rezzmah: nit : package below license

rezbera (approved):
Looks reasonable assuming you've tested this and it meets the requirement. 

My only concern is around the fact that multiple val pubkeys/indexes can have the same withdrawal credential. What impact does that have on any guarantees the smart contract is trying to enforce?

calbera:
> My only concern is around the fact that multiple val pubkeys/indexes can have the same withdrawal credential. What impact does that have on any guarantees the smart contract is trying to enforce?

I believe the idea is that for a validator to join the staking pool (contract), it must verify that their credentials provided (to the contract call) matches the true credentials of the beacon chain. cc @bearmigiano

node-api/handlers/proof/merkle/testdata/many_validators_validator_credentials_proof.json
calbera: credentials address: `0x0102030000000000000000000000000000000000`

PR #2812, opened 2025-06-18
wip: no-op finalize block - 4

abi87:


consensus/cometbft/service/process_proposal.go
```diff
@@ -45,43 +45,45 @@ func (s *Service) processProposal(
...
 			"height", req.Height,
 			"time", req.Time,
 			"hash", fmt.Sprintf("%X", req.Hash),
 			"err", err,
 		)
+		return &cmtabci.ProcessProposalResponse{Status: status}, nil
 	}
+
+	if req.Height > s.initialHeight {
+		s.candidateStates[string(req.Hash)] = &CacheElement{
```
fridrik01: IIUC we don't delete from this cache for failed/rejected proposals

abi87: you just don't add. There is an early return if block verification fails right?
Much simpler design than attempt 3

fridrik01: I mean if there are multiple calls to ProcessProposal where we have several rounds, won't we end up just cleaning up a single entry in the cache (the block we finalize) leaving the cache items for previous rounds orphaned?

abi87: Properly cleaned up candidateStates cache in block commit. 
LMK if this answer your point.

beacon/blockchain/state_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/berachain/beacon-kit/config/spec"
+	"github.com/berachain/beacon-kit/node-core/components/metrics"
+	"github.com/berachain/beacon-kit/node-core/components/storage"
+	"github.com/berachain/beacon-kit/primitives/math"
+	statetransition "github.com/berachain/beacon-kit/testing/state-transition"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestStateCopyIsIndependentFromOriginal(t *testing.T) {
```
abi87: an attempt at understanding why `payload_test.go` are failing

beacon/blockchain/metrics.go
```diff
@@ -107,12 +107,3 @@ func (cm *chainMetrics) markOptimisticPayloadBuildFailure(
 		err.Error(),
 	)
 }
-
-// measureStateRootVerificationTime measures the time taken to verify the state
-// root of a block.
-// It records the duration from the provided start time to the current time.
-func (cm *chainMetrics) measureStateRootVerificationTime(start time.Time) {
```
abi87: metric used for ProcessProposal, replaced by its FinalizeBlock companion. TODO: consider activating this under fork

consensus/cometbft/service/finalize_block.go
```diff
@@ -48,52 +54,91 @@ func (s *Service) finalizeBlockInternal(
...
+			); err != nil {
+				return nil, fmt.Errorf("failed finalizing sidecars: %w", err)
+			}
+			if err = s.Blockchain.PostFinalizeBlockOps(
+				finalState.Context(),
+				blk,
+			); err != nil {
+				return nil, fmt.Errorf("finalize block: failed post finalize block ops: %w", err)
+			}
+			return formatFinalizeBlockResponse(cached.ValUpdates, len(req.Txs), s.cmtConsensusParams)
```
abi87: This is the fast track to finalize blocks already verified. We just store blobs/block, send FCU and pull deposits to be included in next block

consensus/cometbft/service/finalize_block.go
```diff
@@ -48,52 +54,91 @@ func (s *Service) finalizeBlockInternal(
...
-		}
+	// Block has not been cached already, as it happens when we block sync.
+	// Normally we would directly process it but first block post initial height
+	// needs special handling.
+	var finalState *cache.State
+	cachedFinalHash, cachedFinalState, err := s.cachedStates.GetFinal()
+	switch {
+	case errors.Is(err, cache.ErrNoFinalState):
+		hash = string(req.Hash) // not necessary but clarifies intent
+		finalState = s.resetState(ctx)
```
abi87, line 103: this is basically the block sync case now.

consensus/cometbft/service/finalize_block.go
```diff
@@ -48,52 +54,91 @@ func (s *Service) finalizeBlockInternal(
...
+
+	case err == nil:
+		hash = cachedFinalHash
+		finalState = cachedFinalState
+
+		// Preserve the CosmosSDK context while using the correct base ctx.
+		finalState.SetContext(finalState.Context().WithContext(ctx))
```
abi87, line 110: this is the special case when we finalize the first block. To simplify implementation I chose not to properly cache it but to execute it all in FinalizeBlock, similarly to what we do today.

consensus/cometbft/service/finalize_block.go
```diff
@@ -166,3 +212,37 @@ func (s *Service) validateFinalizeBlockHeight(
...
+	if err != nil {
+		return nil, err
+	}
+
+	cp := cmtConsensusParams.ToProto()
+	return &cmtabci.FinalizeBlockResponse{
+		TxResults:             txResults,
+		ValidatorUpdates:      formattedValUpdates,
+		ConsensusParamUpdates: &cp,
+	}, nil
```
abi87: just moved out formatting of response to its own method

beacon/blockchain/common.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"fmt"
+
+	ctypes "github.com/berachain/beacon-kit/consensus-types/types"
+	"github.com/berachain/beacon-kit/consensus/cometbft/service/encoding"
+	"github.com/berachain/beacon-kit/da/types"
+	"github.com/berachain/beacon-kit/primitives/math"
+)
+
+func (s *Service) ParseBeaconBlock(req encoding.ABCIRequest) (
```
abi87, line 32: With fast finalization due to state caching, we need to pull out and reuse specific pieces of code currently executed only in `FinalizeBlock`. This is one of them

abi87:
@coderabbitai review

consensus/cometbft/service/cache/cache.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/berachain/beacon-kit/primitives/math"
+	"github.com/berachain/beacon-kit/primitives/transition"
+)
+
+// TODO: Replace this with stable block time activation height to make sure
+// chains does not get faster before SBT is activated. Once activated, we
+// can drop the activation height and act as if the activation height is 2
+// (Can't be one!!!)
+const activationHeight = math.Slot(2)
```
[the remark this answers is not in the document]
abi87: Indeed, this will be done as soon as another PR (Stable block time) is merged in. Plan is to activate caching only after stable block time is activated.

consensus/cometbft/service/cache/cache.go
```diff
@@ -0,0 +1,112 @@
...
+func New() States {
+	return &candidateStates{
+		states:         make(map[string]*Element),
+		finalStateHash: nil,
+	}
+}
+
+func (cs *candidateStates) Cache(hash string, toCache *Element) {
+	cs.states[hash] = toCache
+}
```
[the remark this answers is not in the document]
abi87: They key used for states is the Comet block hash of the corresponding block and Comet should guarantee its uniqueness (i.e. that each block is submitted for verification only once).
We do want to overwrite state.

consensus/cometbft/service/cache/cache.go
```diff
@@ -0,0 +1,112 @@
...
+	if !found {
+		return "", nil, ErrFinalStateIsNil
+	}
+	return *cs.finalStateHash, cached.State, nil
+}
+
+func (cs *candidateStates) Reset() {
+	cs.states = make(map[string]*Element)
+	cs.finalStateHash = nil
+}
```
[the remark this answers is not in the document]
abi87: Cache is only used for CometBFT callbacks, which are guaranteed to be serialized, so we do not need locking. Do you agree?

beacon/blockchain/process_proposal.go
```diff
@@ -157,28 +131,30 @@ func (s *Service) ProcessProposal(
...
+			return nil, err
 		}
 	}
 
 	// Process the block.
+	s.logger.Debug( // needed for some checks in sim tests
+		"Processing block with fork version",
+		"block", req.Height,
+		"fork", blkVersion.String(),
+	)
```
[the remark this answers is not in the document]
abi87: What would test-specific logging look like? Isn't it just logging?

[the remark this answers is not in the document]
abi87: I think this is fine. Some of our tests to assert on logs content and this PR moved the asserts from `FinalizeBlock` to `ProcessProposal`. The comment is just meant to clarify that missing these logs would make those tests fail.

beacon/blockchain/finalize_block.go
```diff
@@ -69,45 +57,67 @@ func (s *Service) FinalizeBlock(
...
 		)
 	}
+	return nil
+}
 
-	// STEP 3: Finalize the block.
-	consensusBlk := types.NewConsensusBlock(blk, req.GetProposerAddress(), req.GetTime())
+func (s *Service) PostFinalizeBlockOps(ctx sdk.Context, blk *ctypes.BeaconBlock) error {
+	// TODO ABENEGIA: verify whether this is correct in the face of the extra copies
 	st := s.storageBackend.StateFromContext(ctx)
```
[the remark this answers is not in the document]
abi87: Indeed, I will extract the data needed here as we process the block and avoid passing state around

beacon/blockchain/finalize_block.go
```diff
@@ -116,7 +116,7 @@ func (s *Service) FinalizeSidecars(
 }
 
 func (s *Service) PostFinalizeBlockOps(ctx sdk.Context, blk *ctypes.BeaconBlock) error {
-	// TODO ABENEGIA: verify whether this is correct in the face of the extra copies
+	// TODO: consider extracting LatestExecutionPayloadHeader instead of using state here
```
abi87, line 119: I don't like regenerating the state and I could cache lph way earlier and pass it down here. However I am not sure this is more efficient.
On the other hand, we are only reading the state here so I believe all is correct. State is not accessed concurrently, as it happens for optimistic payload building. Instead these operations are done sequentially

consensus/cometbft/service/finalize_block.go
```diff
@@ -50,57 +55,91 @@ func (s *Service) finalizeBlockInternal(
...
-		s.finalizeBlockState = s.resetState(ctx)
-	} else {
-		// Preserve the CosmosSDK context while using the correct base ctx.
-		s.finalizeBlockState.SetContext(s.finalizeBlockState.Context().WithContext(ctx))
-	}
+	// Check whether currently block hash is already available. If so
+	// we may speed up block finalization.
+	hash := string(req.Hash)
+	switch cached, err := s.cachedStates.GetCached(hash); {
+	case err == nil:
```
fridrik01, line 50: I find normal if check more readable than using switch/case here (ie. always treat error case as indented)

consensus/cometbft/service/cache/cache.go
```diff
@@ -0,0 +1,113 @@
...
+
+	ErrStateNotFound          = errors.New("state not found")
+	ErrNoFinalState           = errors.New("no state marked as final")
+	ErrFinalStateIsNil        = errors.New("state marked as final is nil")
+	ErrFinalizingUnknownState = errors.New("attempt at finalizing unknown state")
+)
+
+type States interface {
+	Cache(hash string, toCache *Element)
+	GetCached(hash string) (*Element, error)
```
fridrik01: Consider renaming to `Get` and `Set` maybe?

consensus/cometbft/service/cache/cache.go
```diff
@@ -0,0 +1,113 @@
...
+var (
+	_ States = (*candidateStates)(nil)
+
+	ErrStateNotFound          = errors.New("state not found")
+	ErrNoFinalState           = errors.New("no state marked as final")
+	ErrFinalStateIsNil        = errors.New("state marked as final is nil")
+	ErrFinalizingUnknownState = errors.New("attempt at finalizing unknown state")
+)
+
+type States interface {
```
fridrik01, line 54: Needs some comment for readability around how this is used, especially around the MarkasFinal/GetFinal and that Reset is called after each commit, etc

consensus/cometbft/service/finalize_block.go
```diff
@@ -50,57 +55,91 @@ func (s *Service) finalizeBlockInternal(
...
-	)
-	if err != nil {
-		return nil, err
+	// Block has not been cached already, as it happens when we block sync.
+	// Normally we would directly process it but first block post initial height
+	// needs special handling.
+	var finalState *cache.State
+	cachedFinalHash, cachedFinalState, err := s.cachedStates.GetFinal()
+	switch {
+	case errors.Is(err, cache.ErrNoFinalState):
```
fridrik01, line 101: nit: same here with switch/case, I would treat error case as indented, now happy with sits between two error cases

consensus/cometbft/service/service.go
```diff
@@ -75,23 +76,10 @@ type Service struct {
...
-
-	// finalizeBlockState is used for FinalizeBlock, which is set based on the
-	// previous block's state. This state is committed. finalizeBlockState is
-	// set
-	// on InitChain and FinalizeBlock and set to nil on Commit.
-	finalizeBlockState *state
+	// cachedStates tracks in memory the post block states
+	// of blocks which were successfully verified. It allows
+	// finalizing without re-execution
+	cachedStates cache.States
```
fridrik01, line 82: nice, combining these three into one!

fridrik01 (commented):
first pass

consensus/cometbft/service/finalize_block.go
abi87: Can you suggest an if based implementation here? Not against an if implementation but I don't see how it is a pile of three if stack on top of another, which is effectively the same?

consensus/cometbft/service/service.go
abi87: To be precise, process and finalize block states are combined together in caches. Prepare proposal state just does not need to be an attribute of service: it is still an ephemeral state that is dropped once block is done being built.
The reason that state is not cached is that we carry out only partial verification on it when building.
We could consider doing the full verification and cache it right away: however I believe this would  helpl only the block builder. Any other node would receive the block slightly later (due to the time it takes to carry out the extra checks). So overall I chose to keep the state ephemeral.
Maybe I should write this comment in the codebase. Not sure if you agree with the decision

consensus/cometbft/service/finalize_block.go
fridrik01: Sure, with help of claude fired up this version -> https://gist.github.com/fridrik01/221876b2d93ce084c5495683949391eb. That said, not a hill I’m gonna die on hehe.

consensus/cometbft/service/cache/cache.go
abi87: Added some comments, LMK if they are clear/sufficient

consensus/cometbft/service/finalize_block.go
abi87: I see now. The nested if to handle the error is indeed not too bad. I am only mildly annoyed by the inverted logic for the `errors.Is(err, cache.ErrStateNotFound)` test. Also you don't really reduce indentation I think, because you gotta guard the fast finalization with `if err == nil`
I will let majority of reviewers pick their favourite version here

abi87: I left the switch structure and encapsulated the cached state handling in its own method `tryFinalizingCachedState`. Not sure if it is any better

consensus/cometbft/service/finalize_block.go
```diff
@@ -34,73 +37,99 @@ func (s *Service) finalizeBlock(
 	ctx context.Context,
 	req *cmtabci.FinalizeBlockRequest,
 ) (*cmtabci.FinalizeBlockResponse, error) {
-	res, err := s.finalizeBlockInternal(ctx, req)
-	if res != nil {
-		res.AppHash = s.workingHash()
```
abi87, line 39: just moved this to `formatFinalizeBlockResponse` method and drop the need of a `finalizeBlockInternal`

consensus/cometbft/service/finalize_block.go
fridrik01: not a big deal honestly, fine with keeping it as is for sake of getting this merged

consensus/cometbft/service/finalize_block.go
```diff
@@ -22,85 +22,112 @@ package cometbft
...
-	}
+	// Check whether currently block hash is already available. If so
+	// we may speed up block finalization.
+	hash := string(req.Hash)
+	switch cached, err := s.cachedStates.GetCached(hash); {
+	case err == nil:
+		// Block with height equal to initial height is special and we can't rely on its cache.
+		// This is because Genesis state is cached but not committed (and purged from s.cachedStates)
+		// We handle the case below.
+		if req.Height > s.initialHeight {
```
fridrik01, line 54: Shouldn't this be `if req.Height == s.initialHeight` ?

abi87: Nope. The special handling for `req.Height == s.initialHeight` is done below and it is similar to what we do, e.g. when we block sync: create a new state, do the full block execution and commit.
Here we handle caches states.
The pickle is that we have cached the state at s.initialHeight (and declared it final too), BUT we must not process it here

fridrik01: oh wait, this is the happy path and the special case is handled below not inside this if statement

fridrik01: Can you maybe update the comment to be explicit that this checks for the happy path and the special case is handled below (I know the comment says `We handle the case below` but I thought it mentioned the if statement)

abi87: Done in 8a6437bf4

consensus/cometbft/service/finalize_block.go
```diff
@@ -217,3 +243,45 @@ func (s *Service) validateFinalizeBlockHeight(
...
+
+func (s *Service) formatFinalizeBlockResponse(
+	req *cmtabci.FinalizeBlockRequest,
+	valUpdates transition.ValidatorUpdates,
+) (*cmtabci.FinalizeBlockResponse, error) {
+	// Update Stable block time related data
+	if s.cmtConsensusParams.Feature.SBTEnableHeight == 0 && req.Height == s.delayCfg.SbtConsensusUpdateHeight() {
+		s.cmtConsensusParams.Feature.SBTEnableHeight = s.delayCfg.SbtConsensusEnableHeight()
+	}
+	nextBlockTime := s.nextBlockDelay(req)
```
fridrik01, line 255: nit: updating SBT in a format function seems a bit out of place

consensus/cometbft/service/process_proposal.go
```diff
@@ -45,43 +47,52 @@ func (s *Service) processProposal(
...
 	)
 
 	// errors to consensus indicate that the node was not able to understand
 	// whether the block was valid or not. Viceversa, we signal that a block
 	// is invalid by its status, but we do return nil error in such a case.
-	status := cmtabci.PROCESS_PROPOSAL_STATUS_ACCEPT
-	err := s.Blockchain.ProcessProposal(
-		s.processProposalState.Context(),
+	valUpdates, err := s.Blockchain.ProcessProposal(
+		processProposalState.Context(),
```
fridrik01, line 68: this was a bit confusing to me as we pass in the context here and not the state, but inside we fetch the state from the context

consensus/cometbft/service/encoding/interfaces.go
```diff
@@ -20,8 +20,11 @@
 
 package encoding
 
+import "time"
+
 // ABCIRequest represents the interface for an ABCI request.
 type ABCIRequest interface {
 	// GetTxs returns the transactions included in the request.
 	GetTxs() [][]byte
+	GetTime() time.Time
```
fridrik01, line 29: super nit: missing comment

consensus/cometbft/service/finalize_block.go
```diff
@@ -22,85 +22,112 @@ package cometbft
...
+				blk,
+			); err != nil {
+				return nil, fmt.Errorf("finalize block: failed post finalize block ops: %w", err)
+			}
+			return s.formatFinalizeBlockResponse(req, cached.ValUpdates)
 		}
+
+	case errors.Is(err, cache.ErrStateNotFound):
+		// this is a benign error, it just signal it's the first time we see the block being finalized
+		// Keep processing below
```
fridrik01, line 89: so when catching up during block sync we always end up here

consensus/cometbft/service/finalize_block.go
```diff
@@ -22,85 +22,112 @@ package cometbft
...
+		return nil, fmt.Errorf("failed checking cached state, hash %s, height %d: %w", hash, req.Height, err)
 	}
 
-	finalizeBlock, err := s.Blockchain.FinalizeBlock(
-		s.finalizeBlockState.Context(),
-		req,
-	)
-	if err != nil {
-		return nil, err
+	// Block has not been cached already, as it happens when we block sync.
```
fridrik01, line 95: and if we are here, that means we are either:
- in the special case where `req.Height == s.initialHeight` represented by the `err == nil` switch case below , or
-  in a block sync running finalize block on blocks already achieving consensus (so no processpropsal called allowing it to be cached) represented by the  `errors.Is(err, cache.ErrNoFinalState)` switch case below

consensus/cometbft/service/finalize_block.go
abi87: I agree but it needs to be there as there are two code path leading to a response (cached and non-cached) and I really want to avoid to exporting these computations outside of this function and duplicate code.
I think a better alternative is to rename this to `calculateFinalizeBlockResponse`, which should tell that we don't just format a response but complete some operations before responding back to consensus

consensus/cometbft/service/finalize_block.go
abi87: Yes! I believe it is currently the only case where this happens, but cannot guarantee it would stay the same (e.g. if we introduce state sync)

PR #2819, opened 2025-06-23
chore(payload): do not query state when building RequestPayload

abi87:
This PR simplifies and fixes some code path related to block building, which are prerequisites for block state caching in https://github.com/berachain/beacon-kit/pull/2812. It does the following:

- It stops blockBuilder from accessing blockchain state when building the next block payload. There are cases where next block payloads are built concurrently and access to blockchain state is not safe for concurrent access.
- It extend `payload_test.go` tests, adding checks on state changes performed by the block being verified
- It execute the block being verified on the state provided by consensus rather than on its copy (currently called postState). This is necessary in anticipation for caching the state
- It takes care on consolidating fetching of data to optimistically build next block payload into a single function
- It takes care on consolidating protecting the block state (to be cached) from any changed performed to collect data for next block payload

Special thanks to @calbera for the help in squashing the bugs I introduced while coding.
I have verified that all test pass when optimistic payload building is activated in tests (then revert them back to not using optimistic payload as we currently do in main)


beacon/blockchain/process_proposal.go
```diff
@@ -280,10 +270,19 @@ func (s *Service) VerifyIncomingBlock(
...
+		if err != nil {
+			return fmt.Errorf("failed preFetching data for rejection: %w", err)
+		}
+	}
+
 	// Verify the state root of the incoming block.
 	err = s.verifyStateRoot(
 		ctx,
-		postState,
+		state,
```
calbera, line 300: q: why shouldn't this be a copy of the state? i.e. if we end up rejecting the block here (verifyStateRoot returns non-nil err) then we dont want to persist any changes on state itself?

calbera: Or does writing to this ctx not matter as it never gets persisted anyhow?

abi87: The important point is:
- changes carried out on `state` will be visible from `copiedState` but not viceversa.

So, as you said, before this PR we used copiedState to make sure that whatever partial change an invalid block would carried out won't pollute `state` which was used to optimistically request a new block.

With my changes however, this does not matter as we pre-fetch all the relevant data.
Also in upcoming state caching we do need to carry out the block changes on `state` in order to cache it and not on `copiedState`. So I took advantage of this PR to do that.

This is a subtle but important point. LMK if this makes sense to you

beacon/blockchain/payload_test.go
```diff
@@ -100,46 +101,36 @@ func TestOptimisticBlockBuildingRejectedBlockStateChecks(t *testing.T) {
...
+			require.Empty(t, r.FinalEth1BlockHash)          // this is first block post genesis
+			require.Equal(t, constants.GenesisSlot, r.Slot) // genesis slot in state
 		},
 	).Return(nil, common.Version{0xff}, errors.New("does not matter")) // return values do not really matter in this test
 	wg.Add(1)
 
+	// check slot pre test
+	slot, err := st.GetSlot()
+	require.NoError(t, err)
+	require.Equal(t, constants.GenesisSlot, slot)
```
abi87, line 132: these checks are not in `main` because there we don't cache state and we don't really care about state pollution.
In preparation of state caching we are stricter in the kind of invariants that optimistic block building should provide

beacon/blockchain/payload_test.go
```diff
@@ -220,43 +212,32 @@ func TestOptimisticBlockBuildingVerifiedBlockStateChecks(t *testing.T) {
...
-			require.NoError(t, err)
-			require.Equal(t, validBlk.Slot+1, stateSlot)
-			require.Equal(t, slot, stateSlot)
+			require.Equal(t, validBlk.HashTreeRoot(), r.ParentBlockRoot)
+			require.Equal(t, validBlk.Slot+1, r.Slot)
 		},
 	).Return(nil, common.Version{0xff}, errors.New("does not matter")) // return values do not really matter in this test
 	wg.Add(1)
 
+	// check slot pre test
```
abi87, line 236: these checks are not in `main` because there we don't cache state and we don't really care about state pollution.
In preparation of state caching we are stricter in the kind of invariants that optimistic block building should provide

beacon/blockchain/payload.go
```diff
@@ -125,167 +127,111 @@ func (s *Service) forceSyncUponFinalize(
 	}
 }
 
-// handleRebuildPayloadForRejectedBlock handles the case where the incoming
-// block was rejected and we need to rebuild the payload for the current slot.
-func (s *Service) handleRebuildPayloadForRejectedBlock(
-	ctx context.Context,
+func (s *Service) preFetchBuildData(
```
abi87: once you provide the right state, we really need to carry out the very same operations to extract the data necessary to build the next block, whether current block is being rejected or accepted.
This is way there can be (and so should be) a single function doing these ops.

beacon/blockchain/payload.go
```diff
@@ -125,167 +127,111 @@ func (s *Service) forceSyncUponFinalize(
...
 
 	stateSlot, err := st.GetSlot()
 	if err != nil {
-		return err
+		return nil, fmt.Errorf("failed retrieving slot from state: %w", err)
 	}
+	blkSlot := stateSlot + 1
 
-	// Set the previous state root on the header.
-	latestHeader.SetStateRoot(st.HashTreeRoot())
```
abi87, line 176: this really mimicks `ProcessSlot` so let's just do `ProcessSlot` (which is conceptually correct)

beacon/blockchain/payload_test.go
```diff
@@ -100,46 +101,36 @@ func TestOptimisticBlockBuildingRejectedBlockStateChecks(t *testing.T) {
...
+			require.Equal(t, expectedParentBlockRoot, r.ParentBlockRoot)
 
-			require.Empty(t, finalEth1BlockHash)          // this is first block post genesis
-			require.Equal(t, constants.GenesisSlot, slot) // genesis slot in state
-			var stateSlot math.Slot
-			stateSlot, err = st.GetSlot()
-			require.NoError(t, err)
-			require.Equal(t, constants.GenesisSlot, stateSlot)
+			require.Empty(t, r.FinalEth1BlockHash)            // this is first block post genesis
+			require.Equal(t, constants.GenesisSlot+1, r.Slot) // rebuild block on top of genesis
```
abi87, line 124: This is actually a change wrt to `main` but I believe `main` is wrong.
The scenario is:
- a block `B_0` of height `h` is already final (so in `state`)
- a block `B_1` of height `h+1` is being verified and it is rejected.

What should be the height of the next block being built optimistically?
We should build a sibling of `B+1` so the requested height should be `B_0` height +1.
This is what it is asserted in this test (which does not happen in prod)

beacon/validator/config.go
```diff
@@ -26,7 +26,7 @@ const (
 
 	// defaultEnableOptimisticPayloadBuilds is the default
 	// for enabling the optimistic payload builder.
-	defaultEnableOptimisticPayloadBuilds = false
+	defaultEnableOptimisticPayloadBuilds = true
```
abi87: TMP: to be removed, with the rest of changes in commit 517878f007e4635130190db44fd2ad09c34959d4

beacon/blockchain/payload.go
calbera: can you add this comment to the function doc. Makes sense to me now but will be helpful in the code.

calbera: and maybe remove "pre"? Feels confusing to me since we absolutely must fetch the payload data before requesting to build another payload

beacon/blockchain/process_proposal.go
```diff
@@ -280,10 +271,20 @@ func (s *Service) VerifyIncomingBlock(
...
 
+	var nextBlockData *builder.RequestPayloadData
+	if s.shouldBuildOptimisticPayloads() {
+		// state copy makes sure that preFetchBuildDataForSuccess does not affect state
+		copiedState := state.Copy(ctx)
+		nextBlockData, err = s.preFetchBuildData(copiedState, consensusTime)
+		if err != nil {
+			return fmt.Errorf("failed preFetching data for rejection: %w", err)
+		}
+	}
```
calbera: I dont get why this branch has to be here. i.e. we can just do this copy and fetching of build data in the case we really need to? if the changes to `state` (via `verifyStateRoot` below) will persist on this `copiedState`, might as well just make the copy only when its really needed, i.e. `err` of `verifyStateRoot` is non-nil.

calbera: Also I feel it doesnt make sense that we reject the proposal (i.e. returning error on line280) if our fetch for buildData fails, without even running `verifyStateRoot` on the existing `beaconBlk` proposal

beacon/blockchain/process_proposal.go
```diff
@@ -295,52 +296,29 @@ func (s *Service) VerifyIncomingBlock(
...
 	if s.shouldBuildOptimisticPayloads() {
-		lph, lphErr := postState.GetLatestExecutionPayloadHeader()
-		if lphErr != nil {
-			return fmt.Errorf("failed loading LatestExecutionPayloadHeader: %w", lphErr)
+		// state copy makes sure that preFetchBuildDataForSuccess does not affect state
+		copiedState := state.Copy(ctx)
+		nextBlockData, err = s.preFetchBuildData(copiedState, consensusTime)
+		if err != nil {
+			return fmt.Errorf("failed preFetching data for success: %w", err)
 		}
```
calbera: might be a silly question -- In this case (already verified the proposal is good), why is a state copy needed here? Are we planning on re-using this `processProposalState` (from [here](https://github.com/berachain/beacon-kit/blob/f4bf85a49e347690b0e02648045c60ab18da5994/consensus/cometbft/service/process_proposal.go#L73)) anywhere else?

abi87: Not silly at all. In next PR I will cache the state AND `preFetchBuildData` does modify the state (via `ProcessSlots` and `ProcessFork`). So I really think the copy is needed (and if you avoid that you should see one of the `payload_test.go` UTs failing on the post execution checks, showing the wrong slot.

On the brighter side, the copies now are only done with optimistic payload building is active (currently you make a copy even if it s not); also only a single copy is done (the block is either accepted or rejected but not both).

Thanks for double checking this, please LMK if this makes sense to you

beacon/blockchain/process_proposal.go
abi87: As soon as you start processing the block, the associated state can be modified. Even if the block is rejected, the state can be partially written already and so we can't use that state anymore.
This is why we really need to fetch the data before processing the block.
Still why the copy? Unfortunately `preFetchBuildData` does modify the state as well (via `ProcessSlot` and `ProcessFork`) so I need to protect the `state` used to verify the block from the `copiedState` used to load data for next block.

We could imagine to carry out `ProcessSlot` and `ProcessFork` only once on `state`, pause the state transition, load the data we care about and only then going on with the actual `ProcessBlock`. However I feel stopping and resuming the state transition is more complex?

beacon/blockchain/process_proposal.go
abi87: I should add that overall I think optimistic payload building is a big perf boost.
We may be even more careful with copying (even if I think this requires stopping and starting state transition), but the huge gain is given by letting the EVM work on the payload while CometBFT is waiting on the networking, which it does most of the time.
The latter fact is indirectly confirmed by the fact that reusing cached states does reduce block time only marginally (as we say on some preliminary measures on https://github.com/berachain/beacon-kit/pull/2812).
This is a bit of a bummer, but it also tells you how much would we gain if instead of waiting for 850ms to build a block after `FinalizeBlock`, we had the block immediately ready.

beacon/blockchain/process_proposal.go
```diff
@@ -245,22 +242,16 @@ func (s *Service) VerifyIncomingBlock(
...
 		"Received incoming beacon block",
 		"state_root", beaconBlk.GetStateRoot(),
 		"slot", beaconBlk.GetSlot(),
 	)
 
-	// We purposefully make a copy of the BeaconState in order
-	// to avoid modifying the underlying state, for the event in which
-	// we have to rebuild a payload for this slot again, if we do not agree
-	// with the incoming block.
-	postState := preState.Copy(ctx)
```
abi87, line 260: no state copies anymore if optimistic payload building is disabled.
However one copy needs to be done whether the block passes verification or not.
If the former case we copy the state containing the block changes (to build on top of it), in the latter the state before execution the block.
Copies are necessary because `preFetchBuildData`  does modify the state is receives (via `ProcessSlot` and `ProcessFork`)

beacon/blockchain/process_proposal.go
abi87: About rejecting the proposal if fetching goes wrong, you may be write. I think it's akin a 500 Internal Server Error, but it may be better to:
- log the issue
- let the block be executed
- if the block (unexpectedly) verifies, build the next block JIT after FinalizeBlock is called.

abi87: Implemented. Great catch @calbera!

beacon/blockchain/process_proposal.go
```diff
@@ -295,52 +309,37 @@ func (s *Service) VerifyIncomingBlock(
...
+			// We don't mark the block as rejected if it is valid but pre-fetch fails.
+			// Instead we log the issue and move to process the current block.
+			// Next block can always be built right after current height is finalized.
+			s.logger.Warn(
+				"Failed pre fetching data for optimistic block building",
+				"case", "block success",
+				"err", errFetch,
+			)
+			return nil
 		}
```
calbera, line 341: so just confirming my understanding. We need to 
- use a copy of `state` here since `state` itself will be cached
- preFetch build data again for the next block (since the previous preFetch was only done in case we are rejecting the incoming block)

calbera (approved):
make sense to me now, LGTM

beacon/blockchain/process_proposal.go
abi87: 100%

PR #2821, opened 2025-06-24
Fix float64 unmarshal precision

fridrik01:
This PR fixes a precision loss issue when processing large timestamp values in genesis JSON files. The issue occurred when unmarshaling JSON into `map[string]interface{}`, which by default converts numbers to `float64`, causing precision loss for large integers like `9223372036854775807`.

**Test plan**

Update `testing/files/eth-genesis.json` and set the follwing:

>   "pragueTime": 9223372036854775807 

Run `make start`

Confirmed that the pragueTime timestamp in `.tmp/beacond/eth-genesis.json`  was exactly `9223372036854775807`

### DIFF
--- a/cli/commands/genesis/storage.go
+++ b/cli/commands/genesis/storage.go
@@ -21,6 +21,8 @@
 package genesis
 
 import (
+	"bytes"
+	"encoding/json"
 	"math/big"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


cli/commands/genesis/storage.go
```diff
@@ -175,9 +176,11 @@ func writeGenesisAllocToFile(
 		return err
 	}
 
-	// Unmarshal existing genesis
+	// Unmarshal existing genesis using json.Number to preserve integer precision
 	var existingGenesis map[string]interface{}
-	if err = json.Unmarshal(existingBz, &existingGenesis); err != nil {
+	decoder := json.NewDecoder(bytes.NewReader(existingBz))
+	decoder.UseNumber()
```
abi87, line 182: so our `encoding/json` library looses precision at times. Are there other instances where this is an issue? Should we fix our library implementation instead of adding this implementation?

fridrik01: I looked at other instance but didn't see any. One in particular was the unmarshaling in `cli/commands/genesis/payload.go` where we call ethGenesis.UnmarshalJSON which is implemented in go-etherum, but that unmarshals numbers into math.HexOrDecimal64 which does not lose precision

rezbera (approved):

PR #2823, opened 2025-06-25
chore(EL): upgrade EL client images to latest/nightly builds

rezzmah:
NOTE: merge after cloud syncing tests are complete

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -732,7 +732,6 @@ func ProvideBlockchainService(
 
 ### Build Tags
 - `bls12381` - BLS cryptography
-- `ckzg` - KZG blob commitments
 - `test` - Testing utilities
 - `e2e` - End-to-end tests
 - `simulated` - Simulation tests


node-core/services/version/version.go
calbera, line 1: looking at the GetEthVersion --> we should make sure in both bera-reth and bera-geth that getClientVersion API returns name: "bera-..." with the version: like "v1.011511.0" (correct tag and not the upstream tags)

calbera: bera-geth change: https://github.com/berachain/bera-geth/pull/20

testing/networks/80069/eth-genesis.json
```diff
@@ -94,6 +94,13 @@
         "max": 6,
         "baseFeeUpdateFraction": 3338477
       }
+    },
+    "berachain": {
+      "prague1": {
+        "time": 17446744073709551614,
```
calbera, line 100: convention to consider: geth follows the convention that (0 = fork always enabled, nil = no fork). So if reth also agrees we can actually just remove the "time" line until we have a specific time which we'll enable at.

da/kzg/ckzg/ckzg.go
calbera, line 1: courtesy Cursor - claude-opus-4-max

## Summary

I've successfully implemented option 2 by upgrading beacon-kit to use c-kzg-4844 v2, which matches bera-geth's version. Here's what I did:

### Changes Made:

1. **Updated go.mod**:
   - Changed `github.com/ethereum/c-kzg-4844 v1.0.3` to `github.com/ethereum/c-kzg-4844/v2 v2.1.0`
   - Removed the indirect v2 dependency since it's now direct

2. **Updated imports in beacon-kit KZG files**:
   - Changed imports from `"github.com/ethereum/c-kzg-4844/bindings/go"` to `"github.com/ethereum/c-kzg-4844/v2/bindings/go"` in:
     - `da/kzg/ckzg/ckzg.go`
     - `da/kzg/ckzg/ckzg_cgo.go`

3. **Updated the NewVerifier function**:
   - Modified to handle v2's API which requires 4 parameters: `LoadTrustedSetup(g1Mon, g1Lag, g2s, precomputeWidth)`
   - Since beacon-kit's trusted setup doesn't have monomial data, I used the lagrange data for both g1Mon and g1Lag as a compatibility workaround
   - Used precompute width of 6 (same as bera-geth)

### Result:
- ✅ Build succeeds with `make build`
- ✅ No more duplicate symbol errors
- ✅ Binary runs successfully with `make start`
- ✅ Both beacon-kit and bera-geth now use the same c-kzg-4844 v2 library

The duplicate symbol issue has been resolved by ensuring both projects use the exact same version of the c-kzg-4844 library.

calbera (approved):

rezbera:
lgtm. nit in CLAUDE.md
`- `ckzg` - KZG blob commitments`

PR #2824, opened 2025-06-25
use bera-geth

calbera:


testing/files/eth-genesis.json
```diff
@@ -37,7 +37,7 @@
     },
     "berachain": {
       "prague1": {
-        "time": 9223372036854775807,
+        "time": 0,
```
calbera, line 40: keep our devnet with latest forks starting at genesis

scripts/build/testing.mk
```diff
@@ -118,7 +118,7 @@ start-geth-bepolia:
 	--rm -v $(PWD)/${TESTAPP_FILES_DIR}:/${TESTAPP_FILES_DIR} \
 	--rm -v $(PWD)/${BEPOLIA_NETWORK_FILES_DIR}:/${BEPOLIA_NETWORK_FILES_DIR} \
 	-v $(PWD)/.tmp:/.tmp \
-	ethereum/client-go init \
```
abi87, line 121: we reth fork with the original image for bepolia and mainnet. Why this difference?
I see you have synced, so we should be fine. Still noting we should probably do the same thing across the two forks?

abi87 (approved):
Approved with a single comment

PR #2830, opened 2025-06-26
fix: spec file validation now supports embedded structs

fridrik01:


config/spec/creator_test.go
```diff
@@ -127,3 +130,60 @@ func TestCreateChainSpec_File(t *testing.T) {
...
+			name: "missing max-effective-balance after block-delay-configuration",
+			specContent: `[block-delay-configuration]
+max-block-delay = 300_000_000_000
+target-block-time = 2_000_000_000
+const-block-delay = 500_000_000
+consensus-update-height = 1
+consensus-enable-height = 2
+`,
+			wantErr: "missing required configuration for key: max-effective-balance",
+		},
```
calbera, line 175: can we also add a test case where err is nil? i.e. the valid way to load `block-delay-configuration`, as reference

calbera (approved):
utACK

7fantasy7 (approved):

config/spec/creator_test.go
fridrik01: @calbera good idea, updated the PR

PR #2831, opened 2025-06-26
refactor(spec): remove import cycle and cleanup

calbera:
Some cleanup refactoring to make the fixes code clean

### DIFF
--- a/state-transition/core/state_processor_fixes.go
+++ b/state-transition/core/state_processor_fixes.go
@@ -37,6 +37,12 @@ const (
 	luganodesAmount = 901_393_690_000_000 * params.Wei // 901_393.69 Gwei
 )
 
+//nolint:gochecknoglobals // unexported
+var (
+	luganodesCredentials = ctypes.WithdrawalCredentials(hex.MustToBytes(luganodesCreds))
+	luganodesPubKeyBytes = crypto.BLSPubkey(hex.MustToBytes(luganodesPubKey))
+)
+
 // processElectra1Fixes handles some fixes made necessary by accidents or wrong validator choices in mainnet
 func (sp *StateProcessor) processElectra1Fixes(st *state.StateDB) error {
 	if !sp.cs.IsMainnet() {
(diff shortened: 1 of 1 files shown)


state-transition/core/state_processor_fixes.go
```diff
@@ -61,8 +54,8 @@ func (sp *StateProcessor) processLuganodesRecovery(st *state.StateDB) error {
 
 	// Make a one-time hardcoded deposit
 	deposit := ctypes.Deposit{
-		Pubkey:      luganodesPubKeyBytes,
-		Credentials: luganodesCredentials,
+		Pubkey:      crypto.BLSPubkey(hex.MustToBytes(luganodesPubKey)),
+		Credentials: ctypes.WithdrawalCredentials(hex.MustToBytes(luganodesCreds)),
```
abi87: super nit: if we cast in the vars at package level as we do above, the binary would immediately crash in case of errors, while if we only do the casts here it will only crash when executing the fork.
I would move the vars up

fridrik01: Yeah I do prefer keeping these vars at the package level like they were

state-transition/core/state_processor_fixes.go
```diff
@@ -32,21 +32,14 @@ import (
...
-//nolint:gochecknoglobals // unexported
-var (
-	luganodesCredentials = ctypes.WithdrawalCredentials(hex.MustToBytes(luganodesCreds))
-	luganodesPubKeyBytes = crypto.BLSPubkey(hex.MustToBytes(luganodesPubKey))
-)
-
 // processElectra1Fixes handles some fixes made necessary by accidents or wrong validator choices in mainnet
 func (sp *StateProcessor) processElectra1Fixes(st *state.StateDB) error {
-	if sp.cs.DepositEth1ChainID() != forkChainID {
+	if !sp.cs.IsMainnet() {
```
fridrik01, line 48: 👍

fridrik01 (approved):
lgtm, though would prefer the vars at package level

PR #2834, opened 2025-06-27
Smilee fix on mainnet at electra1 fork

abi87:


rezbera (commented):
Code looks good but there may be another issue around the withdrawal credentials for Smilee not allowing for withdrawals since `receive()` is not triggered.

https://berascan.com/address/0x333bd3beb65f965eea04999ba3c0349a596e9ffe#code

Waiting for confirmation

PR #2836, opened 2025-07-01
chore(engine): Add previous proposer public key to engine API and validate

abi87:
Expanding Engine API to pass previous block proposer public key, thus unlocking BRIP-4, aka PoL Enshrinement.


beacon/blockchain/payload.go
```diff
@@ -236,3 +246,26 @@ func (s *Service) handleOptimisticPayloadBuild(
 
 	s.metrics.markOptimisticPayloadBuildSuccess(buildData.Slot)
 }
+
+func PrevBlockProposerPubKey[
+	ForkChainSpec interface {
+		ActiveForkVersionForTimestamp(timestamp math.U64) common.Version
+	},
```
abi87: this generic seems necessary to reuse this function in both block building  package and here for optimistic block building

engine-primitives/engine-primitives/attributes.go
```diff
@@ -45,6 +46,10 @@ type PayloadAttributes struct {
 	// to the block currently being processed. This field was added for
 	// EIP-4788.
 	ParentBeaconBlockRoot common.Root `json:"parentBeaconBlockRoot"`
+
+	// PrevProposerPubKey carries the public key of previous block proposed
+	// Needed to enshrine distributeFor in EVM
+	PrevProposerPubKey crypto.BLSPubkey `json:"prevProposerPubKey"`
```
abi87: I wonder if, proactivelly, we should make a map of BLSKey -> EffectiveBalance, the latter being a proxy of percentage of reward to be granted to a specific validator

calbera:
```suggestion
	PrevProposerPubKey crypto.BLSPubkey `json:"parentProposerPubkey"`
```
nit: using this tag as it follows the pattern used by `parentBeaconBlockRoot`

engine-primitives/engine-primitives/attributes.go
```diff
@@ -45,6 +46,10 @@ type PayloadAttributes struct {
 	// to the block currently being processed. This field was added for
 	// EIP-4788.
 	ParentBeaconBlockRoot common.Root `json:"parentBeaconBlockRoot"`
+
+	// PrevProposerPubKey carries the public key of previous block proposed
+	// Needed to enshrine distributeFor in EVM
```
calbera:
```suggestion
	// This field was added for BRIP-0004.
```
nit: following the pattern for the ParentBeaconBlockRoot field

engine-primitives/engine-primitives/attributes.go
```diff
@@ -92,5 +99,16 @@ func (p *PayloadAttributes) Validate(forkVersion common.Version) error {
 		return ErrNilWithdrawals
 	}
 
+	emptyBLSPubKey := crypto.BLSPubkey{}
+	if version.IsBefore(forkVersion, version.Electra1()) {
+		if p.PrevProposerPubKey != emptyBLSPubKey {
+			return ErrNonEmptyPrevProposerPubKey
```
calbera, line 105: question: should we actually enforce that its empty? before electra1 we simply do not use / nor need the prevProposerPubkey field

beacon/blockchain/payload.go
calbera: nit: just for code cleanliness, can we define ForkChainSpec in interfaces.go? or better even, use the existing ServiceChainSpec already defined there

engine-primitives/engine-primitives/attributes.go
calbera: Thinking more on it, it's probably fine to enforce the strict usage based on fork version. But reminds me that this struct can now become `Versionable`, i.e. `forkVersion` can be stored as a private field, and `Validate()` doesn't need to take any params. Follows the pattern used throughout

beacon/blockchain/payload.go
abi87: The chain specs in blockchain and payload packages defines different methods and I currently cannot reuse/embed one into another. This is the minimum interface which both must implement to reuse the method.
I am not found of the generic either but don't currently see a better solution. Do you have suggestions.

engine-primitives/engine-primitives/attributes.go
abi87: Renamed across the codebase

engine-primitives/engine-primitives/attributes.go
abi87: Done

engine-primitives/engine-primitives/attributes.go
abi87: I am not sure I would as far as making this `PayloadAttributes` `Versionable` as I think we only use that for SSZ-encoded structs.
BUT I can still simplify the code by unexporting `validate` method, which in prod code is only called in the constructur. I updated the only test where `validate` was explicitly called.

beacon/blockchain/finalize_block.go
```diff
@@ -58,11 +60,18 @@ func (s *Service) FinalizeBlock(
...
 	)
 	blk := signedBlk.GetBeaconBlock()
+	st := s.storageBackend.StateFromContext(ctx)
 
 	// Send an FCU to force the HEAD of the chain on the EL on startup.
 	var finalizeErr error
 	s.forceStartupSyncOnce.Do(func() {
-		finalizeErr = s.forceSyncUponFinalize(ctx, blk)
+		var parentProposerPubKey crypto.BLSPubkey
+		parentProposerPubKey, finalizeErr = st.PrevBlockProposerPubKey(blk.GetTimestamp())
```
abi87: it's important to retrieve `parentProposerPubKey` on the state `st` **before** executing the block.

consensus-types/types/payload_requests.go
```diff
@@ -50,46 +52,56 @@ type newPayloadRequest struct {
 	// versionedHashes is the versioned hashes of the execution payload.
 	versionedHashes []common.ExecutionHash
 	// parentBeaconBlockRoot is the root of the parent beacon block.
-	parentBeaconBlockRoot *common.Root
+	parentBeaconBlockRoot common.Root
 	// ExecutionRequests is introduced in Pectra. It is only non-nil after Pectra.
 	executionRequests []EncodedExecutionRequest
+	// ParentProposerPubkey is introduced in Pectra1. It is only non-nil after Pectra1.
+	parentProposerPubkey *crypto.BLSPubkey
```
calbera, line 59: aiming to standardize this pattern from geth throughout:

For these types which are "optional" (i.e. not in every fork version) we keep them as pointers or nillable. For types like parentBeaconBlockRoot, this is required in every fork version we support so choosing to keep it not nil.

scripts/build/build.mk
```diff
@@ -44,7 +44,7 @@ build_tags += blst
 build_tags += bls12381
 
 # always include ckzg
-build_tags += ckzg
```
abi87, line 47: @calbera why this change?

go.mod
```diff
@@ -6,6 +6,7 @@ replace (
 	github.com/cometbft/cometbft => github.com/berachain/cometbft v1.0.1-0.20250424093341-6fe80335f54a
 	github.com/cometbft/cometbft/api => github.com/berachain/cometbft/api v1.0.1-0.20250424093341-6fe80335f54a
 	github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.52.0-rc.1
+	github.com/ethereum/go-ethereum => ../bera-geth
```
abi87: Is this a long term thing that now BeaconKit depends on bera-geth?
Should we double check for other dependencies?

abi87: Also do we have a public repo to depend on?

calbera: So because of [these checks](https://github.com/berachain/beacon-kit/blob/public-key-in-FCU/state-transition/core/state_processor_payload.go#L182-L183), it now becomes a strict requirement to use bera-geth (since we're basically enforcing the EL level protocol rules in the CL). 

bera-geth is public (https://github.com/berachain/bera-geth/pull/21), and as soon as this branch is well tested, will merge it and release a tag to be used here

scripts/build/build.mk
calbera: something I needed to do to get bera-geth building as a dependency. will investigate further

kurtosis/src/networks/kurtosis-devnet/network-configs/genesis.json.template
```diff
@@ -50,6 +51,16 @@
...
+        "nonce": "0x1",
+        "balance": "0x0"
+      },
+      "0x4200000000000000000000000000000000000043": {
+        "code": "0x608060405234801561000f575f80fd5b5060043610610034575f3560e01c80634b28f9a214610038578063999da65b14610052575b5f80fd5b6100405f5481565b60405190815260200160405180910390f35b6100656100603660046100b8565b610067565b005b5f8054908061007583610126565b91905055507fb3a8fa51f8d3759f320e88b7f8d3fb73a2a51b31b3324b37833c4816cf41e7c45f546040516100ac91815260200190565b60405180910390a15050565b5f80602083850312156100c9575f80fd5b823567ffffffffffffffff8111156100df575f80fd5b8301601f810185136100ef575f80fd5b803567ffffffffffffffff811115610105575f80fd5b856020828401011115610116575f80fd5b6020919091019590945092505050565b5f6001820161014357634e487b7160e01b5f52601160045260245ffd5b506001019056fea2646970667358221220350dbb7eed9e4d6e4ff72594b0582b0b2e4c1d6b11c3e5ad382201b47a638cc664736f6c634300081a0033",
+        "nonce": "0x1",
+        "balance": "0x0"
+      },
```
rezzmah, line 63: mock pol

scripts/build/build.mk
```diff
@@ -44,7 +44,7 @@ build_tags += blst
...
+# build_tags += ckzg
```
rezzmah, line 47: we still need to figure this out

beacon/blockchain/finalize_block.go
```diff
@@ -54,7 +54,7 @@ func (s *Service) FinalizeBlock(
 	s.forceStartupSyncOnce.Do(func() {
 		var parentProposerPubkey *crypto.BLSPubkey
 		parentProposerPubkey, finalizeErr = st.ParentProposerPubkey(blk.GetTimestamp())
-		if err != nil {
+		if finalizeErr != nil {
```
abi87, line 57: nice, thanks for catching this up

go.mod
```diff
@@ -6,6 +6,7 @@ replace (
 	github.com/cometbft/cometbft => github.com/berachain/cometbft v1.0.1-0.20250718121541-9671b1dc07e5
 	github.com/cometbft/cometbft/api => github.com/berachain/cometbft/api v1.0.1-0.20250718121541-9671b1dc07e5
 	github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.52.0-rc.1
+	github.com/ethereum/go-ethereum => github.com/berachain/bera-geth v0.0.0-20250721211541-88a3626f73ad
```
abi87, line 9: it'd probably be best to tag and depend on a specific tag rather than a commit

consensus-types/types/payload_requests.go
```diff
@@ -116,16 +116,9 @@ func (n *newPayloadRequest) GetParentBeaconBlockRoot() common.Root {
...
 func (n *newPayloadRequest) GetParentProposerPubkey() (*crypto.BLSPubkey, error) {
-	if version.IsBefore(n.GetForkVersion(), version.Electra1()) {
-		return nil, errors.Wrap(
-			ErrForkVersionNotSupported,
-			"parent proposer pubkey not supported in newPayloadRequest before electra",
-		)
-	}
-	if n.parentProposerPubkey == nil {
-		return nil, errors.Wrap(ErrNilValue, "parent proposer pubkey cannot be nil")
-	}
```
abi87: all of this logic is in `BuildNewPayloadRequestFromFork` and `parentProposerPubkey` is unexported. So the getter can be as simple as possible and let the client have a nil pub key (which is fine we usually just pass it around to EngineAPI)

consensus-types/types/payload_requests.go
```diff
@@ -100,10 +112,16 @@ func (n *newPayloadRequest) GetVersionedHashes() []common.ExecutionHash {
...
-func (n *newPayloadRequest) GetParentBeaconBlockRoot() *common.Root {
+func (n *newPayloadRequest) GetParentBeaconBlockRoot() common.Root {
 	return n.parentBeaconBlockRoot
 }
 
+// GetParentProposerPubkey may return a nil parent pub key. See BuildNewPayloadRequestFromFork
+// to understand how parentProposerPubkey gets populated
+func (n *newPayloadRequest) GetParentProposerPubkey() *crypto.BLSPubkey {
+	return n.parentProposerPubkey
+}
```
abi87, line 123: all of this logic is in BuildNewPayloadRequestFromFork and parentProposerPubkey is unexported. So the getter can be as simple as possible and let the client have a nil pub key (which is fine we usually just pass it around to EngineAPI)

PR #2845, opened 2025-07-09
chore: fix some minor issues in comments

liangmulu:
fix some minor issues in comments


calbera:
Thank you for the PR! But we'd like to include more productive commits rather than typo fixes :)

PR #2855, opened 2025-07-17
perf(blockchain): Efficient optimistic block building

abi87:
Currently if efficient block building is active, every block builder will try and build next block, even if only one will propose the next block (under normal network condition).
Following https://github.com/berachain/cometbft/pull/33, `ProcessProposal` will carry the expected next block builder (whether is the first block of next height or the block at the next round of current height). This allow a single node to get ready to propose while others just quiesce.

### DIFF
--- a/beacon/blockchain/process_proposal.go
+++ b/beacon/blockchain/process_proposal.go
@@ -27,7 +27,7 @@ import (
 	"time"
 
 	ctypes "github.com/berachain/beacon-kit/consensus-types/types"
-	"github.com/berachain/beacon-kit/consensus/cometbft/service/encoding"
+	"github.com/berachain/beacon-kit/consensus/cometbft/service/cache"
 	"github.com/berachain/beacon-kit/consensus/types"
 	datypes "github.com/berachain/beacon-kit/da/types"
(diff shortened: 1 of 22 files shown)


beacon/blockchain/process_proposal.go
```diff
@@ -249,11 +256,12 @@ func (s *Service) VerifyIncomingBlock(
 	}
 
 	var (
-		nextBlockData *builder.RequestPayloadData
-		errFetch      error
+		nextBlockData        *builder.RequestPayloadData
+		errFetch             error
+		shouldBuildNextBlock = s.shouldBuildOptimisticPayloads(isNextBlockProposer)
```
fridrik01, line 261: nit: imo more readable as:

```suggestion
		shouldBuildNextBlock = isNextBlockProposer && s.shouldBuildOptimisticPayloads()
```

abi87: My issue with this is that we may forgot to check this condition if `s. shouldBuildOptimisticPayloads` is called multiple time. As ugly as passing `isNextBlockProposer` as parameter is, we get the help of the compiler to avoid forgetting this check

PR #2862, opened 2025-07-28
chore(bepolia): pectra11 hardfork

calbera:
Pectra11 hard fork on Bepolia at _August 6th, noon EST_ (unix: `1754496000`). Stable block time enabled ~24 hours before at block `7768335` so that as PoL switches over to the new distribution method, there are no hiccups with an already stable block time.

### DIFF
--- a/config/spec/testnet.go
+++ b/config/spec/testnet.go
@@ -21,8 +21,6 @@
 package spec
 
 import (
-	"math"
-
 	"github.com/berachain/beacon-kit/chain"
 )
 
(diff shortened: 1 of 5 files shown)


config/spec/testnet.go
```diff
@@ -44,8 +42,12 @@ func TestnetChainSpecData() *chain.SpecData {
 	// Timestamp of the Electra fork on Bepolia.
 	specData.ElectraForkTime = 1746633600
 
+	// Enable stable block time before the Electra1 fork.
+	specData.Config.ConsensusUpdateHeight = 7_768_334
+	specData.Config.ConsensusEnableHeight = 7_768_335
```
abi87, line 47: At the moment of my review Bepolia is at block `
7_422_212` and has an average block time of `1.8 seconds`.
This means that it will take `(7_768_335 - 7_422_212) *1.8 seconds == 173 hours` to activate the fork. So fork will happen somewhere `5th August` which is what we expect

abi87 (approved):

PR #2866, opened 2025-07-30
WIP: Deprecate Karalabe for fastssz (and use sszgen where possible)

fridrik01:
-->Do not review yet<--

TODO:
- [ ] fix failing linters
- [ ] Some compatibility tests are not correct and need to be updated
- [ ] for structs with sszgen we need to explicitly run ValidateAfterDecodingSSZ as this is not generated by sszgen
- [ ] fix failing e2e test (due to no rpc error) 
- [ ] Investigate into splitting into multiple PRs (however it is difficult to get individual PRs that compile due to dependencies between structs)
- [ ] make sure is stable on `make start-devnet` 
- [ ] deploy to devnet-core and observe stability/compatibility with other validators using `main` branch


calbera:
Maybe worth upgrading to v1.0.0 as well (https://github.com/berachain/beacon-kit/pull/2876)

PR #2870, opened 2025-08-06
align SBT consensus parameters with chain specs

abi87:
- align consensus parameters with chain specs at node restart
- cleanup initialize command to reuse chain specs (with UT)

### DIFF
--- a/testing/simulated/components.go
+++ b/testing/simulated/components.go
@@ -85,6 +85,7 @@ func ProvideElectraGenesisChainSpec() (chain.Spec, error) {
 	specData.GenesisTime = 0
 	specData.Deneb1ForkTime = 0
 	specData.ElectraForkTime = 0
+	specData.Electra1ForkTime = 9223372036854775807
 	// We set slots per epoch to 2 for faster observation of withdrawal behaviour
 	specData.SlotsPerEpoch = 2
 	// We set this to 4 so tests are faster
@@ -120,6 +121,7 @@ func ProvidePectraForkTestChainSpec() (chain.Spec, error) {
 	specData.GenesisTime = 0
 	specData.Deneb1ForkTime = 0
 	specData.ElectraForkTime = 10
+	specData.Electra1ForkTime = 9223372036854775807
 	chainSpec, err := chain.NewSpec(specData)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


consensus/cometbft/service/service.go
```diff
@@ -156,6 +156,13 @@ func NewService(
 		panic(fmt.Errorf("failed loading latest version: %w", err))
 	}
 
+	// Update Stable block time related data
+	lastCommitID := s.sm.GetCommitMultiStore().LastCommitID()
+	lastBlockHeight := lastCommitID.Version
```
abi87: similarly to what we do in `s.Info`, we retrieve latest block height here and set `SBTEnableHeight` parameter if needed.
This is done to avoid any change to FinalizeBlock code, where we took care of making writes to `SBTEnableHeight` happen only once (at `ConsensusUpdateHeight`)

consensus/cometbft/service/configs.go
```diff
@@ -121,9 +122,7 @@ func DefaultConsensusParams(consensusKeyAlgo string) *cmttypes.ConsensusParams {
 	res.Feature.PbtsEnableHeight = 1
 	res.Synchrony.Precision = precision
 	res.Synchrony.MessageDelay = messageDelay
-
-	// Activate Stable Block Time (SBT) by default.
-	res.Feature.SBTEnableHeight = 1
+	res.Feature.SBTEnableHeight = cs.SbtConsensusEnableHeight()
```
abi87, line 125: Just avoid hardcoding SBT enable height and duly use chainSpecs to pick the right value for it.
This is likely to make a difference only on devnet tests.

consensus/cometbft/service/service.go
```diff
@@ -156,6 +156,14 @@ func NewService(
 		panic(fmt.Errorf("failed loading latest version: %w", err))
 	}
 
+	// Make sure that SBT consensus parameters are duly set when the node restart.
+	// Note that we can't rely on genesis.json having these parameters set right
+	// because we introduced stable block time post (mainnet) genesis.
+	lastBlockHeight := s.sm.GetCommitMultiStore().LastCommitID().Version
```
abi87, line 162: See https://github.com/berachain/beacon-kit/pull/2870#discussion_r2256830191

fridrik01 (approved):
lgtm

PR #2873, opened 2025-08-07
fix: update simulated tests electra1 timestamp

rezbera:


testing/simulated/components.go
```diff
@@ -85,6 +85,7 @@ func ProvideElectraGenesisChainSpec() (chain.Spec, error) {
 	specData.GenesisTime = 0
 	specData.Deneb1ForkTime = 0
 	specData.ElectraForkTime = 0
+	specData.Electra1ForkTime = 9223372036854775807
```
abi87, line 88: should we instead keep the tests updated to the latest fork?

rezzmah: we should, i just dont have time to update them atm

abi87: It can definitely be a post-fork activation task

abi87 (approved):

PR #2868, opened 2025-08-01
temp mock pol

rezzmah:
```solidity
// SPDX-License-Identifier: MIT
pragma solidity 0.8.26;

import "./ValidatorRegistry.sol";

/// @title SimplePoLDistributor
/// @notice Simple mock PoL distributor for testing bera-reth
/// @dev Updated to interact with ValidatorRegistry to test multi-contract state changes
contract SimplePoLDistributor {
    /// @notice System address that can call distributeFor (execution layer client)
    address private constant SYSTEM_ADDRESS = 0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE;
    
    /// @notice The validator registry contract address (hardcoded for genesis deployment)
    ValidatorRegistry private constant VALIDATOR_REGISTRY = ValidatorRegistry(0x4200000000000000000000000000000000000043);
    
    /// @notice Event emitted when distributeFor is called
    event PoLDistributed(bytes pubkey);
    
    /// @notice Counter for total distributions
    uint256 public totalDistributions;
    
    /// @notice Error thrown when caller is not the system address
    error NotSystemAddress();
    
    /// @dev Modifier to restrict function access to system address.
    /// @dev This ensures only the execution layer client can call `distributeFor` function.
    modifier onlySystemCall() {
        if (msg.sender != SYSTEM_ADDRESS) {
            revert NotSystemAddress();
        }
        _;
    }
    
    /// @notice Main function that bera-reth will call
    /// @param pubkey The validator public key
    /// @dev Now calls the ValidatorRegistry to test multi-contract state changes
    function distributeFor(bytes calldata pubkey) external onlySystemCall {
        require(totalDistributions < 10, "Max distributions reached");
        // Update state in this contract
        totalDistributions++;
        
        // Call another contract to update its state
        // This tests whether system calls properly capture state changes from multiple contracts
        VALIDATOR_REGISTRY.recordValidatorActivity(pubkey);
        
        emit PoLDistributed(pubkey);
    }
}


// SPDX-License-Identifier: MIT
pragma solidity 0.8.26;

/// @title ValidatorRegistry
/// @notice A simple registry contract that increments a counter when called
/// @dev This contract will be called by SimplePoLDistributor to test multi-contract state changes
contract ValidatorRegistry {
    /// @notice Simple counter that increments on each call
    uint256 public callCount;
    
    /// @notice Event emitted when the contract is called
    event RegistryCalled(uint256 newCount);
    
    /// @notice Records activity - simply increments counter and emits event
[...]

PR #2878, opened 2025-08-11
feat(testing): Add Contracts for Mock PoL

rezbera:
supercedes https://github.com/berachain/beacon-kit/pull/2843 and https://github.com/berachain/beacon-kit/pull/2868


contracts/script/GetBytecode.s.sol
calbera, line 1: I would prefer not including this and instead explaining in the README how to get code via `forge build —extra-output` or via `forge create` and checking the deployed bytecode

PR #2885, opened 2025-08-18
chore(make): `bera-geth` skip init if home folder already there

abi87:
Apparent `bera-geth init` is not idempotent anymore, so we need to skip it if the beacond home folder is already there.
Note that if you run `make start` and `make start-geth` without then feeding any load to the chain, you can safely restart the nodes with no problem.
Instead if some load has been provided to the chain (e.g. deploying PoL) then an error like the following would appear:

```
INFO [08-18|12:31:43.727] Opened ancient database                  database=/.tmp/eth-home/bera-geth/chaindata/ancient/chain readonly=false
INFO [08-18|12:31:43.727] Opened Era store                         datadir=/.tmp/eth-home/bera-geth/chaindata/ancient/chain/era
INFO [08-18|12:31:43.727] State scheme set to already existing     scheme=path
INFO [08-18|12:31:43.728] Initialising Ethereum protocol           network=80087 dbversion=9
WARN [08-18|12:31:43.728] Sanitizing invalid node buffer size      provided=653.00MiB updated=256.00MiB
INFO [08-18|12:31:43.728] Load database journal from file          path=/.tmp/eth-home/bera-geth/triedb/merkle.journal
INFO [08-18|12:31:43.756] Opened ancient database                  database=/.tmp/eth-home/bera-geth/chaindata/ancient/state readonly=false
CRIT [08-18|12:31:43.756] Failed to truncate extra state histories err="out of range, tail: 0, head: 1, target: 5"
```

I manually tested that this is not the case for devnet with the proposed PR

### DIFF
--- a/scripts/build/testing.mk
+++ b/scripts/build/testing.mk
@@ -7,8 +7,10 @@
 # ask_reset_dir_func checks if the directory passed in exists, and if so asks the user whether it
 # should delete it. Note that on linux, docker may have created the directory with root
 # permissions, so we may need to ask the user to delete it with sudo
+# Creates a flag file to indicate if the directory was reset
 define ask_reset_dir_func
 	@abs_path=$(abspath $(1)); \
+	rm -f .tmp/.dir_was_reset; \
 	if test -d "$$abs_path"; then \
 		read -p "Directory '$$abs_path' exists. Do you want to delete it? (y/n): " confirm && \
 		if [ "$$confirm" = "y" ]; then \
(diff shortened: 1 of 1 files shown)


scripts/build/testing.mk
```diff
@@ -76,12 +81,18 @@ start-reth:
 ## Start an ephemeral `geth` node with docker
 start-geth: 
 	$(call ask_reset_dir_func, $(ETH_DATA_DIR))
-	docker run \
-	--rm -v $(PWD)/${TESTAPP_FILES_DIR}:/${TESTAPP_FILES_DIR} \
-	-v $(PWD)/.tmp:/.tmp \
-	ghcr.io/berachain/bera-geth:latest init \
-	--datadir ${ETH_DATA_DIR} \
-	${ETH_GENESIS_PATH}
+	@if [ -f .tmp/.dir_was_reset ]; then \
```
fridrik01, line 84: we could just test for the existence of `.tmp/chaindata` and only run `bera-geth init` if that folder does not exist. It would be simpler IMO as we won't need the `.dir_was_reset` file for that

Something like:

```
@if [  ! -d .tmp/chaindata ]; then \
```

abi87: The reason I went for this solution is that I want `ask_reset_dir_func` to be the only place where the logic reset or init is decided. Still I understand your concern.
How about we make `ask_reset_dir_func` return a proper value and implement a conditional on top of it?
No file, but logic in a single place

PR #2891, opened 2025-08-26
chore(node-endpoint) : Implement node_version

abi87:
Implements /eth/v1/node/version
Replaces https://github.com/berachain/beacon-kit/pull/2046.

Reuses the same version provided by the command `beacon version`.
Added UTs in stead of e2e tests.
Most of the LoC are mocks and unit tests.


node-api/backend/backend.go
```diff
@@ -113,3 +115,24 @@ func (b *Backend) Spec() (chain.Spec, error) {
...
 	return b.cs, nil
 }
+
+func (b *Backend) GetVersionData() (
+	string, // appName
+	string, // cometVersion
+	string, // os
+	string, // arch
+) {
+	cometVersionInfo := version.NewInfo() // same used in beacond version command
```
abi87, line 125: while `version.NewInfo()` is a global getter that could be retrieved directly in the `handler.Version` method, I pass it via backend to simplify testing (and avoid modifying in tests the global variable behind `version.NewInfo()`)

node-api/handlers/node/node.go
```diff
@@ -46,18 +51,22 @@ func (h *Handler) Syncing(handlers.Context) (any, error) {
...
+// - 14 → number of commits since that tag
+// - g2414721 → the abbreviated commit hash, **with a leading g**.
+// That g stands for “git”. It’s a prefix Git uses to distinguish the commit hash from other possible identifiers.
 func (h *Handler) Version(handlers.Context) (any, error) {
-	type VersionResponse struct {
-		Data struct {
-			Version string `json:"version"`
-		} `json:"data"`
+	appName, version, os, arch := h.backend.GetVersionData()
+	r := &types.VersionData{
```
abi87, line 62: see https://github.com/OffchainLabs/prysm/blob/240cd1d0588ab660cd6f35800a8f70d785b05290/beacon-chain/rpc/eth/node/handlers.go#L106 for comparison

node-api/handlers/node/node.go
```diff
@@ -46,18 +51,22 @@ func (h *Handler) Syncing(handlers.Context) (any, error) {
 	return response, nil
 }
 
-// Version is a placeholder so that beacon API clients don't break.
-//
-// TODO: Implement with real data.
+// Note: version comes from git describe (via the build process)
```
abi87, line 54: TIL. cc @camembera for knowledge

PR #2894, opened 2025-08-27
chore(node-endpoint) fix beacon genesis return statuses

abi87:


consensus/cometbft/service/node_api_support.go
```diff
@@ -0,0 +1,84 @@
...
+	errorsmod "github.com/berachain/beacon-kit/errors"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+var ErrAppNotReady = fmt.Errorf("%s is not ready; please wait for first block", AppName)
+
+// CreateQueryContext creates a new sdk.Context for a query, taking as args
+// the block height and whether the query needs a proof or not.
+func (s *Service) CreateQueryContext(
```
abi87, line 36: two changes here:

- moved this method here from `service.go` file
- introduce a flag error in case `lastBlockHeight == 0` which signals that genesis has not been processed yet (e.g. genesis time is in the future)

node-api/handlers/beacon/genesis.go
```diff
@@ -24,15 +24,18 @@ import (
...
+	"github.com/berachain/beacon-kit/primitives/common"
 )
 
-func (h *Handler) GetGenesis(_ handlers.Context) (any, error) {
+func (h *Handler) GetGenesis(handlers.Context) (any, error) {
 	genesisRoot, err := h.backend.GenesisValidatorsRoot()
 	if err != nil {
 		return nil, err
 	}
-	if len(genesisRoot) == 0 {
```
abi87, line 34: This is the main point of this PR.
genesisRoot is of type `common.Root` which is `[RootSize]byte`. Length check is non-sense, it would never be true. What we want instead is comparing this with empty root.

node-api/backend/genesis_test.go
```diff
@@ -59,31 +109,17 @@ func setupStateWithGenesisValues(
 	sdkCtx.MultiStore().(storetypes.CacheMultiStore).Write()
 }
 
-func TestGetGenesisData(t *testing.T) {
```
abi87, line 62: diff tripping here. I moved the whole piece of code needed to generate and store CometBFT config into a tmp folder to its own method

node-api/backend/backend_test.go
```diff
@@ -1,81 +0,0 @@
...
-type testConsensusService struct {
-	cms     storetypes.CommitMultiStore
-	kvStore *beacondb.KVStore
-	cs      chain.Spec
-}
-
-func (t *testConsensusService) CreateQueryContext(height int64, _ bool) (sdk.Context, error) {
-	sdkCtx := sdk.NewContext(t.cms.CacheMultiStore(), false, log.NewNopLogger())
-
-	// there validations mimics consensus service, not sure if they are necessary
```
abi87, line 51: dropped and replaced by mock. We don't need duplicated code

node-api/handlers/beacon/genesis.go
fridrik01: you probably did this already, but did you try with running `make start` using a genesis time set in the future to double check this error is returned?

abi87: Yes I did and yes it does return the error (404 now as expected by the specs)

PR #2899, opened 2025-08-28
empty suggested-fee-recipient causes crash - v2

abi87:


cli/utils/parser/validator.go
```diff
@@ -46,24 +45,6 @@ func ConvertPubkey(pubkey string) (crypto.BLSPubkey, error) {
 	return crypto.BLSPubkey(pubkeyBytes), nil
 }
 
-// ConvertWithdrawalAddress converts a string to a withdrawal address.
```
abi87, line 49: Not needed anymore since common.NewExecutionAddressFromHex returns an error

Issue #2378, opened 2025-01-19
Block store service should only be enabled if beacon node-api enabled


calbera:
2 options
- remove block-store-service-enabled from confg
- explicitly mention for node operators that currently its only required for the proof/ namespace of beacon node-api

PR #2901, opened 2025-08-29
chore(flags): cleaned up block-store-service enabled flag

abi87:
Proposal to solve https://github.com/berachain/beacon-kit/issues/2378 as follows:

- Drop `block-store-service.enabled` flag. Reuse `block-store-service.availability-window` instead. Specifically:
  - Disable block store storage if  `availability-window` is set to zero.
  - Keep `availability-window` default value as is for backward compatibility reasons
- If block store is disabled, let storing a block via `Set` proceed without error, but do not store the blocks. Instead whenever a get method is requested, return a block store not enabled error
- Extended config comments to clarify the configuration semantics, mentioning proof/namespace APIs


.vscode/launch.json
```diff
@@ -115,7 +115,6 @@
                 "--beacon-kit.logger.log-level=info",
                 "--home=${workspaceFolder}/.tmp/beacond",
                 "--beacon-kit.engine.jwt-secret-path=${workspaceFolder}/testing/files/jwt.hex",
-                "--beacon-kit.block-store-service.enabled",
```
abi87, line 118: just use availabilityWindow flag (zero means disabled)

config/template/template.go
```diff
@@ -95,10 +95,9 @@ graffiti = "{{ .BeaconKit.Validator.Graffiti }}"
 enable-optimistic-payload-builds = "{{ .BeaconKit.Validator.EnableOptimisticPayloadBuilds }}"
 
 [beacon-kit.block-store-service]
-# Enabled determines if the block store service is enabled.
-enabled = "{{ .BeaconKit.BlockStoreService.Enabled }}"
-
 # AvailabilityWindow is the number of slots to keep in the store.
+# Setting AvailabilityWindow disables block store and does not allow the node
```
fridrik01:
```suggestion
# Setting AvailabilityWindow to 0 disables block store and does not allow the node
```

testing/networks/80069/app.toml
```diff
@@ -189,11 +189,9 @@ graffiti = ""
 # process-proposal to allow for the execution client to have more time to assemble the block.
 enable-optimistic-payload-builds = "false"
 
-[beacon-kit.block-store-service]
-# Enabled determines if the block store service is enabled.
-enabled = "false"
-
 # AvailabilityWindow is the number of slots to keep in the store.
+# Setting AvailabilityWindow disables block store and does not allow the node
```
fridrik01:
```suggestion
# Setting AvailabilityWindow to 0 disables block store and does not allow the node
```

storage/block/store.go
```diff
@@ -30,9 +30,14 @@ import (
 	lru "github.com/hashicorp/golang-lru/v2"
 )
 
+var ErrBlockStoreNotEnabled = errors.New("block store not enabled")
+
 // KVStore is a simple memory store based implementation that stores metadata of
 // beacon blocks.
 type KVStore[BeaconBlockT BeaconBlock] struct {
+	// Setting availabilityWindow to zero upon
```
fridrik01: Not completed comment

node-api/block_store/config.go
```diff
@@ -26,16 +26,13 @@ const (
 
 // Config is the configuration for the block service.
 type Config struct {
-	// Enabled enables the block service.
-	Enabled bool `mapstructure:"enabled"`
```
fridrik01: Was `Enabled` actually used at all? I don't see where we check if its disable and skips enabling the block service

abi87: It was not indeed, but it must have been configurable from config files still.

config/template/template.go
abi87: Fixed

storage/block/store.go
abi87: Fixed

testing/networks/80069/app.toml
abi87: Fixed

PR #2905, opened 2025-09-01
chore(node-api): layers restructuring

abi87:
A minor cleanup of `node-api` folder, to ease up readability. In more details:

- Renamed `engine` to `middleware`. Seems more suggestive to me of its role (map requests to handler, validate requests and map handlers errors to http statuses).
- Hard-coded handlers middleware and backend creation in `node-api Server` constructor and avoided all the dependency injection shenanigans for `Server` internals.
  -  Granted we lose some freedom in doing dependency injections of server components, we don't really use it currently and by removing them we spare a log of code.
- Dropped `AttachQueryBackend`, which does not really make sense to me. Currently consensus backend is passed to node-api backend directly in the constructor, avoiding unnecessary cycles.
- Simplified `node-api Backend` constructor and move data loading to `LoadData`, to be called only if node-apis are effectivelly started.


node-api/middleware/middleware.go
```diff
@@ -27,21 +27,14 @@ import (
...
-// Engine is an implementation of the API engine interface using Echo.
-type Engine struct {
+// Middleware is an implementation of the API engine interface using Echo.
+type Middleware struct {
 	*echo.Echo
 	logger log.Logger
 }
 
-// New initializes a new API engine with the given Echo instance.
-func New(e *echo.Echo) *Engine {
```
abi87, line 37: just moved below and unexported

node-api/server/interfaces.go
```diff
@@ -1,32 +0,0 @@
...
-
-package server
-
-import (
-	"github.com/berachain/beacon-kit/log"
-	"github.com/berachain/beacon-kit/node-api/handlers"
-)
-
-// Engine is an interface for an API engine.
-type Engine interface {
```
abi87, line 29: instantiated directly into the server, we don't have proof this extra interface is needed for any testing, so dropped

storage/block/config.go
```diff
@@ -18,7 +18,7 @@
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, NON-INFRINGEMENT, AND
 // TITLE.
 
-package blockstore
```
abi87, line 21: moved where it belongs. This has not much to do with node-api

cmd/beacond/defaults.go
```diff
@@ -60,17 +60,5 @@ func DefaultComponents() []any {
 		components.ProvideNodeAPIEngine,
 		components.ProvideNodeAPIBackend,
 	)
-
-	c = append(c,
```
abi87, line 64: moved instantiation to node-api server, so all of this is not needed anymore

node-api/server/context/context.go
```diff
@@ -1,26 +0,0 @@
...
-// LICENSOR OR ITS AFFILIATES (PROVIDED THAT YOU MAY USE A TRADEMARK OR LOGO OF
-// LICENSOR AS EXPRESSLY REQUIRED BY THIS LICENSE).
-//
-// TO THE EXTENT PERMITTED BY APPLICABLE LAW, THE LICENSED WORK IS PROVIDED ON
-// AN “AS IS” BASIS. LICENSOR HEREBY DISCLAIMS ALL WARRANTIES AND CONDITIONS,
-// EXPRESS OR IMPLIED, INCLUDING (WITHOUT LIMITATION) WARRANTIES OF
-// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, NON-INFRINGEMENT, AND
-// TITLE.
-
-package context
```
abi87, line 21: unused interface

node-api/server/server.go
```diff
@@ -25,37 +25,55 @@ import (
...
+	// instantiate handlers and register their routes in the middleware
+	var handlers []handlers.Handlers
+	handlers = append(handlers, beaconapi.NewHandler(b))
+	handlers = append(handlers, builderapi.NewHandler())
+	handlers = append(handlers, configapi.NewHandler(b))
+	handlers = append(handlers, debugapi.NewHandler(b))
+	handlers = append(handlers, eventsapi.NewHandler())
+	handlers = append(handlers, nodeapi.NewHandler(b))
+	handlers = append(handlers, proofapi.NewHandler(b))
+
```
abi87: here we lose the ability to do dependency injection of handlers but:

- we don't really use it now and not sure we will in the future (APIs are kinda independent)
- we spare so much scaffolding code

We can always revisit this cleanup and re-enable dep injection if/when we will need it in the future

node-api/server/server.go
```diff
@@ -22,56 +22,95 @@ package server
...
 	}
+
+	mware := middleware.NewDefaultMiddleware()
+
+	// instantiate handlers and register their routes in the middleware
+	b, err := backend.New(storageBackend, cs, cmtCfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed instantiating handlers backend: %w", err)
+	}
+	b.AttachQueryBackend(consensusService)
```
abi87: this used to be done, post-hoc, in node builder. I think it's cleaner to do it here, and clearly signal in one place all the relevant dependencies.

node-core/builder/builder.go
```diff
@@ -101,11 +97,7 @@ func (nb *NodeBuilder) Build(
 	if config == nil {
 		panic("config is nil")
 	}
-	if apiBackend == nil {
-		panic("node or api backend is nil")
-	}
 
 	logger.WithConfig(config.GetLogger())
-	apiBackend.AttachQueryBackend(cmtService)
```
abi87, line 109: moved to node-api server constructor

node-api/server/server.go
```diff
@@ -91,3 +135,7 @@ func (s *Server) Stop() error {
 func (s *Server) Name() string {
 	return "node-api-server"
 }
+
+func (s *Server) GetBackend() *backend.Backend {
+	return s.b
+}
```
abi87, line 145: exposed for simulated tests, which do require to assert state of some validators.
I am very annoyed that we need to go to the backend to get this info. I would find more natural to be able to query the handler or the served to do that

node-api/backend/backend.go
```diff
@@ -86,12 +88,6 @@ func New(
 	return b, nil
 }
 
-// AttachQueryBackend sets the node on the backend for
-// querying historical heights.
-func (b *Backend) AttachQueryBackend(node types.ConsensusService) {
```
abi87, line 91: This never made sense. Now node is passed in backend constructor

node-api/backend/genesis_test.go
```diff
@@ -198,9 +198,8 @@ func TestGetGenesisData_SunnyPath(t *testing.T) {
 			tc.setMockExpectations(cms, kvStore, tcs)
 
 			// 3- Build backend
-			b, err := backend.New(sb, cs, cmtCfg)
-			require.NoError(t, err)
-			b.AttachQueryBackend(tcs)
```
abi87, line 203: AttachQuery is carried out in backend.New, while genesis data caching moved out of backend.New into b.LoadData (to avoid doing this is apis are not enabled)

PR #2907, opened 2025-09-01
chore(node-api): handlers cleanup

abi87:
- Dropped `handler.RegisterRoutes` method and instead registered routes directly in handler constructor
- Dropped `BaseHandler.SetLogger` method and passed along loggers via constructor
- a few other minor nits (mostly moving some interfaces definition around)

### DIFF
--- a/node-api/server/server.go
+++ b/node-api/server/server.go
@@ -28,7 +28,6 @@ import (
 	"github.com/berachain/beacon-kit/log"
 	"github.com/berachain/beacon-kit/log/noop"
 	"github.com/berachain/beacon-kit/node-api/backend"
-	"github.com/berachain/beacon-kit/node-api/handlers"
 	beaconapi "github.com/berachain/beacon-kit/node-api/handlers/beacon"
 	builderapi "github.com/berachain/beacon-kit/node-api/handlers/builder"
 	configapi "github.com/berachain/beacon-kit/node-api/handlers/config"
@@ -41,12 +40,6 @@ import (
 	"github.com/berachain/beacon-kit/node-core/types"
 	cmtcfg "github.com/cometbft/cometbft/config"
 )
-
-// handler is an interface that all handlers must implement.
-// to let middleware register their routes.
-type handler interface {
-	RouteSet() *handlers.RouteSet
-}
 
 // Server is the API Server service.
 type Server struct {
@@ -84,18 +77,13 @@ func New(
 
 	// instantiate handlers and register their routes in the middleware
 	b := backend.New(storageBackend, cs, cmtCfg, consensusService)
-	handlers := make([]handler, 0)
-	handlers = append(handlers, beaconapi.NewHandler(b, apiLogger))
-	handlers = append(handlers, builderapi.NewHandler(apiLogger))
-	handlers = append(handlers, configapi.NewHandler(b, apiLogger))
-	handlers = append(handlers, debugapi.NewHandler(b, apiLogger))
-	handlers = append(handlers, eventsapi.NewHandler(apiLogger))
-	handlers = append(handlers, nodeapi.NewHandler(b, apiLogger))
-	handlers = append(handlers, proofapi.NewHandler(b, apiLogger))
-
-	for _, handler := range handlers {
-		mware.RegisterRoutes(handler.RouteSet())
-	}
+	mware.RegisterRoutes(beaconapi.NewHandler(b, apiLogger).RouteSet())
+	mware.RegisterRoutes(builderapi.NewHandler(apiLogger).RouteSet())
+	mware.RegisterRoutes(configapi.NewHandler(b, apiLogger).RouteSet())
+	mware.RegisterRoutes(debugapi.NewHandler(b, apiLogger).RouteSet())
+	mware.RegisterRoutes(eventsapi.NewHandler(apiLogger).RouteSet())
+	mware.RegisterRoutes(nodeapi.NewHandler(b, apiLogger).RouteSet())
+	mware.RegisterRoutes(proofapi.NewHandler(b, apiLogger).RouteSet())
 
 	return &Server{
 		config:     config,


node-api/handlers/beacon/backend.go
```diff
@@ -49,10 +48,6 @@ type GenesisBackend interface {
 	GenesisTime() (math.U64, error)
 }
 
-type RandaoBackend interface {
-	RandaoAtEpoch(slot math.Slot, epoch math.Epoch) (common.Bytes32, error)
-}
-
```
abi87: I am not really we need there shotgun backend methods.
My idea is that:

- Logic should be in handlers, not backend
- Backend should contain the minimum common logic shared across handlers

node-api/handlers/beacon/genesis_test.go
```diff
@@ -134,8 +134,7 @@ func TestGetGenesis(t *testing.T) {
 
 			// setup test
 			backend := mocks.NewBackend(t)
-			h := beacon.NewHandler(backend)
-			h.SetLogger(noop.NewLogger[log.Logger]())
```
abi87, line 138: This is one of the points of this PR. Just pass the logger immediately in the handler constructor instead of passing the loggers right after

node-api/handlers/beacon/handler.go
```diff
@@ -31,12 +32,11 @@ type Handler struct {
...
+func NewHandler(backend Backend, logger log.Logger) *Handler {
 	h := &Handler{
-		BaseHandler: handlers.NewBaseHandler(
-			handlers.NewRouteSet(""),
-		),
-		backend: backend,
+		BaseHandler: handlers.NewBaseHandler(logger),
+		backend:     backend,
 	}
+	registerRoutes(h)
```
abi87, line 40: This is the other point of this  PR:

- Each handler embeds a `BaseHandler` which contains all the routes for the handler.
  - These routes are exposed to the middleware to duly direct every request
- Route registration requires passing the handler method needed to process the transaction. So there is a bit of circular dependency where:
  - BaseHandler is created when handler is created
  - Routes can be passed to the embedded base handler only after the Handler has been created

I cannot get rid of the circular dependency, but I encapsulated this into the handler constructor, rather than genereting the routing mapping ex post.

node-api/server/server.go
```diff
@@ -42,6 +42,12 @@ import (
 	cmtcfg "github.com/cometbft/cometbft/config"
 )
 
+// handler is an interface that all handlers must implement.
+// to let middleware register their routes.
+type handler interface {
```
abi87: moved here from BaseHandler since it is used only here really

node-api/handlers/routes.go
```diff
@@ -22,8 +22,14 @@ package handlers
 
 import (
 	"github.com/berachain/beacon-kit/log"
+	"github.com/labstack/echo/v4"
 )
 
+type Context = echo.Context
```
abi87, line 28: just moved here from node-api/handlers/handlers.go because there objects are used here

node-api/handlers/handlers.go
```diff
@@ -49,23 +35,27 @@ type BaseHandler struct {
...
+func (b *BaseHandler) NotImplemented(c Context) (any, error) {
+	endpoint := c.Request().URL.Path
+	return nil, fmt.Errorf("endpoint %s is not yet implemented", endpoint)
 }
 
 // Deprecated handles deprecated API endpoints that are no longer supported according to the
 // Ethereum Beacon Node API spec.
 func (b *BaseHandler) Deprecated(c Context) (any, error) {
 	endpoint := c.Request().URL.Path
-	return nil, NewHTTPError(http.StatusInternalServerError, "The endpoint %s is deprecated.", endpoint)
```
abi87, line 68: These errors are wrapped in the proper Http error code by the middleware response. They really don't make sense to me here.
Tested both with a couple of unimplemented/deprecated API and I see the proper error code being propagated to the client

node-api/handlers/beacon/routes.go
```diff
@@ -23,15 +23,11 @@ package beacon
...
-	"github.com/berachain/beacon-kit/log"
 	"github.com/berachain/beacon-kit/node-api/handlers"
 )
 
 //nolint:funlen // routes are long
-func (h *Handler) RegisterRoutes(
-	logger log.Logger,
-) {
-	h.SetLogger(logger)
+func registerRoutes(h *Handler) {
```
abi87, line 30: unexported and made a free function, since this does not need to be called anymore outside of handlers (in `Server`), but it is only called in handler constructors

node-api/handlers/handlers.go
```diff
@@ -21,25 +21,11 @@
...
 
-	"github.com/berachain/beacon-kit/errors"
 	"github.com/berachain/beacon-kit/log"
-	"github.com/labstack/echo/v4"
 )
 
-type Context = echo.Context
-
-// handlerFn enforces a signature for all handler functions.
-type handlerFn func(c Context) (any, error)
```
abi87, line 34: moved to routes.go

node-api/handlers/handlers.go
```diff
@@ -21,25 +21,11 @@
...
-
-// Handlers is an interface that all handlers must implement.
-type Handlers interface {
-	// RegisterRoutes is a method that registers the routes for the handler.
-	RegisterRoutes(logger log.Logger)
-	RouteSet() *RouteSet
-}
```
abi87, line 41: simplified and moved to server file, since it is used only there

node-api/handlers/handlers.go
abi87: Will defer a full cleanup to a upcoming PR where I can also extend unit tests and assert that the right error is returned

node-api/server/server.go
```diff
@@ -74,23 +80,21 @@ func New(
...
+	handlers = append(handlers, debugapi.NewHandler(b, apiLogger))
+	handlers = append(handlers, eventsapi.NewHandler(apiLogger))
+	handlers = append(handlers, nodeapi.NewHandler(b, apiLogger))
+	handlers = append(handlers, proofapi.NewHandler(b, apiLogger))
 
 	for _, handler := range handlers {
-		handler.RegisterRoutes(apiLogger)
-		mware.RegisterRoutes(handler.RouteSet(), apiLogger)
+		mware.RegisterRoutes(handler.RouteSet())
 	}
```
fridrik01: nit: we can now skip handlers slice and for loop and instead to sth like:

```
	// instantiate handlers and register their routes in the middleware
	b := backend.New(storageBackend, cs, cmtCfg, consensusService)

	mware := middleware.NewDefaultMiddleware(apiLogger)
	mware.RegisterRoutes(beaconapi.NewHandler(b, apiLogger).RouteSet())
	mware.RegisterRoutes(builderapi.NewHandler(apiLogger).RouteSet())
	mware.RegisterRoutes(configapi.NewHandler(b, apiLogger).RouteSet())
	mware.RegisterRoutes(debugapi.NewHandler(b, apiLogger).RouteSet())
	mware.RegisterRoutes(eventsapi.NewHandler(apiLogger).RouteSet())
	mware.RegisterRoutes(nodeapi.NewHandler(b, apiLogger).RouteSet())
	mware.RegisterRoutes(proofapi.NewHandler(b, apiLogger).RouteSet())
```

node-api/handlers/handlers.go
```diff
@@ -49,23 +35,27 @@ type BaseHandler struct {
...
 	}
 }
 
 // NotImplemented is the handler for API endpoints that are defined in the Ethereum Beacon Node API
 // spec, but not yet implemented.
-func (b *BaseHandler) NotImplemented(Context) (any, error) {
-	return nil, errors.New("not implemented")
+func (b *BaseHandler) NotImplemented(c Context) (any, error) {
+	endpoint := c.Request().URL.Path
+	return nil, fmt.Errorf("endpoint %s is not yet implemented", endpoint)
```
fridrik01, line 51: 👍

PR #2908, opened 2025-09-02
chore(node-api): fix requested slot to state slot mapping

abi87:


node-api/handlers/utils/slot_mapping.go
```diff
@@ -18,26 +18,27 @@
...
+	reqSlot, err := SlotFromStateID(stateID, b)
 	if err != nil {
-		return common.Bytes32{}, errors.Wrapf(err, "failed to get state from slot %d", slot)
+		return nil, math.Slot(0), err
 	}
-
-	// Infer the epoch if not provided.
-	if epoch == 0 {
-		epoch = b.cs.SlotToEpoch(resolvedSlot)
+	st, stateSlot, err := b.StateAtSlot(reqSlot)
```
abi87, line 39: These two operation are basically always done in sequence, but in current codebase they are often not side by side, and they are split among handlers and backend.
The main focus of this PR is to put them together.
Once this is done we can extend the semantics of `SlotFromStateID`, let it return an int and handle differently genesis and head, finally

PR #2913, opened 2025-09-04
fix(consensus: fix optimistic payload UT following pectra1 activation

abi87:
Some UTs:

- reference mainnet specs
- do use time.Now() to create blocks

in which case they happen to call `processElectra1Fixes` and err if they don't find Smilee validator.
This PR appeases those UTs by warning instead of erring if the validator is not there.

While this is a consensus change, I believe this is safe since:

- these code path could happen only in Mainnet and
- the fork already successfully happened in Mainnet, so we are sure that error paths where not executed nor should be ever executed due to determinism.


rezbera (approved):
makes sense given appHash will happen and act as a fast fail for the unlikely scenario

fridrik01 (approved):
lgtm

PR #2917, opened 2025-09-08
Node api more validators apis to handler

abi87:


fridrik01 (approved):
lgtm, and solid readability improvements.

PR #2920, opened 2025-09-08
chore(node-api): more API migration - 3

abi87:
- Moved `GetPendingPartialWithdrawals` logic from node-api backend to handler
- Moved `GetRandao` logic from node-api backend to handler
- Moved `Specs` logic from node-api backend to handler
- Moved `BlockHeaderAtSlot` logic from node-api backend to handler


node-api/backend/randao.go
```diff
@@ -1,43 +0,0 @@
...
-
-package backend
-
-import (
-	"github.com/berachain/beacon-kit/errors"
-	"github.com/berachain/beacon-kit/primitives/common"
-	"github.com/berachain/beacon-kit/primitives/math"
-)
-
-func (b *Backend) RandaoAtEpoch(slot math.Slot, epoch math.Epoch) (common.Bytes32, error) {
```
fridrik01, line 29: An interface and backend for essentially a modulo operation, its much nicer to handle this in the handler implementation like its done now!

PR #2926, opened 2025-09-10
chore(all): upgrade to v.1.25 + govulcheck + tenv replacement

abi87:


.golangci.yaml
```diff
@@ -297,8 +297,8 @@ linters:
     - spancheck # checks for mistakes with OpenTelemetry/Census spans
     - sqlclosecheck # checks that sql.Rows and sql.Stmt are closed
     - stylecheck # is a replacement for golint
-    - tenv # detects using os.Setenv instead of t.Setenv since Go1.17
```
abi87, line 300: this is deprecated in favour of `usertesting`. Replacing it fixes the warning message we got when running `make golangci-fix`

cli/commands/genesis/storage_test.go
```diff
@@ -37,10 +37,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+//nolint:paralleltest // t.TempDir() do not allow parallelism
 func TestSetDepositStorageCmd(t *testing.T) {
-	t.Parallel()
```
abi87, line 41: Changes in the testing codebase are all related to `usetesting` replacement of `tenv`

scripts/build/codegen.mk
```diff
@@ -12,7 +12,7 @@ generate: ## generate all the code
 		(cd $$module && \
 			GETH_PKG_INCLUDE=$(GETH_PKG_INCLUDE) go generate ./...) || exit 1; \
 	done
-	@go run github.com/vektra/mockery/v2@v2.49.0
+	@go run github.com/vektra/mockery/v2@v2.53.5
```
abi87, line 15: Needed to run with go v.1.25

scripts/build/linting.mk
```diff
@@ -79,6 +79,17 @@ gosec: gosec-install
...
+#################
+#   vulncheck   #
+#################
+
+vulncheck-install:
+	@go install golang.org/x/vuln/cmd/govulncheck@latest
+
+vulncheck: vulncheck-install
+	@echo "--> Running govulncheck"
+	@govulncheck $(shell go list ./... | grep -v '/testing/')
```
abi87, line 91: excluded testing files since there are a couple of vulnerabilities with Kurtosis. To be fixed

consensus-types/types/deposit_message_test.go
```diff
@@ -46,7 +46,7 @@ func TestCreateAndSignDepositMessage(t *testing.T) {
 		0x01, 0x00, 0x00, 0x00,
 	}
 
-	mocksSigner := &mocks.BLSSigner{}
```
fridrik01, line 49: why this change?

scripts/build/linting.mk
fridrik01: Cool, its checking against known vulnerabilities reported and confirmed by Go security team and stored in https://vuln.go.dev/. It has over 2k reported vulnerabilities in the database (https://pkg.go.dev/vuln/list)

Dockerfile
```diff
@@ -46,7 +46,7 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
 ###         Stage 2 - Build the Application         ###
 #######################################################
 
-FROM golang:${GO_VERSION}-alpine3.20 AS builder
```
fridrik01, line 49: We should check that bumping the version here does not cause us to link against newer libc library causing things not to work on older distributions due to libc errors

consensus-types/types/deposit_message_test.go
abi87: The new version of mockery (older does not compile with go v.1.25), has a different naming scheme.

calbera (approved):
utACK

Dockerfile
calbera: Agree before merging, its worth making docker build and running/testing against it

abi87: @camembera got `beacond` and `bera-geth` syncing on debian 11, 12, 13 and ubuntu 20.04, 22.04, 25.04

PR #2928, opened 2025-09-10
chore(testing): introduce EL prague2 fork

calbera:
For EL fork details, see https://github.com/berachain/bera-geth/pull/89

### DIFF
--- a/testing/simulated/el-genesis-files/eth-genesis.json
+++ b/testing/simulated/el-genesis-files/eth-genesis.json
@@ -37,7 +37,7 @@
         "polDistributorAddress": "0x4200000000000000000000000000000000000042"
       },
       "prague2": {
-        "time": 0,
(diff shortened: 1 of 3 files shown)


rezbera (approved):

testing/simulated/el-genesis-files/eth-genesis.json
```diff
@@ -35,6 +35,10 @@
         "baseFeeChangeDenominator": 48,
         "minimumBaseFeeWei": 10000000000,
         "polDistributorAddress": "0x4200000000000000000000000000000000000042"
+      },
+      "prague2": {
+        "time": 0,
```
rezzmah: time needs to be after or equal to prague1 or bera-reth fails. Changed to same as prague1 for all sim tests @abi87

PR #2930, opened 2025-09-11
chore(node-api): serve genesis data

abi87:
- [X] Recreate genesis state when node-api starts and use it to serve genesis related information
- [X] Return genesis state when genesis is requested, instead of returning block 1 state.
- [X] Ensure state thread-safety

### DIFF
--- a/node-api/backend/backend.go
+++ b/node-api/backend/backend.go
@@ -60,7 +60,7 @@ type Backend struct {
 	// Genesis related data
 	sp           GenesisStateProcessor // only needed to recreate genesis state upon API loading
 	db           dbm.DB                // in-memory DB to store genesis state
-	muCms        sync.RWMutex          // mutex to allow copying genesisState in a thread safe way
+	muSt         sync.RWMutex          // mutex to allow initializing and copying genesisState in a safe way
 	cms          storetypes.CommitMultiStore
 	genesisState *state.StateDB // caches genesis data to serve API requests
 }
@@ -106,33 +106,23 @@ func (b *Backend) Close() error {
 	return b.db.Close()
 }
 
-// If checkChainIsReady returns nil, we assume genesis state is created
-func (b *Backend) checkChainIsReady() error {
-	switch err := b.node.IsAppReady(); {
-	case err == nil:
-		// chain finally ready, time to loading genesis
-		return b.loadGenesisState()
-	case errors.Is(err, cometbft.ErrAppNotReady):
-		return cometbft.ErrAppNotReady
-	default:
-		return fmt.Errorf("unable to check whether app is ready: %w", err)
-	}
-}
-
 type backendKVStoreService struct {
 	ctx sdk.Context
 }
 
 func (kvs *backendKVStoreService) OpenKVStore(context.Context) corestore.KVStore {
-	//nolint:contextcheck // fine with tests
+	//nolint:contextcheck // fine with this node-api backend
 	store := sdk.UnwrapSDKContext(kvs.ctx).KVStore(backendStoreKey)
 	return kvstorage.NewKVStore(store)
 }
 
-//nolint:gochecknoglobals // todo: fix later
+//nolint:gochecknoglobals // fine with this node-api backend
 var backendStoreKey = storetypes.NewKVStoreKey("backend-genesis")
 
 func (b *Backend) loadGenesisState() error {
+	b.muSt.Lock()
+	defer b.muSt.Unlock()
+
 	if b.genesisState != nil {
 		// genesis state already initialized, we're fine
 		return nil
--- a/node-api/handlers/beacon/header.go
+++ b/node-api/handlers/beacon/header.go
@@ -23,6 +23,7 @@ package beacon
 import (
 	"errors"
 	"fmt"
+	stdmath "math"
 
 	"github.com/berachain/beacon-kit/node-api/handlers"
 	beacontypes "github.com/berachain/beacon-kit/node-api/handlers/beacon/types"
@@ -42,7 +43,7 @@ func (h *Handler) GetBlockHeaders(c handlers.Context) (any, error) {
 	switch {
 	case len(req.Slot) == 0 && len(req.ParentRoot) == 0:
 		// no parameter specified, pick chain HEAD
-		// by requesting special height 0.
+		// by requesting special height -1.
 		height := utils.Head
 		return h.makeBlockHeaderResponse(height, true /*resultsInList*/)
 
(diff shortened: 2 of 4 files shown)


node-core/components/interfaces.go
```diff
@@ -382,7 +383,7 @@ type (
 		GetTotalSlashing() (math.Gwei, error)
 		GetNextWithdrawalIndex() (uint64, error)
 		GetNextWithdrawalValidatorIndex() (math.ValidatorIndex, error)
-		GetTotalValidators() (uint64, error)
+		GetTotalValidators() (math.U64, error)
```
abi87, line 386: just realized this was left behind.
Not even sure this is needed rn

consensus/cometbft/service/node_api_support.go
```diff
@@ -25,11 +25,27 @@ import (
...
+	ErrAppNotReady = fmt.Errorf("%s is not ready; please wait for first block", AppName)
+
+	_ types.ConsensusService = (*Service)(nil)
+)
+
+// IsAppReady is used mainly by node-apis to understand the chain state,
+// before loading data
+func (s *Service) IsAppReady() error {
+	// use custom query multi-store if provided
+	lastBlockHeight := s.sm.GetCommitMultiStore().LatestVersion()
```
abi87, line 43: minor code duplication with `CreateQueryContext` (which uses `lastBlocHeight`), but this makes a simpler API for other services (node-api) notably to understand whether chain is ready or not.
Likely matters only if genesis time is set in the future, but still

node-api/backend/backend.go
```diff
@@ -63,75 +85,141 @@ func New(
...
-
-	// If using height 0 for the query context, make sure to return the latest slot.
-	if slot == 0 {
-		slot, err = st.GetSlot()
-		if err != nil {
-			return st, slot, fmt.Errorf("GetSlot failed: %w", err)
-		}
-	}
-	return st, slot, nil
+	return b.db.Close()
```
abi87, line 106: for an in-memory DB, `Close` does not really do anything, but I like having clean places to create and release resources

node-api/backend/backend.go
```diff
@@ -55,7 +85,141 @@ func New(
...
+
+func (b *Backend) Close() error {
+	if b.db == nil {
+		return nil
+	}
+	return b.db.Close()
+}
+
+// If checkChainIsReady returns nil, we assume genesis state is created
+func (b *Backend) checkChainIsReady() error {
```
fridrik01: naming indicates a readonly method, consider renaming it since it mutates states when chain  ready.

node-api/backend/backend.go
```diff
@@ -55,7 +85,141 @@ func New(
...
+		return fmt.Errorf("unable to check whether app is ready: %w", err)
+	}
+}
+
+type backendKVStoreService struct {
+	ctx sdk.Context
+}
+
+func (kvs *backendKVStoreService) OpenKVStore(context.Context) corestore.KVStore {
+	//nolint:contextcheck // fine with tests
```
fridrik01:
```suggestion
func (kvs *backendKVStoreService) OpenKVStore(_ context.Context) corestore.KVStore {
```

node-api/backend/backend.go
```diff
@@ -55,7 +85,141 @@ func New(
...
+	// Note: we process genesis here as soon as node start, but
+	// chain would wait for genesisTime to come if genesisTime
+	// is set in the future. We replicate this behaviour with checkChainIsReady.
+	if _, err = b.sp.InitializeBeaconStateFromEth1(
+		b.genesisState,
+		genesisData.GetDeposits(),
+		genesisData.GetExecutionPayloadHeader(),
+		genesisData.GetForkVersion(),
+	); err != nil {
+		return fmt.Errorf("failed processing genesis: %w", err)
```
fridrik01, line 154: Since we are using an in-mem db its probably not an issue, but could we have a partial initialization problem if we return an error here since we ran `initGenesisState` successfully?

node-api/backend/interface.go
```diff
@@ -0,0 +1,62 @@
...
+	InitializeBeaconStateFromEth1(
+		st *statedb.StateDB,
+		deposits ctypes.Deposits,
+		execPayloadHeader *ctypes.ExecutionPayloadHeader,
+		genesisVersion common.Version,
+	) (transition.ValidatorUpdates, error)
+}
+
+// Keep just getters currently used. To be expanded as we increase API endpoints available
+type ReadOnlyBeaconState interface {
```
fridrik01, line 42: Did you consider creating an immutable concrete type wrapper instead of interface?

node-api/handlers/beacon/header.go
```diff
@@ -42,38 +42,41 @@ func (h *Handler) GetBlockHeaders(c handlers.Context) (any, error) {
 	switch {
 	case len(req.Slot) == 0 && len(req.ParentRoot) == 0:
 		// no parameter specified, pick chain HEAD
-		// by requesting special slot 0.
-		slot := utils.Head
-		return h.makeBlockHeaderResponse(slot, true /*resultsInList*/)
+		// by requesting special height 0.
```
fridrik01: Should it not say: 

```suggestion
		// by requesting special height -1.
```

node-api/handlers/beacon/header.go
```diff
@@ -42,38 +42,41 @@ func (h *Handler) GetBlockHeaders(c handlers.Context) (any, error) {
...
 
 	case len(req.Slot) == 0 && len(req.ParentRoot) != 0:
-		parentSlot, errParent := utils.SlotFromBlockID(req.ParentRoot, h.backend)
+		parentSlot, errParent := utils.BlockIDToHeight(req.ParentRoot, h.backend)
 		if errParent != nil {
 			return nil, fmt.Errorf("%w, failed retrieving parent root with error: %w", handlertypes.ErrNotFound, errParent)
 		}
-		slot := parentSlot + 1
-		return h.makeBlockHeaderResponse(slot, true /*resultsInList*/)
+		// TODO ABENEGIA: what happens here if HEAD is requested. Should be already broken??
```
fridrik01: ABENEGIA?

node-api/handlers/beacon/validators_filters.go
```diff
@@ -42,8 +42,8 @@ var errStatusFilterMismatch = errors.New("validator status does not match status
 // FilterValidators is a helper function to provide implementation
 // consistency between GetStateValidators and PostStateValidators, since they
 // are intended to behave the same way.
-func (h *Handler) FilterValidators(slot math.Slot, ids []string, statuses []string) ([]*beacontypes.ValidatorData, error) {
-	st, resolvedSlot, err := h.backend.StateAtSlot(slot)
+func (h *Handler) FilterValidators(height int64, ids []string, statuses []string) ([]*beacontypes.ValidatorData, error) {
```
fridrik01, line 45: Maybe dumb question but is there some rule where we use height and where slots?

node-core/components/interfaces.go
fridrik01: not sure I follow, `GetTotalValidators` is used in the codebase  so not sure if you can remove it?

node-api/backend/interface.go
fridrik01: nevermind, I like it like this

node-api/handlers/beacon/validators_filters.go
abi87: So the API can requests slots in a few ways:

- by specifying the slot
- by indicating it wants HEAD chain, whatever slot that is
- by indicating a block hash

We used to handle the second case, request HEAD, by passing a slot 0. This is not good anymore, since I want to use 0 for genesis. So I thought to use -1 to indicate the tip, similarly to what python does to indicate last element of an array.

So I introduce `height int64` to indicate a not-yet-well-identified slot and `slot uint64` as the actual slot.

This is be best I could come up with to indicate the HEAD. I considered using `math.Uint64` but it is not as natural to me as `-1` (although it would spare us the type conversion headache).

Definitely open to suggestions here

node-api/backend/backend.go
abi87: I don't think so.
- If genesis processing fails upon startup, then `services.Start` would fail hence stopping the binary
- if it fails during an API call, it must have failed when chain started, which would have stopped the node

I think in both cases, if we reuse the state processor, we should be fine

node-core/components/interfaces.go
abi87: The return type in the interface was wrong and as soon as I tried using this exact interface in `node-api`s, I got a compilation error.
Since this never happened before, it must be that this interface is not used, or at least not fully used.

node-api/backend/backend.go
abi87: Done

node-api/handlers/beacon/header.go
abi87: Done, thanks for spotting this

node-api/handlers/beacon/header.go
abi87: Left over a check. Added

node-api/backend/backend.go
abi87: Indeed, dropped and move its implementation to the only place where it was used

node-api/handlers/beacon/validators_filters.go
fridrik01: agree, besides using special string tags like "latest" I think -1 is best to represent HEAD

fridrik01 (approved):
utACK

PR #2937, opened 2025-09-22
fix: resolve JWT authentication race condition on validator restart

fridrik01:
This commit addresses a race condition that caused JWT authentication failures when validators restarted after extended downtime. 

I noticed this when I was testing blob distribution where I stopped a single validator for an extended period of time (over 1hour) and then restarted it expecting it to fetch blobs from other peers, instead it caused a JWT token invalid error.

The issue occurred because the RPC client's JWT refresh was happening asynchronously in a go routine while the main thread immediately attempted to make RPC calls with a expired token.

I added also logging to the JWT refresh mechanics.

### DIFF
--- a/execution/client/ethclient/rpc/client.go
+++ b/execution/client/ethclient/rpc/client.go
@@ -29,6 +29,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/berachain/beacon-kit/log"
 	"github.com/berachain/beacon-kit/primitives/encoding/json"
 	beaconhttp "github.com/berachain/beacon-kit/primitives/net/http"
 	"github.com/berachain/beacon-kit/primitives/net/jwt"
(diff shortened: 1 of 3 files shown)


execution/client/ethclient/rpc/client.go
```diff
@@ -94,22 +100,30 @@ func (rpc *client) Start(ctx context.Context) {
 	ticker := time.NewTicker(rpc.jwtRefreshInterval)
 	defer ticker.Stop()
 
-	if err := rpc.updateHeader(); err != nil {
-		panic(err)
-	}
```
fridrik01, line 99: Start() is called in a gouroutine, now the caller calls `Initialize()` before running Start() guaranteeing a valid token

rezbera (approved):
nice

PR #2938, opened 2025-09-22
Blob p2p distribution using Reactor API

### LINKED ISSUE #2665: Improve blobs distribution
Blobs are currently distributed as transactions in a CometBFT block.
This scheme is simple because taps into Comet facility to distribute data without requiring any code on BeaconKit side (other than blobs validation).
However it has a major defect: it does not allow to properly cleanup blobs. Blobs have a finite lifespan and are supposed to be cleaned up after a couple of weeks from their creation; however, as CometBFT transaction, a blob will never be cleaned up and will be distributed to any node requiring the CometBFT block.

We should devise a different scheme to distribute blobs that does not force CometBFT to store blobs forever.

See [this draft](https://docs.google.com/document/d/14CzIgtE9H2_r-VgJFGs-FC71HTRGG_NC72ROIvuYOVs/edit?tab=t.0#heading=h.826bexokgb93) for the proposed solution (soon to be made public)

fridrik01:
issue: https://github.com/berachain/beacon-kit/issues/2665
depends on cometbft blob PR: https://github.com/berachain/cometbft/pull/26

**Context**
Currently, blobs are stored as as CometBFT transaction so a blob will never be cleaned up and will take up wasted space since they have a finite lifespan. 

This PR addresses this by implementing a complete p2p blob distribution system using the cometbft [Reactor API](https://docs.cometbft.com/v0.38/spec/p2p/reactor-api/) which fetches missing blobs from other peers. It introduces a `BlobConsensusEnableHeight` configuration parameter that defines when the chain transitions from storing blobs as transactions to using P2P distribution which requires a hard fork.

**Components**
- `BlobReactor` (in `da/blobreactor/)` that handles p2p blob requests/responses with timeout handling and makes sure received blobs pass verification (from potential byzantine peers). 
- `BlobFetcher` background service (in `beacon/blockchain/`) that manages a persistent filesystem based retry queue. When a block arrives with missing blobs, the request is queued and the background worker attempts to fetch from peers at regular interval with a max retry count. The queue is crash-safe with atomic writes, includes request deduplication, and automatically cleans up requests that exceed retry limits or fall outside the DA availability window.

**Consensus flow (after blob enable height reached)**
- `PrepareProposal`: Blobs included in the block are now returned in `PrepareProposalResponse` in a separate `Blob` field instead of as CometBFT transactions. This prevents blobs from being persisted in CometBFTs block store.
- `ProcessProposal`: Blobs are now included in the `ProcessProposalRequest` in a separate `Blob` field and are validated synchronously. Proposals are rejected if blobs are missing or invalid. Validated blobs are cached with the state for FinalizeBlock.
- `FinalizeBlock`: If node is in consensus mode, then blobs come from the `ProcessProposal` cache and are processed immediately. Otherwise, if the node is syncing (catching up) and comes upon a block that should have blobs the node will make async fetch request to `BlobFetcher`. `FinalizeBlock` returns immediately without blocking, allowing the chain to continue syncing while blobs are fetched in the background.

TODO:
- [x] Add peer scoring to avoid bad peers (implemented in see https://github.com/berachain/beacon-kit/pull/2938) 
- [x] Evaluate if throttling is needed  (implemented in see https://github.com/berachain/beacon-kit/pull/2938) 
- [x] Update RPC endpoint to return "blobs pending" if user requests blobs for a block that are still queued in blobreactor
- [ ] Update cometbft dependency
- [ ] Evaluate if we need to split blobs into parts like cometbft does
- [ ] Evaluate if we need to enable/disable Blobreactor when not needed


go.mod
```diff
@@ -3,8 +3,8 @@ module github.com/berachain/beacon-kit
 go 1.25
 
 replace (
-	github.com/cometbft/cometbft => github.com/berachain/cometbft v1.0.1-0.20250718121541-9671b1dc07e5
-	github.com/cometbft/cometbft/api => github.com/berachain/cometbft/api v1.0.1-0.20250718121541-9671b1dc07e5
+	github.com/cometbft/cometbft => github.com/alesforz/cometbft v0.0.0-20250916204208-b48cac03c4e5
+	github.com/cometbft/cometbft/api => github.com/alesforz/cometbft/api v0.0.0-20250916204208-b48cac03c4e5
```
calbera, line 7: is the plan to merge this into the berachain/cometbft fork? We should probably do it before going ahead with this.

Also can you link the cometbft changes/PR here for reference as well?

consensus/cometbft/service/encoding/encoding.go
```diff
@@ -93,24 +65,75 @@ func UnmarshalBeaconBlockFromABCIRequest(
...
+
+		var sidecars datypes.BlobSidecars
+		if err := ssz.Unmarshal(sidecarBz, &sidecars); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal blobs from Txs[1]: %w", err)
+		}
+
+		return sidecars, nil
+	}
+
+	// After BlobEnableHeight: blobs are distributed via BlobReactor and must be retrieved from cache or storage
```
calbera:
```suggestion
	// At or after BlobEnableHeight: blobs are distributed via BlobReactor and must be retrieved from cache or storage
```

consensus/cometbft/service/encoding/encoding.go
```diff
@@ -93,24 +65,75 @@ func UnmarshalBeaconBlockFromABCIRequest(
...
-// UnmarshalBlobSidecarsFromABCIRequest extracts blob sidecars from an ABCI
-// request.
-func UnmarshalBlobSidecarsFromABCIRequest(
+// UnmarshalBlobSidecarsFromABCIRequest extracts blob sidecars from an ABCI FinalizeBlockRequest based on blob consensus parameters.
+func UnmarshalBlobSidecarsFromABCIRequest(req *cmtabci.FinalizeBlockRequest, blobEnableHeight int64) (datypes.BlobSidecars, error) {
+	if req == nil {
+		return nil, ErrNilABCIRequest
+	}
+
+	// Before or at BlobEnableHeight: blobs are in Txs[1] if present
```
calbera:
```suggestion
	// Before BlobEnableHeight: blobs are in Txs[1] if present
```

consensus/cometbft/service/blobreactor/config.go
```diff
@@ -0,0 +1,76 @@
...
+	BlobConsensusUpdateHeight() int64
+	// BlobConsensusEnableHeight returns the height when P2P blob distribution via BlobReactor is enabled.
+	// A value of 0 means the BlobReactor is disabled.
+	BlobConsensusEnableHeight() int64
+	// BlobMaxBytes returns the maximum size of blob data in bytes for BlobReactor consensus params.
+	BlobMaxBytes() int64
+}
+
+// Config contains configuration for the P2P BlobReactor component.
+type Config struct {
```
calbera, line 52: nit: this doesnt implement `BlobConsensusUpdateHeight()` ?

consensus/cometbft/service/prepare_proposal.go
```diff
@@ -84,7 +84,11 @@ func (s *Service) prepareProposal(
 		return &cmtabci.PrepareProposalResponse{Txs: [][]byte{}}, nil
 	}
 
-	return &cmtabci.PrepareProposalResponse{
-		Txs: [][]byte{blkBz, sidecarsBz},
-	}, nil
+	if s.chainSpec.BlobConsensusEnableHeight() <= 0 || req.Height < s.chainSpec.BlobConsensusEnableHeight() {
```
calbera: Can we make a helper function for this that returns bool? I see these comparisons quite often and makes reading a little confusing

go.mod
fridrik01: @calbera yes this is a workaround until we merge the blob distribution in github.com/alesforz/cometbft into our fork. @jmalicevic is working on getting it merged, hopefully in the next few days.

consensus/cometbft/service/blobreactor/config.go
fridrik01: Config doesn't need getters, removed them.

consensus/cometbft/service/prepare_proposal.go
fridrik01: good idea, it was replicated in 7 places, lol

consensus/cometbft/service/finalize_block.go
```diff
@@ -43,6 +49,22 @@ func (s *Service) finalizeBlock(
 		return nil, err
 	}
 
+	getBlobsFunc := func(cachedBlobData []byte) (datypes.BlobSidecars, error) {
```
abi87, line 52: It seems to me this function can be simplified as follow:

```
	if len(cachedBlobData) != 0 {
		var sidecars datypes.BlobSidecars
		if err := ssz.Unmarshal(cachedBlobData, &sidecars); err != nil {
			return nil, fmt.Errorf("finalize block: failed to unmarshal cached blob data: %w", err)
		}
		return sidecars, nil
	}

	// not cached
	sidecars, err := encoding.UnmarshalBlobSidecarsFromABCIRequest(req, s.chainSpec)
	if err != nil {
		return nil, fmt.Errorf("finalize block: failed parsing blobs from request: %w", err)
	}
	return sidecars, nil
```

`UnmarshalBlobSidecarsFromABCIRequest` already does check `IsBlobConsensusEnabledAtHeight`, so no need to do it twice

abi87: another nit: maybe we can declare thing function as service method? Not sure why it is declared inside `finalizeBlock`. Maybe to make sure it is used only in finalized and no where else?

abi87: question for self: can we cache sidecars instead of blobs, to avoid any ops in case blob is cached? I guess sizes of the two objects matters too, we'd like to cache the smallest

consensus/cometbft/service/finalize_block.go
```diff
@@ -113,7 +150,14 @@ func (s *Service) finalizeBlock(
 		return nil, fmt.Errorf("failed checking cached final state, hash %s, height %d: %w", hash, req.Height, err)
 	}
 
-	valUpdates, err := s.Blockchain.FinalizeBlock(finalState.Context(), req)
+	var cachedBlobData []byte
```
abi87: @fridrik01 : I see `cachedBlobData` is cached below, but isn't this always an empty slice?
Whatever change is done in  `getBlobsFunc`, `cachedBlobData` won't be affected as it is passed via value, right?

Check the snippet https://go.dev/play/p/iNPBeyfhtBb

consensus/cometbft/service/blobreactor/config.go
```diff
@@ -0,0 +1,69 @@
...
+
+const (
+	// Height at which consensus params are upgraded to use blobreactor
+	blobConsensusUpdateHeight int64 = 0
+
+	// Height to enable blobreactor
+	blobConsensusEnableHeight int64 = math.MaxInt64
+
+	// maxBytes is the maximum size of blob data in bytes for BlobReactor consensus params (e.g., 800KB).
+	blobMaxBytes int64 = 800 * 1024 // 800KB
```
abi87, line 33: note for self: surprised we had to define this quantity. Don't we already enforce explicitly a max size for blobs? If not shouldn't we express it in terms of other quantities?

beacon/blockchain/finalize_block.go
```diff
@@ -89,26 +102,45 @@ func (s *Service) FinalizeSidecars(
 	blk *ctypes.BeaconBlock,
 	blobs datypes.BlobSidecars,
 ) error {
+	processBlobsFunc := func(blobs datypes.BlobSidecars) error {
```
abi87, line 105: Consider rewriting the function as follow. It marginally reduces indentation and feels clearer flow to me

```
	processBlobsFunc := func(blobs datypes.BlobSidecars) error {
		// Process the blobs which saves them to storage
		if err := s.blobProcessor.ProcessSidecars(s.storageBackend.AvailabilityStore(), blobs); err != nil {
			return fmt.Errorf("failed to process %d blob sidecars for slot %d: %w", len(blobs), blk.GetSlot(), err)
		}

		// Final verification
		if !s.storageBackend.AvailabilityStore().IsDataAvailable(ctx, blk.GetSlot(), blk.GetBody()) {
			return fmt.Errorf("data not available after processing blobs for slot %d: %w", blk.GetSlot(), ErrDataNotAvailable)
		}

		return nil
	}

	// SyncingToHeight is always the tip of the chain both during sync and when
	// caught up. We don't need to process sidecars unless they are within DA period.
	//
	//#nosec: G115 // SyncingToHeight will never be negative.
	if !s.chainSpec.WithinDAPeriod(blk.GetSlot(), math.Slot(syncingToHeight)) {
		// Here outside Data Availability window. Just log if needed
		if len(blobs) > 0 {
			s.logger.Info(
				"Skipping blob processing outside of Data Availability Period",
				"slot", blk.GetSlot().Base10(), "head", syncingToHeight,
			)
		}
		return nil
	}

	// here within DA period. Any blob to process
	expectedBlobs := len(blk.GetBody().GetBlobKzgCommitments())
	if expectedBlobs == 0 {
		return nil // No blobs expected for this block
	}

	// If blobs were passed in (normal consensus mode), use them
	if len(blobs) > 0 {
		return processBlobsFunc(blobs)
	}

	// Queue the blob fetch request to be handled asynchronously
	err := s.blobFetcher.QueueBlobRequest(blk.GetSlot(), blk)
	if err != nil {
		// TODO: should we log and continue here instead of erroring out?
		return fmt.Errorf("failed to queue blob fetch request for slot %d: %w", blk.GetSlot().Unwrap(), err)
	}

	// Return immediately without waiting for blobs
	// The background fetcher will handle retrieval and storage
	return nil

```

beacon/blockchain/finalize_block.go
```diff
@@ -89,26 +102,45 @@ func (s *Service) FinalizeSidecars(
...
-		) {
-			return ErrDataNotAvailable
+		// Check the block to see if there should be blobs
+		expectedBlobs := len(blk.GetBody().GetBlobKzgCommitments())
+		if expectedBlobs == 0 {
+			return nil // No blobs expected for this block
 		}
+
+		// Queue the blob fetch request to be handled asynchronously
+		err := s.blobFetcher.QueueBlobRequest(blk.GetSlot(), blk)
```
abi87: nit: slot parameter seems redundant here, since it comes from the block? Can't we just pass the block?

beacon/blockchain/blob_queue.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	if len(files) == 0 {
+		return BlobFetchRequest{}, "", errNoMoreRequests
+	}
+
+	// Update queue depth metric with current file count
+	q.metrics.setQueueDepth(len(files))
+
+	for _, filename := range files {
```
abi87, line 129: IIUC, files are named after the slot.
Question: would there be any advantage in sorting the files in a specific way? IDK if the os takes a specific order by default or its random (we should assume the latter as it may depends on OSes, versions etc).
I wonder if e.g. requesting newer slots first or older first has any benefit for the network as a whole

beacon/blockchain/blob_queue.go
```diff
@@ -0,0 +1,212 @@
...
+			q.logger.Warn("Request is outside availability window, deleting",
+				"slot", request.Header.Slot.Unwrap(),
+				"head_slot", headSlot.Unwrap(),
+				"failure_count", request.FailureCount)
+			_ = q.Remove(filename)
+			continue
+		}
+
+		// Check if request has exceeded max retry limit
+		if request.FailureCount >= maxRetries {
```
abi87, line 162: question: why should there be a max number of retrials?
There blobs are valid (a supermajority approved them when they were finalized), so there should always be a supermajority of non-byzantine nodes which can serve them. Why stopping and not trying forever?

beacon/blockchain/blob_queue.go
fridrik01: The glob function in Go's standard library explicitly calls slices.Sort on filenames. Changing to a different ordering (e.g., newest-first or random) would I think complicate debugging without clear benefits, as we'd lose the predictable sequential processing order.

beacon/blockchain/blob_fetcher.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// DefaultBlobFetcherConfig returns the default configuration.
+//
+//nolint:mnd // Just defaults
+func DefaultBlobFetcherConfig() BlobFetcherConfig {
+	return BlobFetcherConfig{
+		CheckInterval: 10 * time.Second,
+		RetryInterval: 1 * time.Minute,
+		MaxRetries:    10,
```
abi87, line 52: I wonder if there should be a cap here. What happens at cap? A node gives up on syncing?

beacon/blockchain/blob_fetcher.go
```diff
@@ -0,0 +1,224 @@
...
+			storageBackend: storageBackend,
+			logger:         logger,
+		},
+	}, nil
+}
+
+// Start begins the background blob fetching process.
+func (bf *blobFetcher) Start(ctx context.Context) {
+	bf.ctx, bf.cancel = context.WithCancel(ctx)
+	go bf.run()
```
abi87, line 110: unlike Stop, this is not idempotent. Should it be?

beacon/blockchain/blob_queue.go
abi87: I agree, but we'll have to monitor for possible performances gains

abi87 (commented):
a few questions but overall LGTM

beacon/blockchain/blob_fetcher.go
fridrik01: If cap is reached, the blob request is deleted from disk, meaning that blob won't exist on this node. The node always continues syncing regardless and other blob requests are processed independently.  If/when this happens, we record a `blob_fetcher_requests_expired_total{reason="max_retries"})` metric so operators can detect if this happens frequently. Alternatively we can keep this uncapped and instead do something like double retryInterval (exponential backoff) each time it fails. Eventually it will pass the DA window be cleaned up.

beacon/blockchain/blob_fetcher.go
fridrik01: Probably not, kept it simple since `depositCatchupFetcher` isn't idempotent either and is called in the same Service.Start() method  (https://github.com/berachain/beacon-kit/blob/blobreactor/beacon/blockchain/service.go#L113-L116). Happy to add `sync.Once` to both if we think it's worth it.

PR #2940, opened 2025-09-25
use atomic pointer for fcu state

calbera:


beacon/blockchain/execution_engine.go
```diff
@@ -51,11 +51,8 @@ func (s *Service) sendPostBlockFCU(
 		FinalizedBlockHash: lph.GetParentHash(),
 	}
 
-	s.muLatestFcuReq.Lock()
-	latestRequestedFCU := s.latestRequestedFCU
-	s.latestRequestedFCU = engineprimitives.ForkchoiceStateV1{} // reset and prepare for next block
-	s.muLatestFcuReq.Unlock()
-
+	latestRequestedFCU := s.latestFcuReq.Load()
+	s.latestFcuReq.Store(&engineprimitives.ForkchoiceStateV1{}) // reset and prepare for next block
```
calbera, line 55: is this reset even necessary? I'm not entirely convinced it is

abi87: I think it's cleaner like this. True that a failed comparison would cause the FCU to be called (so we should err on the correct side). Still I really like the idea of knowing the lifespan of a value.
Any bug due to stale data is generally a nightmare to debug

PR #2943, opened 2025-09-29
chore(CometBFT): study custom network topology - DO NOT MERGE

abi87:
NEVER MERGE THIS TO MAINNET

- Enforce a specific network topology across validator, to study latencies distirbutions.
  - I find this complex to do with infra configs, so resorting to playing with BeaconKit. I would:
    - Assume the entire list of validators is passed to each node as persistent_peers
    - Assume that laddr of a node is set, even if that is a private address
    - Then code would build a tree of validators and reshape persistent peers list and a few other P2P configs to enforce the topology 

TODOs:

- [X] Enforce tree like network topology. Add UTs
- [ ] Add persistent peers list to e2e tests


abi87:
Run some tests with @fridrik01 and realized configs like:

- `peer_gossip_sleep_duration`
- `peer_query_maj23_sleep_duration`
- `max_packet_msg_payload_size`
- `send_rate` (possibly)
- `recv_rate` (possibly)

have a dominat role in determining tail latency. Topology does a small difference (if any)

PR #2944, opened 2025-09-30
[1/n] chore(blockchain): dropped `enable_optimistic_payload_builds` flag

abi87:
Disable optimistic block building as a config, and let it be the one way we build payloads.

Note: a few `simulated` tests fails when enforcing optimistic block building. For some of the UTs the matter is changing the tests to account for the different way next payload timestamp is set; other uncovers genuine bugs. Below the list of broken tests, if they have been fixed:

- [X] TestTimestampFork_ELAndCLInSync_IsSuccessful, FIXED
- [X] TestProcessProposal_InvalidTimestamps_Errors, FIXED thanks to @rezbera 
- [X] TestMaliciousUser_MakesConsolidationRequest_IsIgnored FIXED
- [X] TestValidProposer_ProposesPostForkBlockIsNotFinalized_IsSuccessful FIXED
- [X] TestFullLifecycle_WithFullWithdrawalRequest_IsSuccessful, FIXED
- [X] TestFullLifecycle_WithPartialWithdrawalRequests_IsSuccessful, FIXED
- [X] TestExcessValidatorBeforeFork_CorrectlyEvicted, FIXED
- [X] TestWithdrawalFromExcessStake_WithPartialWithdrawal_CorrectAmountWithdrawn FIXED


beacon/blockchain/execution_engine.go
```diff
@@ -30,10 +30,7 @@ import (
 )
 
 // sendPostBlockFCU sends a forkchoice update to the execution client after a
-// block is finalized.This function should only be used to notify
-// the EL client of the new head and should not request optimistic builds, as:
-// Optimistic clients already request builds in handleOptimisticPayloadBuild()
-// Non-optimistic clients should never request optimistic builds.
```
abi87, line 36: IMO this comment is not clear nor really informative. Please correct me if I am wrong/ you disagree

beacon/blockchain/metrics.go
```diff
@@ -66,8 +66,8 @@ func (cm *chainMetrics) markRebuildPayloadForRejectedBlockSuccess(
 }
 
 // markRebuildPayloadForRejectedBlockFailure increments the counter for the
-// number of times
-// the validator failed to build an optimistic payload due to a failure.
+// number of times the validator failed to build an optimistic payload
```
abi87, line 69: just formatting

beacon/blockchain/payload_test.go
```diff
@@ -61,14 +61,13 @@ import (
...
-	optimisticPayloadBuilds := true // key to this test
 	cs, err := spec.MainnetChainSpec()
 	require.NoError(t, err)
 
-	chain, st, _, ctx, _, b, sb, eng, depStore := setupOptimisticPayloadTests(t, cs, optimisticPayloadBuilds)
+	chain, st, _, ctx, _, b, sb, eng, depStore := setupOptimisticPayloadTests(t, cs)
 	sb.EXPECT().StateFromContext(mock.Anything).Return(st)
 	sb.EXPECT().DepositStore().RunAndReturn(func() deposit.StoreManager { return depStore })
-	b.EXPECT().Enabled().Return(optimisticPayloadBuilds)
+	b.EXPECT().Enabled().Return(true)
```
abi87, line 70: not: enabled only tells whether node should build block (must be true for validators, should be false for full nodes).
Given the tests are about optimistic block building, block building must be enabled

testing/networks/80094/app.toml
```diff
@@ -185,10 +185,6 @@ payload-timeout = "850ms"
 # Graffiti string that will be included in the graffiti field of the beacon block.
 graffiti = ""
 
-# EnableOptimisticPayloadBuilds enables building the next block's payload optimistically in
-# process-proposal to allow for the execution client to have more time to assemble the block.
-enable-optimistic-payload-builds = "false"
```
abi87, line 190: testing note: AFAIK a few validators already set this to true on mainnet

testing/networks/80069/app.toml
```diff
@@ -185,10 +185,6 @@ payload-timeout = "850ms"
 # Graffiti string that will be included in the graffiti field of the beacon block.
 graffiti = ""
 
-# EnableOptimisticPayloadBuilds enables building the next block's payload optimistically in
-# process-proposal to allow for the execution client to have more time to assemble the block.
-enable-optimistic-payload-builds = "false"
```
abi87, line 190: testing note: this has been set to true on Bepolia for some time

abi87:
Note: `TestTimestampFork_ELAndCLInSync_IsSuccessful` seems to be failing because, given optimistic block building, the 5th payload gets built with a fork set to `version.Deneb1()`, while timestamp of its comet block maps to `version.Electra()` fork.
The block rejection is a known. Working on fixing the UT

testing/simulated/pectra_fork_test.go
```diff
@@ -186,20 +190,25 @@ func (s *PectraForkSuite) TestTimestampFork_ELAndCLInSync_IsSuccessful() {
 		"Processing block with fork version service=blockchain\u001B[0m block=2\u001B[0m fork=0x04010000\u001B[0m",
 		"Processing block with fork version service=blockchain\u001B[0m block=3\u001B[0m fork=0x04010000\u001B[0m",
 		"Processing block with fork version service=blockchain\u001B[0m block=4\u001B[0m fork=0x04010000\u001B[0m",
-		"Processing block with fork version service=blockchain\u001B[0m block=5\u001B[0m fork=0x05000000\u001B[0m",
-		"Processing block with fork version service=blockchain\u001B[0m block=6\u001B[0m fork=0x05000000\u001B[0m",
-		"Processing block with fork version service=blockchain\u001B[0m block=7\u001B[0m fork=0x05000000\u001B[0m",
+		"Processing block with fork version service=blockchain\u001B[0m block=5\u001B[0m fork=0x04010000\u001B[0m",
```
abi87, line 193: so now time is controlled by payload time, built optimistically (as we enforce in the test that consensusTime == payloadTime). PayloadTime increases slower than before, so we need more blocks to get to the fork

testing/simulated/pectra_fork_test.go
```diff
@@ -186,20 +190,25 @@ func (s *PectraForkSuite) TestTimestampFork_ELAndCLInSync_IsSuccessful() {
...
-	startHeight := int64(1)
-	iterations := int64(len(expectedMessages))
-	expectedMessagesIdx := 0
-	submitTxNonce := uint64(0)
+	var (
+		startHeight         = int64(1)
+		iterations          = int64(len(expectedMessages))
+		expectedMessagesIdx = 0
+		submitTxNonce       = uint64(0)
+		consensusTime       = time.Unix(startHeight*2, 0)
```
abi87, line 207: Minor rearrangement due to optimistic block building. Set time for next block at the end of the for loop, instead of setting time for current block at the start of the loop

testing/simulated/pectra_fork_test.go
```diff
@@ -224,7 +233,23 @@ func (s *PectraForkSuite) TestTimestampFork_ELAndCLInSync_IsSuccessful() {
...
+		// the timestamp of the payload built optimistically.
+		forkVersion := s.Geth.TestNode.ChainSpec.ActiveForkVersionForTimestamp(math.U64(consensusTime.Unix())) //#nosec: G115
+		blk, _, err := encoding.ExtractBlobsAndBlockFromRequest(
+			processRequest,
+			blockchain.BeaconBlockTxIndex,
+			blockchain.BlobSidecarsTxIndex,
+			forkVersion,
+		)
+		s.Require().NoError(err)
+		consensusTime = time.Unix(
```
abi87, line 249: this replicates the way to set payload optimistically

testing/simulated/malicious_proposer_test.go
```diff
@@ -149,53 +148,6 @@ func (s *SimulatedSuite) TestProcessProposal_BadBlock_IsRejected() {
 	s.Require().Contains(s.LogBuffer.String(), "max fee per gas less than block base fee: address 0x20f33CE90A13a4b5E7697E3544c3083B8F8A51D4, maxFeePerGas: 10000000, baseFee: 765625000")
 }
 
-// TestProcessProposal_InvalidTimestamps_Errors effectively serves as a test for how a valid node would react to
-// a malicious proposer attempting to use a future timestamp in the block that does not match the consensus timestamp.
-func (s *SimulatedSuite) TestProcessProposal_InvalidTimestamps_Errors() {
```
abi87, line 154: I don't see how can we keep this test with optimistic block building on, which is a bummer.

Without optimistic block building, consensusTime sets payloadTime, so it's relative easy to make payloadInvalid by changing consensusTime among block building and block verification.

With optimistic block building, payloadTime is set by the consensus layer deterministally based on previous block consensus time and payload time. Obviously, we set it so that it always passes verification.

To be clear, note that the payloadTime check we do is
```
bound := max(
	consensusTime+1,
	parentPayloadTimestamp+1,
)
assert(payloadTimestamp <= bound)
```
So:
- if we maliciously increase consensusTime, we push forward the time bound so easier to pass the check
- if we maliciously decrease consensusTime, bound is set by parentPayloadTimestamp and due to the deterministic formula we use in optimistic block building, bound cannot be violated

testing/simulated/utils.go
```diff
@@ -155,7 +158,21 @@ func (s *SharedAccessors) MoveChainToHeight(
 		proposedCometBlocks = append(proposedCometBlocks, proposal)
 		finalizedResponses = append(finalizedResponses, finalizeResp)
 
-		proposalTime = proposalTime.Add(time.Duration(s.TestNode.ChainSpec.TargetSecondsPerEth1Block()) * time.Second)
+		// set consensus time for the next block to match
```
abi87, line 161: minor change to keep proposalTime == payloadTime, as it was without optimistic block building

testing/simulated/utils.go
```diff
@@ -155,7 +158,20 @@ func (s *SharedAccessors) MoveChainToHeight(
...
+		// the timestamp of the payload built optimistically.
+		forkVersion := s.TestNode.ChainSpec.ActiveForkVersionForTimestamp(math.U64(proposalTime.Unix())) //#nosec: G115
```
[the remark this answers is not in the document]
abi87: I believe this cannot practically happens, and this is testing code. So fine leaving this as is

testing/simulated/utils.go
```diff
@@ -128,14 +130,15 @@ func (s *SharedAccessors) MoveChainToHeight(
...
 			Txs:             proposal.Txs,
 			Height:          currentHeight,
 			ProposerAddress: pubkey.Address(),
 			Time:            proposalTime,
-		})
+		}
+		processResp, err := s.SimComet.Comet.ProcessProposal(s.CtxComet, processReq)
 		require.NoError(t, err)
-		require.Equal(t, types.PROCESS_PROPOSAL_STATUS_ACCEPT, processResp.Status)
+		require.Equal(t, types.PROCESS_PROPOSAL_STATUS_ACCEPT.String(), processResp.Status.String())
```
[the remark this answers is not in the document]
abi87: Actually the string comparison improves the error message if the `require.Equal` fails. Without string I'd have a comparison among two enums

testing/simulated/malicious_proposer_test.go
rezzmah: An alternative to removing is to call PrepareProposal twice on the malicious proposer which will force a payload rebuild with the malicious timestamp, e.g. the below test passes. I would prefer this to removing.

```go
// TestProcessProposal_InvalidTimestamps_Errors effectively serves as a test for how a valid node would react to
// a malicious proposer attempting to use a future timestamp in the block that does not match the consensus timestamp.
func (s *SimulatedSuite) TestProcessProposal_InvalidTimestamps_Errors() {
	const blockHeight = 1
	const coreLoopIterations = 1

	// Initialize the chain state.
	s.InitializeChain(s.T())

	// Retrieve the BLS signer and proposer address.
	blsSigner := simulated.GetBlsSigner(s.HomeDir)
	pubkey, err := blsSigner.GetPubKey()
	s.Require().NoError(err)

	// Test happens post Deneb1 fork.
	startTime := time.Now()

	// Go through 1 iteration of the core loop to bypass any startup specific edge cases such as sync head on startup.
	proposals, _, correctConsensusTime := s.MoveChainToHeight(s.T(), blockHeight, coreLoopIterations, blsSigner, startTime)
	s.Require().Len(proposals, coreLoopIterations)
	currentHeight := int64(blockHeight + coreLoopIterations)

	// Prepare a block proposal. This will create a valid payload due to optimistic payload building.
	// It is called to flush the payload cache.
	validProposal, err := s.SimComet.Comet.PrepareProposal(s.CtxComet, &types.PrepareProposalRequest{
		Height:          currentHeight,
		Time:            correctConsensusTime,
		ProposerAddress: pubkey.Address(),
	})
	s.Require().NoError(err)
	s.Require().NotEmpty(validProposal)

	// We PreparePayload again to force the payload to rebuild with the malicious proposal time, as the previous
	// malicious proposal was created optimistically with a valid proposal time.
	maliciousProposalTime := correctConsensusTime.Add(2 * time.Second)
	maliciousProposal, err := s.SimComet.Comet.PrepareProposal(s.CtxComet, &types.PrepareProposalRequest{
		Height:          currentHeight,
		Time:            maliciousProposalTime,
		ProposerAddress: pubkey.Address(),
	})
	s.Require().NoError(err)
	s.Require().NotEmpty(maliciousProposal)

	// Reset the log buffer to discard old logs we don't care about
	s.LogBuffer.Reset()
	// Process the proposal containing the malicious block.
	processResp, err := s.SimComet.Comet.ProcessProposal(s.CtxComet, &types.ProcessProposalRequest{
		Txs:             maliciousProposal.Txs,
		Height:          currentHeight,
		ProposerAddress: pubkey.Address(),
		// Use the correct time as the actual consensus time, which mismatches the proposal time.
		Time: correctConsensusTime,
	})
	s.Require().NoError(err)
	s.Require().Equal(types.PROCESS_PROPOSAL_STATUS_REJECT, processResp.Status)
	s.Require().Contains(s.LogBuffer.String(), payloadtime.ErrTooFarInTheFuture.Error())
}
```

testing/simulated/pectra_fork_test.go
```diff
@@ -348,6 +374,8 @@ func (s *PectraForkSuite) TestMaliciousUser_MakesConsolidationRequest_IsIgnored(
 func (s *PectraForkSuite) TestValidProposer_ProposesPostForkBlockIsNotFinalized_IsSuccessful() {
 	// Initialize the chain state.
 	s.Geth.InitializeChain(s.T())
+	s.Reth.InitializeChain(s.T()) // helper to build "invalid" blocks
+	helperBuilder := s.Reth
```
calbera, line 378: confused about this. I see the test's passing now, but I'm unsure how using this reth node (which with this configuration has the same validator pubkey as the geth node) to build blocks make this test pass

abi87: So the idea is to use the helper to build the "invalid"  payload. We never verify/finalize the blocks on the helper AND the built block is the first (so no optimistic block building already called) so we have a way better control on timestamps then we had on a node where optimistic block building has already being called.

testing/simulated/malicious_proposer_test.go
abi87: Awesome, reinserted!

testing/simulated/pectra_fork_test.go
calbera: Ah makes sense. In #2949, I was able to clean it up a bit to get it working with just the 1 node https://github.com/berachain/beacon-kit/pull/2949/files#r2427840098

PR #2947, opened 2025-10-08
chore(block-building): Block build path cleanup

abi87:


beacon/validator/block_builder.go
```diff
@@ -220,17 +202,11 @@ func (s *Service) retrieveExecutionPayload(
 	slot := slotData.GetSlot()
 	envelope, err := s.localPayloadBuilder.RetrievePayload(ctx, slot, parentBlockRoot)
 	if err == nil {
-		return envelope, nil
+		nextPayloadTimestamp := envelope.GetExecutionPayload().GetTimestamp()
+		return envelope, s.stateProcessor.ProcessFork(st, nextPayloadTimestamp, false)
 	}
 
 	// If we failed to retrieve the payload, request a synchronous payload.
-	//
```
abi87, line 227: outdated/meaningless comment

beacon/validator/block_builder.go
```diff
@@ -220,17 +202,11 @@ func (s *Service) retrieveExecutionPayload(
 	slot := slotData.GetSlot()
 	envelope, err := s.localPayloadBuilder.RetrievePayload(ctx, slot, parentBlockRoot)
 	if err == nil {
-		return envelope, nil
+		nextPayloadTimestamp := envelope.GetExecutionPayload().GetTimestamp()
+		return envelope, s.stateProcessor.ProcessFork(st, nextPayloadTimestamp, false)
```
abi87, line 224: So this is actually not a pure refactoring. I believe this is the only code-path where ProcessFork was not called after ProcessSlots, so this change fixes that.
Note that we are not guessing the timestamp here: fork is determined by the timestamp from the payload, as expected.

payload/attributes/factory.go
```diff
@@ -1,72 +0,0 @@
...
-
-// BuildPayloadAttributes creates a new instance of PayloadAttributes.
-func (f *Factory) BuildPayloadAttributes(
-	timestamp math.U64,
-	payloadWithdrawals engineprimitives.Withdrawals,
-	prevRandao common.Bytes32,
-	prevHeadRoot common.Root,
-	parentProposerPubkey *crypto.BLSPubkey,
-) (*engineprimitives.PayloadAttributes, error) {
-	return engineprimitives.NewPayloadAttributes(
```
abi87, line 63: this does so little that we can just drop it all

beacon/validator/block_builder.go
abi87: cc @calbera is this another bug?

PR #2948, opened 2025-10-08
testing(simulated): optimistic edge cases

calbera:
This PR was originally put out just to introduce `optimistic_test.go` and show how in some cases, optimistic block building may delay block finalization.
There after was modified by @abi87 to improve payloadID cache cleanup, which diverged from the solution proposed by @calbera  in https://github.com/berachain/beacon-kit/pull/2949.

It looks like the tests introduced in this PR can be made green simply by being more conservative in purging payloadID cache.
Normally an EL client that has received `FCU(Head == block_height_N)` and has started building a payload, won't allow to re-build a payload via `FCU(Head == block_height_H)`, where `height_H <= height_N`.
This would cause the tests `TestReth_RebuildPayload_IsSuccessful` and `TestGeth_RebuildPayload_IsSuccessful`.

Current PR fixes this by being more conservative in cleaning up payloadID cache. Specifically payloadIDs are removed in two cases:

- if `getPayload` returns specific errors (`-38001` code) which signal that payloadID in unknown (as it happens when it becomes stale
- when chain moves on, as it already is implemented via pruning

The PR splits `GetAndEvict` method into `Get` and `Delete` and call `Delete` only in the case above, while `Get` does not consume the payloadID.

### DIFF
--- a/testing/simulated/pectra_fork_test.go
+++ b/testing/simulated/pectra_fork_test.go
@@ -372,13 +372,15 @@ func (s *PectraForkSuite) TestMaliciousUser_MakesConsolidationRequest_IsIgnored(
 // The next round will propose a valid pre-fork block that gets finalized due to deviance in the consensus timestamp.
 // The proposer will then propose a valid post-fork block that is correctly finalized.
 func (s *PectraForkSuite) TestValidProposer_ProposesPostForkBlockIsNotFinalized_IsSuccessful() {
+	client := s.Geth
+	helper := s.Reth
+
 	// Initialize the chain state.
-	s.Geth.InitializeChain(s.T())
-	s.Reth.InitializeChain(s.T()) // helper to build "invalid" blocks
-	helperBuilder := s.Reth
+	client.InitializeChain(s.T())
+	helper.InitializeChain(s.T()) // helper to build "invalid" blocks
 
 	// Retrieve the BLS signer and proposer address.
-	blsSigner := simulated.GetBlsSigner(s.Geth.HomeDir)
+	blsSigner := simulated.GetBlsSigner(client.HomeDir)
 	pubkey, err := blsSigner.GetPubKey()
 	s.Require().NoError(err)
 
@@ -388,13 +390,13 @@ func (s *PectraForkSuite) TestValidProposer_ProposesPostForkBlockIsNotFinalized_
 	// 1 - Build a block whose consensus and payloadTimestamp are both post-fork.
 	//     Check that it verifies, but do not finalize it
 	{
-		consensusTime := time.Unix(int64(s.Geth.TestNode.ChainSpec.ElectraForkTime()), 0)
+		consensusTime := time.Unix(int64(client.TestNode.ChainSpec.ElectraForkTime()), 0)
 		prepareReq := &types.PrepareProposalRequest{
 			Height:          nextBlockHeight,
 			Time:            consensusTime,
 			ProposerAddress: pubkey.Address(),
 		}
-		proposal, err = helperBuilder.SimComet.Comet.PrepareProposal(helperBuilder.CtxComet, prepareReq)
+		proposal, err = helper.SimComet.Comet.PrepareProposal(helper.CtxComet, prepareReq)
 		s.Require().NoError(err)
 		s.Require().Len(proposal.Txs, 2)
 
@@ -406,8 +408,8 @@ func (s *PectraForkSuite) TestValidProposer_ProposesPostForkBlockIsNotFinalized_
 		}
 
 		// Process the proposal
-		s.Geth.LogBuffer.Reset()
-		processResp, respErr := s.Geth.SimComet.Comet.ProcessProposal(s.Geth.CtxComet, processRequest)
+		client.LogBuffer.Reset()
+		processResp, respErr := client.SimComet.Comet.ProcessProposal(client.CtxComet, processRequest)
 		s.Require().NoError(respErr)
 		s.Require().Equal(types.PROCESS_PROPOSAL_STATUS_ACCEPT.String(), processResp.Status.String())
 	}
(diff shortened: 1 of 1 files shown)


testing/simulated/optimistic_test.go
```diff
@@ -0,0 +1,207 @@
...
+//   - reth EL builds payload for height N
+//   - reth accepts process proposal (N, i)
+//   - marking N - 1 as "finalized" to the EL
+//   - supermajority does not finalize (N, i)
+//   - block height N, next round i + 1
+//   - **BUG** reth EL does NOT store the already built payload for height N, so tries rebuilding
+//   - any EL who accepted proposal (N, i) returns empty (comet block has 0 txs when it should have 2)
+//     for proposal (N, i + 1) because engine API returns "Received nil payload ID on VALID engine response"
+//   - since N - 1 is "finalized" to the EL, asking to re-build payload N by sending FCU with
+//     N - 1 as "head" block, EL always returns nil.
```
calbera, line 47: Note that this can be overriden on reth with

```
--engine.always-process-payload-attributes-on-canonical-head
          Always process payload attributes and begin a payload build process even if `forkchoiceState.headBlockHash` is already the canonical head or an ancestor. See `TreeConfig::always_process_payload_attributes_on_canonical_head` for more details.
```

More details https://github.com/paradigmxyz/reth/pull/16676

calbera: This change should make this test pass. Its not a fix for the identified bug ` **BUG** reth EL does NOT store the already built payload for height N, so tries rebuilding`, but it does practically solve this edge case

testing/simulated/optimistic_test.go
```diff
@@ -0,0 +1,207 @@
...
+			Time:            consensusTime,
+		}
+		_, finalizeErr := s.Reth.SimComet.Comet.FinalizeBlock(s.Reth.CtxComet, finalizeRequest)
+		s.Require().NoError(finalizeErr)
+		_, commitErr := s.Reth.SimComet.Comet.Commit(s.Reth.CtxComet, &types.CommitRequest{})
+		s.Require().NoError(commitErr)
+	}
+}
+
+// This tests a reth validator proposing a block. It then accepts the proposal in
```
abi87: typo: geth validator

testing/simulated/optimistic_test.go
```diff
@@ -0,0 +1,207 @@
...
+		processRequest := &types.ProcessProposalRequest{
+			Txs:             proposal.Txs,
+			Height:          nextBlockHeight,
+			ProposerAddress: pubkey.Address(),
+			Time:            consensusTime,
+		}
+		// This will trigger a optimistic payload build for block height 2.
+		processResp, respErr := s.Reth.SimComet.Comet.ProcessProposal(s.Reth.CtxComet, processRequest)
+		s.Require().NoError(respErr)
+		s.Require().Equal(types.PROCESS_PROPOSAL_STATUS_ACCEPT, processResp.Status)
```
abi87: nit: I know this is how we usually assert these checks but
```
s.Require().Equal(types.PROCESS_PROPOSAL_STATUS_ACCEPT.String(), processResp.Status.String())
```
reads better if there is an error.
Same for the assert below

testing/simulated/optimistic_test.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	// For some reason, the supermajority does not finalize the block.
+	// Next round is height 1, but simulating consensus time is 1 second after previous round.
+	time.Sleep(1 * time.Second)
```
abi87, line 88: can you clarify what is this sleep for? Is it to make sure payloadTime set optimistically is changed?
I am not sure is needed since payloadTime must be strictly greater than its parent

testing/simulated/optimistic_test.go
```diff
@@ -0,0 +1,207 @@
...
+	consensusTime = time.Unix(int64(s.Reth.TestNode.ChainSpec.ElectraForkTime())+1, 0)
```
abi87, line 89: nit: consider rewriting as `consensusTime = consensusTime.Add(time.Second)`

testing/simulated/optimistic_test.go
```diff
@@ -0,0 +1,207 @@
...
+		s.Require().Len(proposal.Txs, 2)
+
+		// Process the proposal.
+		processRequest := &types.ProcessProposalRequest{
+			Txs:             proposal.Txs,
+			Height:          nextBlockHeight,
+			ProposerAddress: pubkey.Address(),
+			Time:            consensusTime,
+		}
+		// This will trigger a optimistic payload build for block height 2.
```
abi87, line 80: self note:  this assumes this branch is built on top of `drop-enable_optimistic_payload_builds-flag` branch we can only build blocks optimistically, so no need to assert that optimistic block building is active.

testing/simulated/optimistic_test.go
```diff
@@ -0,0 +1,207 @@
...
+	consensusTime = time.Unix(int64(s.Reth.TestNode.ChainSpec.ElectraForkTime())+1, 0)
+	{
+		// Prepare the proposal.
+		proposal, prepareErr := s.Reth.SimComet.Comet.PrepareProposal(s.Reth.CtxComet, &types.PrepareProposalRequest{
+			Height:          nextBlockHeight,
+			Time:            consensusTime,
+			ProposerAddress: pubkey.Address(),
+		})
+		s.Require().NoError(prepareErr)
+		s.Require().Len(proposal.Txs, 2) // FAILS HERE
```
abi87, line 98: note for self: fails if `always-process-payload-attributes-on-canonical-head` is dropped

payload/builder/payload.go
```diff
@@ -132,7 +132,10 @@ func (pb *PayloadBuilder) RequestPayloadSync(
 		return nil, ctx.Err()
 	}
 
-	// Get the payload from the execution client.
+	// Get the payload from the execution client. Whether we successfully
+	// retrieve a valid payload or not, drop its payloadID from the cache
+	// as we should never use it again.
+	_, _ = pb.pc.GetAndEvict(r.Slot, r.ParentBlockRoot)
```
abi87: this is the change I believe we need to add to properly clean up the cache. This is the only code path where we retrieve the payload but we keep its payloadID. The propose change fixes that (and makes a few UTs failed as I expect they should)

payload/cache/payload_id.go
fridrik01: I don't get it, the comment says it does not evict payloadID from cache, but its being done here... hmm

PR #2949, opened 2025-10-09
feat(payload): more precise payload eviction for optimistic

calbera:
Continuing this work from https://github.com/berachain/beacon-kit/pull/2547 further 
> With the fix, the payload ID is removed from the beacon-kit cache after the first engine_getPayload call. Thus subsequent calls on secondary rounds will result in new payload builds getting started instead of attempting to use the same stale build.

We only want to evict payload IDs from our cache upon 3 conditions:
- We reject a process proposal that we built ourselves. -> We will never want to propose this same payload again.
  - We also should not evict if we are chosen to build a block for the next round (this happens if state root verification fails and we are the nextProposer). In this case do not evict.
- We finalize a block -> We will never need a payload for this height again.
- getPayload from the EL returns nil payload for `UnknownPayload -38001` -> We will not be able to retrieve this payload ever again (even if we want it); remove it now and force a rebuild down the line if needed.

Just the tests (and necessary sim test framework changes) without any logical changes can be found in #2952

### DIFF
--- a/beacon/blockchain/process_proposal.go
+++ b/beacon/blockchain/process_proposal.go
@@ -368,9 +368,9 @@ func (s *Service) shouldBuildNextPayload(isNextBlockProposer bool) bool {
 	return isNextBlockProposer && s.localBuilder.Enabled()
 }
 
-// evictLocalPayloadIfNecessary evicts the payload from local builder cache if necessary.
-// We deem it necessary only when we are rejecting the incoming block proposal and it has
-// been built by us.
+// evictLocalPayloadIfNecessary evicts the payload from local builder cache if necessary. We deem it
+// necessary only when we are rejecting the incoming block proposal and it has been built by us.
+// Also we ensure we don't evict any payload if we are chosen to propose for the next round.
 func (s *Service) evictLocalPayloadIfNecessary(
 	preState *statedb.StateDB,
 	processProposalErr *error,
@@ -387,7 +387,8 @@ func (s *Service) evictLocalPayloadIfNecessary(
 	// Check the conditions for evicting the local payload.
 	// - There must be a process proposal error (i.e. we are rejecting the proposal).
 	// - The proposer address must be the same as this node's address.
-	// - The process proposal error must not be due to rebuilding for a failed state transition.
+	// - The process proposal error must not be due to rebuilding for a failed state transition. We
+	//   don't want to evict the payload since it is being built async and will be used next round.
 	if *processProposalErr == nil || !bytes.Equal(thisNodeAddress, proposerAddress) ||
 		errors.Is(*processProposalErr, ErrRebuildForFailedStateTransition) {
 		return


testing/simulated/pectra_fork_test.go
```diff
@@ -374,13 +374,9 @@ func (s *PectraForkSuite) TestMaliciousUser_MakesConsolidationRequest_IsIgnored(
...
 	s.Geth.InitializeChain(s.T())
-	s.Reth.InitializeChain(s.T()) // helper to build "invalid" blocks
-	helperBuilder := s.Reth
-
-	// Retrieve the BLS signer and proposer address.
-	blsSigner := simulated.GetBlsSigner(s.Geth.HomeDir)
-	pubkey, err := blsSigner.GetPubKey()
+	gethNodeAddress, err := s.Geth.SimComet.GetNodeAddress()
 	s.Require().NoError(err)
+	s.Geth.SimComet.Comet.SetNodeAddress(gethNodeAddress)
```
calbera, line 379: this will allow the logic dependent on `thisNodeAddress` in the code to behave as intended

testing/simulated/pectra_fork_test.go
```diff
@@ -445,43 +441,16 @@ func (s *PectraForkSuite) TestValidProposer_ProposesPostForkBlockIsNotFinalized_
...
-	}
-
-	// The next block the proposer proposes with a pre-fork timestamp will actually have a pre-fork time
-	// Since the previous payload in cache has been evicted and a new payload is retrieved.
-	{
-		consensusTime := time.Unix(int64(s.Geth.TestNode.ChainSpec.ElectraForkTime())-2, 0)
-		prepareReq := &types.PrepareProposalRequest{
-			Height:          nextBlockHeight,
-			Time:            consensusTime,
-			ProposerAddress: pubkey.Address(),
```
calbera, line 475: the above deleted block is not necessary. Once the new pre-fork payload is built again and accepted process proposal, it can be now finalizeBlock and committed

testing/simulated/malicious_proposer_test.go
```diff
@@ -157,37 +157,24 @@ func (s *SimulatedSuite) TestProcessProposal_InvalidTimestamps_Errors() {
 
 	// Initialize the chain state.
 	s.InitializeChain(s.T())
-
-	// Retrieve the BLS signer and proposer address.
-	blsSigner := simulated.GetBlsSigner(s.HomeDir)
-	pubkey, err := blsSigner.GetPubKey()
+	nodeAddress, err := s.SimComet.GetNodeAddress()
 	s.Require().NoError(err)
+	s.SimComet.Comet.SetNodeAddress(nodeAddress)
```
calbera, line 162: this effectively enforces that optimistic builds do NOT happen in this test (since `MoveChainToHeight` does not set the `NextProposerAddress` field in `ProcessProposalRequest`s). This allows us to skip the extra PrepareProposal call on deleted line 176.

beacon/blockchain/process_proposal.go
```diff
@@ -62,10 +62,17 @@ func (s *Service) ProcessProposal(
 	req *cmtabci.ProcessProposalRequest,
 	thisNodeAddress []byte,
 ) (transition.ValidatorUpdates, error) {
-	signedBlk, sidecars, err := s.ParseBeaconBlock(req)
+	var err error
+	defer s.evictLocalPayloadIfNecessary(ctx, &err, thisNodeAddress, req.GetProposerAddress())
```
abi87: I am not sure this is the right place to do this.  The main reason is that I expect that if/what we evict depends on the output of `handleOptimisticPayloadBuild` and possibly `handleRebuildPayloadForRejectedBlock`, which happens asynchronously in their own go routine.

Consider moving this logic to `VerifyIncomingBlock`

node-core/components/interfaces.go
```diff
@@ -93,6 +93,8 @@ type (
 			ctx context.Context,
 			r *builder.RequestPayloadData,
 		) (ctypes.BuiltExecutionPayloadEnv, error)
+		// EvictPayload evicts a local payload from the cache if it exists.
+		EvictPayload(slot math.Slot, parentBlockRoot common.Root)
```
abi87, line 97: I am experimenting in https://github.com/berachain/beacon-kit/pull/2953 whether we can drop the `parentBlockRoot` side of the key. I am positive this is possible and this would simplify the code but maybe we don't want to mix concerns here.

payload/builder/interfaces.go
```diff
@@ -32,8 +32,9 @@ import (
 )
 
 type PayloadCache interface {
-	GetAndEvict(slot math.Slot, stateRoot common.Root) (cache.PayloadIDCacheResult, bool)
+	Get(slot math.Slot, stateRoot common.Root) (cache.PayloadIDCacheResult, bool)
 	Set(slot math.Slot, stateRoot common.Root, pid engineprimitives.PayloadID, version common.Version)
+	Delete(slot math.Slot, stateRoot common.Root)
```
abi87, line 37: This is the change I like the most in this PR

payload/builder/payload.go
```diff
@@ -53,7 +53,7 @@ func (pb *PayloadBuilder) RequestPayloadAsync(
 		return nil, common.Version{}, ErrPayloadBuilderDisabled
 	}
 
-	if payloadID, found := pb.pc.GetAndEvict(r.Slot, r.ParentBlockRoot); found {
+	if payloadID, found := pb.pc.Get(r.Slot, r.ParentBlockRoot); found {
```
abi87, line 58: so we don't evict here and let BeaconKit the opportunity to try and load again the payload if needed

payload/builder/payload.go
```diff
@@ -151,14 +151,17 @@ func (pb *PayloadBuilder) RetrievePayload(
...
+	payloadID, found := pb.pc.Get(slot, parentBlockRoot)
 	if !found {
 		return nil, ErrPayloadIDNotFound
 	}
 
 	// Get the payload from the execution client.
 	envelope, err := pb.getPayload(ctx, payloadID.PayloadID, payloadID.ForkVersion)
 	if err != nil {
+		// Evict the payload from cache as the EL can no longer retrieve it.
+		pb.EvictPayload(slot, parentBlockRoot)
```
abi87: I wonder if we should evict for some errors (like -38001) and still keep the payload for other?

Not that in the payloadID cache we still clear up payloadID that are too old (two heights behind), so there is no risk of leaks

testing/simulated/execution/reth.go
```diff
@@ -52,6 +52,7 @@ func defaultRethCmdStrBuilder(genesisFile string) string {
 			 --full \
 			 --engine.persistence-threshold=0 \
 			 --engine.memory-block-buffer-target=0 \
+			 --engine.always-process-payload-attributes-on-canonical-head \
```
abi87, line 55: @calbera do we really need to keep this? I assume that with the fixes on BeaconKit side we can drop this and have the "native" Reth/Geth when it comes to payload construction behaviour

beacon/blockchain/process_proposal.go
calbera: So we'll actually have to do it here because we need to handle errors to ParseBeaconBlock as well (in the sim tests we reject blocks from an invalid fork because parsing fails). 

Essentially we need to trigger this function if and only if we are rejecting the proposal (i.e. status == REJECT). That is the case iff this function returns error. I was thinking to move this 1 level higher instead of lower to [cometbft/service](https://github.com/berachain/beacon-kit/edit/precise-payload-eviction/consensus/cometbft/service/process_proposal.go#L73-L81) and have it call `blockchain.EvictLocalPayload` or smth of the sort

calbera: In what cases might eviction depend on `optimstic` or `rebuildPayload` routines?

testing/simulated/execution/reth.go
calbera: I think we do need to keep it and in fact recommend to be turned on in production (as other chains do as well). For the 2 tests in payload_cache_test.go with the comment `NOTE: this test requires reth with the --engine.always-process-payload-attributes-on-canonical-head flag`, you can see in which scenarios a reth node with this flag is necessary to "unstuck" the chain.

payload/builder/payload.go
calbera: I'm good with this. Change made in https://github.com/berachain/beacon-kit/pull/2949/commits/384157af5e56dfab64d2efdcbef9bdeed9969da0

payload/builder/payload.go
```diff
@@ -204,9 +216,6 @@ func (pb *PayloadBuilder) getPayload(
 	if err != nil {
 		return nil, err
 	}
-	if envelope == nil {
-		return nil, ErrNilPayloadEnvelope
-	}
```
calbera, line 209: nit: this error was never checked

execution/client/engine.go
```diff
@@ -145,7 +145,8 @@ func (s *EngineClient) GetPayload(
 		return result, s.handleRPCError(err)
 	}
 	if result == nil {
-		return result, engineerrors.ErrNilExecutionPayloadEnvelope
+		// Engine API returns the Unknown Payload (-38001) error if a nil result is returned.
+		return result, engineerrors.ErrUnknownPayload
```
calbera, line 149: `ErrNilExecutionPayloadEnvelope` error is never checked/used. To make more consistent with how the EL returns, for nil results also return the same error EL returns (unknown payload, -38001). 

Based on [how geth returns](https://github.com/berachain/bera-geth/blob/main/eth/catalyst/api.go#L504-L506), a nil result will always be accompanied by ErrUnknownPayload. So in beacon-kit we should actually never reach this case; it will be caught above on line 145 by `handleRPCError`

testing/simulated/execution/reth.go
abi87: I will double check once I rebase https://github.com/berachain/beacon-kit/pull/2956 on top of this PR if we still need it.

payload/builder/payload.go
abi87: Two things:

- it may be we inserted this if as guard for nilaway, to avoid accessing the envelope if nil. In such a case maybe we want to keep this
- I recall it used to be use, will double check histoyr

beacon/blockchain/process_proposal.go
```diff
@@ -354,8 +364,66 @@ func (s *Service) verifyStateRoot(
...
+
+	// Check the conditions for evicting the local payload.
+	// - There must be a process proposal error (i.e. we are rejecting the proposal).
+	// - The proposer address must be the same as this node's address.
+	if *processProposalErr == nil || !bytes.Equal(thisNodeAddress, proposerAddress) {
+		return
+	}
+
+	// Get the slot and parent block root from the state.
+	st := s.storageBackend.StateFromContext(ctx)
```
abi87: this is not enough to insulate state from any unnecessary modifications, and you need to add `copiedState := state.Copy(ctx)` and work on the copied state, similarly to what is done in `VerifyIncomingBlock` before `s.preFetchBuildData`

abi87: It's probably better to purge the `parentBlockRoot` from the cache altogether and be done with all of this.
I think `parentBlockRoot` is a relic of a past where we allowed empty blocks and basically forks. This is not the case anymore and for every slot there is a single `parentBlockRoot`

beacon/blockchain/process_proposal.go
```diff
@@ -354,8 +364,66 @@ func (s *Service) verifyStateRoot(
...
+		s.logger.Warn(
+			"Skipping payload eviction on rejected proposal",
+			"reason", "st.GetBlockRootAtIndex()",
+			"error", blockRootFetchErr,
+		)
+		return
+	}
+
+	// Evict the payload from the local builder cache.
+	s.localBuilder.EvictPayload(slot+1, parentBlockRoot)
```
abi87, line 429: I am not sure here why this is slot+1.
at this point of the code there is an error.
If the error comes before `VerifyIncomingBlock` then no optimistic block building has been requested, so this is moot.
However if the error comes from `VerifyIncomingBlock` it must be that we have asked the node to rebuild the block at `slot`, not `slot+1`?
In such a case, this is moot again?

I guess I am not sure I see what are we cleaning here?

payload/builder/payload.go
calbera: - nilaway still passes as is, the error is never really used
- ive consolidated nil payloads with a unilateral error that is also shared by the EL. see [here](https://github.com/berachain/beacon-kit/pull/2949#discussion_r2433189050) for details

beacon/blockchain/process_proposal.go
calbera: For incoming blk, `blk.GetSlot() == st.GetSlot()+1`. here `slot == st.GetSlot()`. So to purge the cache for the incoming block's height, we have to purge `slot+1`

abi87: Ok, so we are evicting a payload at the `slot == block_just_verified.Slot`, correct?

beacon/blockchain/process_proposal.go
calbera: added state copy https://github.com/berachain/beacon-kit/pull/2949/commits/e84b9b7bc02af3e8d50e83266e106d0ce4fbdeb5

testing/simulated/payload_cache_test.go
```diff
@@ -0,0 +1,1066 @@
...
+}
+
+// Test a scenario where reth must rebuild a payload for a failed state transition.
+//
+// NOTE: this test requires reth with the --engine.always-process-payload-attributes-on-canonical-head flag.
+func (s *PayloadCacheSuite) TestReth_MustRebuildForFailedStateTransition_IsSuccessful() {
+	// Initialize the chain state.
+	s.Reth.InitializeChain2Validators(s.T()) // 1 reth validator
+	s.Geth.InitializeChain2Validators(s.T()) // 1 geth validator
+
```
abi87, line 918: I would add something like

```
	testNode := s.Reth
	helpNode := s.Geth
```

and use them in the rest of the test. It simplifies playing with the test, in case we want to swap clients

testing/simulated/payload_cache_test.go
```diff
@@ -0,0 +1,1066 @@
...
+			NextProposerAddress: rethNodeAddress,
+		}
+		// Note: there is no payload cache eviction here because of 2 reasons. Either is sufficient
+		// to prevent eviction.
+		//  1) reth never had a payload built for block 1.
+		//  2) state root verification failed and we are chosen to propose block 1 next round.
+		processResp, processErr := s.Reth.SimComet.Comet.ProcessProposal(s.Reth.CtxComet, processRequest)
+		s.Require().NoError(processErr)
+		s.Require().Equal(types.PROCESS_PROPOSAL_STATUS_REJECT.String(), processResp.Status.String())
+		s.Require().Contains(s.Reth.LogBuffer.String(), "state root verification failed, rebuilding new block")
```
abi87, line 980: AWESOME! This is probably the best we can do to check whether optimistic block building kicked in!

testing/simulated/payload_cache_test.go
abi87: Something like
```
	// Initialize the chain state.
	testClient := s.Reth
	helperClient := s.Geth
	testClient.InitializeChain2Validators(s.T())
	helperClient.InitializeChain2Validators(s.T())

	rethNodeAddress, err := testClient.SimComet.GetNodeAddress()
	s.Require().NoError(err)
	gethNodeAddress, err := helperClient.SimComet.GetNodeAddress()
	s.Require().NoError(err)

	const blkHeight = int64(1)
	var (
		specs           = testClient.TestNode.ChainSpec
		consensusTime   = time.Unix(int64(specs.ElectraForkTime()), 0)
		validTxsHeight1 [][]byte
	)

	{
		// 1- Build a valid block at height 1, via the geth
		prepareRequest := &types.PrepareProposalRequest{
			Height:          blkHeight,
			Time:            consensusTime,
			ProposerAddress: gethNodeAddress,
		}
		proposal, prepareErr := helperClient.SimComet.Comet.PrepareProposal(helperClient.CtxComet, prepareRequest)
		s.Require().NoError(prepareErr)
		s.Require().Len(proposal.Txs, 2)
		validTxsHeight1 = proposal.Txs

		// 2- Process the block via reth node. The proposal is expected
		// to pass and start building payload for height 2, optimistically.
		processRequest := &types.ProcessProposalRequest{
			Txs:                 proposal.Txs,
			Height:              blkHeight,
			ProposerAddress:     gethNodeAddress,
			Time:                consensusTime,
			NextProposerAddress: rethNodeAddress,
		}
		processResp, respErr := testClient.SimComet.Comet.ProcessProposal(testClient.CtxComet, processRequest)
		s.Require().NoError(respErr)
		s.Require().Equal(types.PROCESS_PROPOSAL_STATUS_ACCEPT.String(), processResp.Status.String())
	}

	// For some reason, the supermajority does not finalize the block.
	// Another block comes, still at height 1, this time *invalid*. This would force
	// the node to rebuild height 1, which the EL cannot do since it has already received
	// an FCU(head == block_at_height_2, finalized == block_at_height_1)
	{
		invalidTxs := buildInvalidTestBlock(s, helperClient, validTxsHeight1, blkHeight, consensusTime)

		// 3- Process the invalid proposal proposal. It will be rejected
		// and attempt to build optimistically a block at height 1.
		processRequest := &types.ProcessProposalRequest{
			Txs:                 invalidTxs,
			Height:              blkHeight,
			ProposerAddress:     gethNodeAddress,
			Time:                consensusTime,
			NextProposerAddress: rethNodeAddress,
		}
		// Note: there is no payload cache eviction here because of 2 reasons. Either is sufficient
		// to prevent eviction.
		//  1) reth never had a payload built for block 1.
		//  2) state root verification failed and we are chosen to propose block 1 next round.
		processResp, processErr := testClient.SimComet.Comet.ProcessProposal(testClient.CtxComet, processRequest)
		s.Require().NoError(processErr)
		s.Require().Equal(types.PROCESS_PROPOSAL_STATUS_REJECT.String(), processResp.Status.String())
		s.Require().Contains(testClient.LogBuffer.String(), "state root verification failed, rebuilding new block")
	}

	{
		// 4- Finally let reth node build block at height 1, process and finalize it. This only
		// succeeds on Reth with the --engine.always-process-payload-attributes-on-canonical-head flag.
		prepareRequest := &types.PrepareProposalRequest{
			Height:          blkHeight,
			Time:            consensusTime,
			ProposerAddress: rethNodeAddress,
		}
		proposal, prepareErr := testClient.SimComet.Comet.PrepareProposal(testClient.CtxComet, prepareRequest)
		s.Require().NoError(prepareErr)
		s.Require().Len(proposal.Txs, 2)

		// Process the proposal via reth node. The proposal is expected
		// to pass and start building payload for height 2, optimistically.
		processRequest := &types.ProcessProposalRequest{
			Txs:                 proposal.Txs,
			Height:              blkHeight,
			ProposerAddress:     rethNodeAddress,
			Time:                consensusTime,
			NextProposerAddress: rethNodeAddress,
		}
		processResp, respErr := testClient.SimComet.Comet.ProcessProposal(testClient.CtxComet, processRequest)
		s.Require().NoError(respErr)
		s.Require().Equal(types.PROCESS_PROPOSAL_STATUS_ACCEPT.String(), processResp.Status.String())

		finalizeRequest := &types.FinalizeBlockRequest{
			Txs:             proposal.Txs,
			Height:          blkHeight,
			ProposerAddress: rethNodeAddress,
			Time:            consensusTime,
		}
		_, finalizeErr := testClient.SimComet.Comet.FinalizeBlock(testClient.CtxComet, finalizeRequest)
		s.Require().NoError(finalizeErr)
		_, commitErr := testClient.SimComet.Comet.Commit(testClient.CtxComet, &types.CommitRequest{})
		s.Require().NoError(commitErr)
	}
```

testing/simulated/payload_cache_test.go
```diff
@@ -0,0 +1,1066 @@
...
+		s.Require().NoError(processErr)
+		s.Require().Equal(types.PROCESS_PROPOSAL_STATUS_ACCEPT, processResp.Status)
+		s.Require().Contains(s.Reth.LogBuffer.String(), "Processing execution requests")
+	}
+}
+
+// Test a scenario where reth must rebuild a payload for a failed state transition.
+//
+// NOTE: this test requires reth with the --engine.always-process-payload-attributes-on-canonical-head flag.
+func (s *PayloadCacheSuite) TestReth_MustRebuildForFailedStateTransition_IsSuccessful() {
```
abi87, line 914: Note for self: this is the only case in which this test passes. Dropping the flag or swapping Reth for Geth causes a failure

beacon/blockchain/process_proposal.go
```diff
@@ -160,7 +174,7 @@ func (s *Service) ProcessProposal(
 		return nil, err
 	}
 
-	return valUpdates.CanonicalSort(), nil
+	return valUpdates.CanonicalSort(), err
```
fridrik01, line 177: `err` should be guaranteed to be `nil` per the check above

```suggestion
	return valUpdates.CanonicalSort(), nil
```

beacon/blockchain/process_proposal.go
```diff
@@ -354,8 +363,67 @@ func (s *Service) verifyStateRoot(
...
+	if processSlotErr != nil {
+		s.logger.Warn(
+			"Skipping payload eviction on rejected proposal",
+			"reason", "s.stateProcessor.ProcessSlots()",
+			"error", processSlotErr,
+		)
+		return
+	}
+	parentBlockRoot, blockRootFetchErr := preState.GetBlockRootAtIndex(
+		slot.Unwrap() % s.chainSpec.SlotsPerHistoricalRoot(),
```
fridrik01, line 417: Does this not need to be the parent slot? 
```suggestion
	parentBlockRoot, blockRootFetchErr := preState.GetBlockRootAtIndex(
		(slot.Unwrap()-1 ) % s.chainSpec.SlotsPerHistoricalRoot(),
```

beacon/blockchain/process_proposal.go
fridrik01: nvmd, missed the defer() above which can modify this err

beacon/blockchain/process_proposal.go
```diff
@@ -62,10 +62,21 @@ func (s *Service) ProcessProposal(
...
+		// Copy the state now before it is modified by the state transition.
+		s.storageBackend.StateFromContext(ctx).Copy(ctx),
+		&err, thisNodeAddress, req.GetProposerAddress(),
+	)
+
+	var (
+		signedBlk *ctypes.SignedBeaconBlock
+		sidecars  datypes.BlobSidecars
+	)
+	signedBlk, sidecars, err = s.ParseBeaconBlock(req)
```
abi87, line 76: note to self: we may evict the cached payload even if the error is at this level (see test `TestValidProposer_ProposesPostForkBlockIsNotFinalized_IsSuccessful`).
This does not seem correct, as it is very easy for a malicious proposer to enforce this error (just send garbage txs) and I don't see why the consequence should be to evict this node payload

PR #2951, opened 2025-10-13
fix(block-building): Cache verified payload

abi87:
Whenever building optimistically, and a block successifully verifies, cache the latest verified payload, to be reused in case there are network issues and verified block cannot be finalized.
Should node be requested to rebuild a sibling of the verified payload, it cannot access the EVM as once we mark a block as HEAD in FCU, we cannot go back and build an HEAD sibling, according to specs

### DIFF
--- a/payload/builder/payload.go
+++ b/payload/builder/payload.go
@@ -108,17 +108,8 @@ func (pb *PayloadBuilder) RequestPayloadSync(
 		return nil, ErrPayloadBuilderDisabled
 	}
 
-	pb.muEnv.Lock()
-	defer pb.muEnv.Unlock()
-	if pb.latestEnvelope != nil && r.Slot == pb.latestEnvelopeSlot {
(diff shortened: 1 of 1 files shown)


beacon/blockchain/payload.go
```diff
@@ -244,4 +245,11 @@ func (s *Service) handleOptimisticPayloadBuild(
 	}
 
 	s.metrics.markOptimisticPayloadBuildSuccess(buildData.Slot)
+	s.localBuilder.CacheLatestVerifiedPayload(
```
abi87: when we successfully verified the block and start building optimistically its child, we cache verified payload.
This in case this node is requested to build a verified block sybliing, which can't be done via this node EVM (see [Engine API specs](https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#specification-1))

PR #2952, opened 2025-10-13
[do not merge] payload cache sim tests

calbera:
Just the tests of #2949 without any changes to payload cache eviction rules.

`TestReth_RebuildPayload_IsSuccessful` and `TestGeth_RebuildPayload_IsSuccessful` (regardless of reth flag) fail because a failed proposal is still kept in payload cache. Upon resubmitting this proposal, it will fail the next round as well.

Also, `TestReth_MustRebuildPostForkPayload_IsSuccessful` fails (even with reth flag on) because a pre-fork proposal is still kept in payload cache. Subsequent rounds have a timestamp passed into a post-fork time. The subsequent rounds will fail if we resubmit the same proposal. I.e. the cache must be pruned and we must force rebuild a new payload.

Also, note that `TestReth_MustRebuildPreForkPayload_IsSuccessful` passes. However it fails if we call `GetAndEvict` upon `RequestPayloadSync` calls.

### DIFF
--- a/testing/simulated/malicious_proposer_test.go
+++ b/testing/simulated/malicious_proposer_test.go
@@ -157,37 +157,24 @@ func (s *SimulatedSuite) TestProcessProposal_InvalidTimestamps_Errors() {
 
 	// Initialize the chain state.
 	s.InitializeChain(s.T())
-
-	// Retrieve the BLS signer and proposer address.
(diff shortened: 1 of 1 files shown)


calbera:
Closing in favor of #2963

PR #2953, opened 2025-10-14
chore(payload): payloadID cache by slot only

abi87:


payload/cache/payload_id.go
```diff
@@ -47,7 +47,6 @@ type PayloadIDCache struct {
 // payloadID from the cache.
 type payloadIDCacheKey struct {
 	slot math.Slot
-	root common.Root
```
calbera, line 50: should we still be using this struct for key type? Might be simpler to just use math.Slot as the key type which ends up being self explanatory to the reader.

abi87:
@calbera  closing this down in the interest of reducing the outstanding PRs.
We can always reopen this or replay the clenaup later on

PR #2956, opened 2025-10-15
fix(block-building): Cache verified payload

abi87:
Whenever building optimistically, and a block successifully verifies, cache the latest verified payload, to be reused in case there are network issues and verified block cannot be finalized.
Should the node be requested to rebuild a sibling of the verified payload, it cannot access the EVM as once we mark a block as HEAD in FCU, we cannot go back and build an HEAD sibling, according to specs

Importantly I have added the `TestReth_MultiplePayloadRebuilds` that fails without the verified payload reuse and does pass with it.

### DIFF
--- a/beacon/blockchain/process_proposal.go
+++ b/beacon/blockchain/process_proposal.go
@@ -297,9 +297,9 @@ func (s *Service) VerifyIncomingBlock(
 	)
 
 	if shouldBuildNextPayload {
-		// state copy makes sure that preFetchBuildData does not affect state
-		copiedState := state.Copy(ctx)
-		nextBlockData, errFetch = s.preFetchBuildData(copiedState, blk.GetConsensusTime())
+		// makes sure that preFetchBuildData does not affect state
+		ephemeralState := state.Protect(ctx)
+		nextBlockData, errFetch = s.preFetchBuildData(ephemeralState, blk.GetConsensusTime())
 		if errFetch != nil {
 			// We don't return with err if pre-fetch fails. Instead we log the issue
 			// and still move to process the current block. Next block can always be
@@ -338,9 +338,9 @@ func (s *Service) VerifyIncomingBlock(
 	)
 
 	if shouldBuildNextPayload {
-		// state copy makes sure that preFetchBuildDataForSuccess does not affect state
-		copiedState := state.Copy(ctx)
-		nextBlockData, errFetch = s.preFetchBuildData(copiedState, blk.GetConsensusTime())
+		// makes sure that preFetchBuildDataForSuccess does not affect state
+		ephemeralState := state.Protect(ctx)
+		nextBlockData, errFetch = s.preFetchBuildData(ephemeralState, blk.GetConsensusTime())
 		if errFetch != nil {
 			// We don't mark the block as rejected if it is valid but pre-fetch fails.
 			// Instead we log the issue and move to process the current block.
--- a/node-api/backend/getters.go
+++ b/node-api/backend/getters.go
@@ -63,9 +63,9 @@ func (b *Backend) StateAndSlotFromHeight(height int64) (ReadOnlyBeaconState, mat
 		// Copy the state to ensure clients potential changes won't pollute the state
 		// Also we make sure to create the copy in a thread-safe way via the muCms mutex.
 		ms := b.cms.CacheMultiStore()
-		copyCtx := sdk.NewContext(ms, true, log.NewNopLogger())
-		copyGenesisState := b.genesisState.Copy(copyCtx)
-		return copyGenesisState, 0, nil
+		ctx := sdk.NewContext(ms, true, log.NewNopLogger())
+		ephemeralGenesisState := b.genesisState.Protect(ctx)
+		return ephemeralGenesisState, 0, nil
 	}
 
 	height = max(0, height) // CreateQueryContext uses 0 to pick latest height.
--- a/node-core/components/chain_service.go
+++ b/node-core/components/chain_service.go
@@ -24,13 +24,11 @@ import (
 	"cosmossdk.io/depinject"
 	"github.com/berachain/beacon-kit/beacon/blockchain"
 	"github.com/berachain/beacon-kit/chain"
-	"github.com/berachain/beacon-kit/config"
 	"github.com/berachain/beacon-kit/execution/deposit"
 	"github.com/berachain/beacon-kit/execution/engine"
 	"github.com/berachain/beacon-kit/log/phuslu"
 	"github.com/berachain/beacon-kit/node-core/components/metrics"
 	"github.com/berachain/beacon-kit/node-core/components/storage"
-	"github.com/berachain/beacon-kit/primitives/crypto"
 )
 
 // ChainServiceInput is the input for the chain service provider.
@@ -38,11 +36,9 @@ type ChainServiceInput struct {
 	depinject.In
 
 	ChainSpec             chain.Spec
-	Cfg                   *config.Config
 	ExecutionEngine       *engine.Engine
 	LocalBuilder          LocalBuilder
 	Logger                *phuslu.Logger
-	Signer                crypto.BLSSigner
 	StateProcessor        StateProcessor
 	StorageBackend        *storage.Backend
 	BlobProcessor         BlobProcessor
--- a/state-transition/core/state/statedb.go
+++ b/state-transition/core/state/statedb.go
@@ -57,8 +57,13 @@ func NewBeaconStateFromDB(
 	}
 }
 
-// Copy returns a copy of the beacon state.
-func (s *StateDB) Copy(ctx context.Context) *StateDB {
+// Protect returns an almost copy of stateDB. Specifically Protect guarantees that:
+// - No changes done on the returned state will affect the original state
+// - However, changes done on the original state will be carried over to the returned state.
+// The behaviour is probably best understood by considering the context hosts a stack of cache layers.
+// So write operations to the top cache won't be flushed to the lower layer but read operations will walk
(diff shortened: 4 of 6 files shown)


beacon/blockchain/process_proposal.go
```diff
@@ -160,6 +161,26 @@ func (s *Service) ProcessProposal(
...
+	for _, s := range sidecars {
+		blobBundle.Commitments = append(blobBundle.Commitments, s.GetKzgCommitment())
+		blobBundle.Proofs = append(blobBundle.Proofs, s.GetKzgProof())
+
+		blob := s.GetBlob()
+		blobBundle.Blobs = append(blobBundle.Blobs, &blob)
+	}
+	s.localBuilder.CacheLatestVerifiedPayload(
+		blk.Slot,
+		ctypes.NewExecutionPayloadEnvelope[*engineprimitives.BlobsBundleV1](
```
abi87, line 191: TODO: missing execution Requests here

payload/builder/payload.go
```diff
@@ -152,21 +152,39 @@ func (pb *PayloadBuilder) RetrievePayload(
...
-	// Attempt to see if we previously fired off a payload built for
-	// this particular slot and parent block root.
+	// 1- Node could be asked to build a block at slot H even if it already verified
+	// a block at that slot H. This happens if the block passes verification
+	// but is not finalized (e.g. due to network issue). In such a case, the
+	// EVM may not be able to serve a new payload (e.g. if it has received a
+	// FCU call with HEAD == H+1). To avoiding a failure in building the block
+	// we reuse a validated payload if it's available
+	pb.muEnv.Lock()
+	defer pb.muEnv.Unlock()
```
calbera: should we just move this Unlock to right after, instead of being locked for the whole function

payload/builder/payload.go
```diff
@@ -152,21 +152,39 @@ func (pb *PayloadBuilder) RetrievePayload(
 		return nil, ErrPayloadBuilderDisabled
 	}
 
-	// Attempt to see if we previously fired off a payload built for
-	// this particular slot and parent block root.
+	// 1- Node could be asked to build a block at slot H even if it already verified
```
calbera: do we want to switch order? i.e. 1 - first check if we have a payload. 2 - check if we have a valid cached one. 3 - (outside of this function) request to build a new one jit

payload/builder/payload.go
abi87: Yes, thanks. A larger scope used to be needed, but it's not anymore. Introduced a little getter to keep the scope minimal. IDK if a more direct solution is possible since `latestEnvelope` is an interface and I don't have a safe way to copy/clone it before releasing the lock

payload/builder/payload.go
abi87: What you suggest would be ideal. I am only worried about cases where:

- we do have a payloadID cached for the right slot
- but the payload got stale in the EVM so `getPayload` returns -38001

We can still do the lookup for verified payload in this case, but I feel  the code flow becomes complex.
So I would rather work in steps here. Make sure that this all code address the major issue (node not been able to build a block if it has start building optimistically for N+1 height and it is then requested to build for height N).
We can then add a PR on top to prefer local payloads to remote ones, once we are sure the code is correct.
@calbera thoughts?

beacon/blockchain/process_proposal.go
```diff
@@ -160,6 +161,40 @@ func (s *Service) ProcessProposal(
...
+	case errReqs == nil:
+		executionRequests, err = ctypes.GetExecutionRequestsList(reqs)
+		if err != nil {
+			return nil, fmt.Errorf("failed retrieving encoded execution requests from payload: %w", err)
+		}
+	case errors.Is(errReqs, ctypes.ErrFieldNotSupportedOnFork):
+		// nothing to do, executionRequests is nil
+	default:
+		return nil, fmt.Errorf("failed getting execution requests from payload: %w", errReqs)
+	}
```
abi87, line 188: @calbera @rezbera this is probably the most annoying part of the PR. To reuse the payload without changing much code, I need to go back to an `executionPayloadEnvelope` which contains the raw data needed to build the block/blob. I think this is correct but would appreciate you double checking

beacon/blockchain/finalize_block.go
```diff
@@ -150,6 +150,10 @@ func (s *Service) PostFinalizeBlockOps(ctx sdk.Context, blk *ctypes.BeaconBlock)
 		return fmt.Errorf("sendPostBlockFCU failed: %w", err)
 	}
 
+	// reset latest verified payload in block builder to signal
+	// that no payload is available to reuse for blk.Slot
+	s.localBuilder.CacheLatestVerifiedPayload(blk.Slot, nil)
```
abi87, line 155: This is not strictly necessary, since we do track the slot in block builder, so there is no risk of reusing an already finalized payload (CometBFT won't request a node to rebuild a finalized block).
Still I like the idea of having a clean lifetime for the cached payload, and this clearly signal the one case where the cached payload should be cleared out.

payload/builder/payload.go
```diff
@@ -152,21 +152,39 @@ func (pb *PayloadBuilder) RetrievePayload(
...
+	// a block at that slot H. This happens if the block passes verification
+	// but is not finalized (e.g. due to network issue). In such a case, the
+	// EVM may not be able to serve a new payload (e.g. if it has received a
+	// FCU call with HEAD == H+1). To avoiding a failure in building the block
+	// we reuse a validated payload if it's available
+	pb.muEnv.Lock()
+	if pb.latestEnvelope != nil && slot == pb.latestEnvelopeSlot {
+		return pb.latestEnvelope, nil
```
abi87: note: we don't clear the cached payload when we use it.
While CometBFT does guarantee to call prepareProposal only one per height (it must cache the block built and re-issue it if the validator becomes proposer multiple time), block building may still fail or timeout in BeaconKit, before the block gets into CometBFT scope. In such a case the second time, node won't be able to repropose latestEnvelope a second time.
Note also that among different calls for the same slot, `latestEnvelope` may change as we reset it every time a new payload gets successfully verified.

The `latestEnvelope` will be finally cached once the block is finalized (in which case CometBFT should guarantee not to call PrepareProposal for the finalized height, but still nice to have clean lifetimes for objects)

payload/builder/payload.go
abi87: Added a TODO in any case

consensus-types/types/payload_env.go
```diff
@@ -62,6 +62,20 @@ func NewEmptyExecutionPayloadEnvelope[
...
+	BlobsBundleT engineprimitives.BlobsBundle,
+](
+	payload *ExecutionPayload,
+	blobs BlobsBundleT,
+	executionRequests []EncodedExecutionRequest,
+) BuiltExecutionPayloadEnv {
+	return &executionPayloadEnvelope[BlobsBundleT]{
+		ExecutionPayload:  payload,
+		BlobsBundle:       blobs,
+		ExecutionRequests: executionRequests,
```
abi87, line 75: `executionPayloadEnvelope` has other attributes that do not seem used in the payload, so not passing them in this constructor

testing/simulated/optimistic_test.go
```diff
@@ -205,3 +206,151 @@ func (s *PectraForkSuite) TestGeth_RebuildPayload_IsSuccessful() {
...
+	s.Geth.InitializeChain(s.T()) // helper node
+	testEL := s.Reth
+	helpBuilder := s.Geth
+
+	// Retrieve the BLS signer and proposer address.
+	blsSigner := simulated.GetBlsSigner(testEL.HomeDir)
+	pubkey, err := blsSigner.GetPubKey()
+	s.Require().NoError(err)
+
+	const blkHeight = int64(1)
```
abi87, line 222: const to make sure we don't change the height and we keep on verifying and rebuilding blocks at this height

beacon/blockchain/process_proposal.go
```diff
@@ -160,6 +161,40 @@ func (s *Service) ProcessProposal(
...
+		blob := s.GetBlob()
+		blobBundle.Blobs = append(blobBundle.Blobs, &blob)
+	}
+
+	var executionRequests []ctypes.EncodedExecutionRequest
+	switch reqs, errReqs := blk.Body.GetExecutionRequests(); {
+	case errReqs == nil:
+		executionRequests, err = ctypes.GetExecutionRequestsList(reqs)
+		if err != nil {
+			return nil, fmt.Errorf("failed retrieving encoded execution requests from payload: %w", err)
```
abi87, line 182: so technically this error and the one below should never happens since the payload has been successfully verified.
Should these errors ever happens, it's a bug and the block verification will be marked as failed, causing liveness issue.

payload/builder/payload.go
calbera: I think if your main worry is getPayload in EVM returns -38001 (which both geth and reth shouldn't return now since they cache more payloads), then we can explicitly handle that case by deleting the payloadID from cache right then and then continuing down the control flow. 

i.e. 
1. Try payload ID from cache and getPayload
   a. if -38001 delete from cache
2. Try verified and cached "emergency" block 
3. build just in time

My main insistence on this is just to protect the "rights" of a validator. i.e. honor their earned right to propose based on their stake. Putting in pre-filled proposals just feels like it defeats some part of permissionlessness, even if it improves liveness

abi87: Fixed. I was annoyed by the need to call `getLatestVerifiedPayload` twice, but you're right it's not too bad and probably more correct.

abi87:
Replaced by https://github.com/berachain/beacon-kit/pull/2963

PR #2958, opened 2025-10-19
chore(stateDB): renamed `statedb.Copy`

abi87:
`statedb.Copy` is a very bad name for the function as it does not really make a copy, not even a shallow one.
What this function does is generate and object whose content at creation matches the original state, but what happens after creation is not what one would expect from a copy.
Specifically, any change done over the state return by the function won't affect the original state. However any change change done over the original state *is* carried over the return state.
Note that this is different from a shallow copy, where changes are bidirectional (modifying the copy would modify the original state, which is not the case here)

The real purpose of this function is to be able to read and write the returned state without affecting the original state. So I propose the name `Protect` as a better alternative to `Copy`.
Open to suggestion here. I think any name on earth, in any language, is better and would cause less confusion than `Copy`.

### DIFF
--- a/state-transition/core/state/statedb.go
+++ b/state-transition/core/state/statedb.go
@@ -57,9 +57,9 @@ func NewBeaconStateFromDB(
 	}
 }
 
-// Protect returns an almost copy of stateDB. Specifically Protects guaranteed that:
+// Protect returns an almost copy of stateDB. Specifically Protect guarantees that:
 // - No changes done on the returned state will affect the original state
-// - However, changes carried out on the original state will be carried out to the returned state.
+// - However, changes done on the original state will be carried over to the returned state.
 // The behaviour is probably best understood by considering the context hosts a stack of cache layers.
 // So write operations to the top cache won't be flushed to the lower layer but read operations will walk
 // through the cache stack, so bubbling up changes from the lower layers to the top ones.


state-transition/core/state/statedb_test.go
```diff
@@ -0,0 +1,108 @@
...
+	require.ErrorIs(t, err, collections.ErrNotFound)
+
+	// 3- Show that changes made to original state after protection
+	// are carried over the protecting state
+	wantEthIdx := uint64(1987)
+	require.NoError(t, originalState.SetEth1DepositIndex(wantEthIdx))
+
+	gotEthIdx, err := protectingState.GetEth1DepositIndex()
+	require.NoError(t, err)
+	require.Equal(t, wantEthIdx, gotEthIdx)
```
abi87, line 105: This is what makes `statedb.Copy` not a copy

state-transition/core/state/statedb.go
```diff
@@ -57,8 +57,13 @@ func NewBeaconStateFromDB(
...
 
-// Copy returns a copy of the beacon state.
-func (s *StateDB) Copy(ctx context.Context) *StateDB {
+// Protect returns an almost copy of stateDB. Specifically Protect guarantees that:
+// - No changes done on the returned state will affect the original state
+// - However, changes done on the original state will be carried over to the returned state.
+// The behaviour is probably best understood by considering the context hosts a stack of cache layers.
+// So write operations to the top cache won't be flushed to the lower layer but read operations will walk
+// through the cache stack, so bubbling up changes from the lower layers to the top ones.
+func (s *StateDB) Protect(ctx context.Context) *StateDB {
```
fridrik01, line 66: what about  calling it just`Fork`?  It indicates a one-way relationship where its common to maintain a connection to the parent (via pull from parent)

abi87: `Fork` is a loaded word for us. We have hard forks, CL and EL forks, slot and timestamp based forks.
Doesn't `Fork` add to the overload?

fridrik01: Just throwing ideas, naming is hard... Fair point about terminology confusion here. However, git uses `fork` for repositories and Linux for processes, so the usage `stateDB.Fork()` should make it clear we're forking state, not protocol-level code.  Another idea would be `branch` but other than that I don't have a better suggestion than `protect` atm.

calbera: Protect feels because it isnt immediately obvious that it even does a copy. But i understand the motivation for naming like this

testing/simulated/utils.go
```diff
@@ -326,8 +326,9 @@ func ComputeAndSetStateRoot(
 	block *ctypes.BeaconBlock,
 ) (*ctypes.BeaconBlock, error) {
 
-	// Copy the current state from the storage backend.
-	stateDBCopy := storageBackend.StateFromContext(queryCtx).Copy(queryCtx)
+	// Create an ephemeral state out of storage backend to avoid polluting
+	// the original state
+	ephemeralState := storageBackend.StateFromContext(queryCtx).Protect(queryCtx)
```
calbera, line 331: can we take this opportunity to clean this up a bit? why do we pass in this ctx `queryCtx` in this case twice? 

Maybe 2 options:
- `storageBackend.NewState().Protect(queryCtx)`
- `storageBackend.StateFromContext(queryCtx).Protect()`

I kind of prefer the latter since "Protected" copy shouldnt require another ctx, logically

calbera (commented):
Overall agree with this change

testing/simulated/utils.go
abi87: I agree we could use a cleanup here, but I am not sure this is easily done. I would defer it to a future PR.

In a nutshell there are two levels here:

- the `KVStore` maps may be copied, possibly to avoid polluting the maps across usages, or to allow safe concurrent use
- the state hosted in the context may be directly used or a cache may be built on top of it.

From what I have seen the first operations seems useless but I don't want to venture in cleaning it right now. We can definitely come back to this one we close the optimistic block building chapter

calbera (approved):

PR #2959, opened 2025-10-19
more precise payload eviction for optimistic - fix state handling

abi87:
Following https://github.com/berachain/beacon-kit/pull/2958, a suggestion on how to fix the state handling

### DIFF
--- a/beacon/blockchain/process_proposal.go
+++ b/beacon/blockchain/process_proposal.go
@@ -422,9 +422,8 @@ func (s *Service) evictLocalPayloadIfNecessary(
 		return
 	}
 	// Check conditions for calling this function.
-	// - Function is called with an initialized error to hold process proposal error.
 	// - Our node has the local payload builder enabled.
-	if processProposalErr == nil || !s.localBuilder.Enabled() {
+	if !s.localBuilder.Enabled() {
 		return
 	}
 


beacon/blockchain/process_proposal.go
```diff
@@ -372,15 +409,21 @@ func (s *Service) shouldBuildNextPayload(isNextBlockProposer bool) bool {
 // We deem it necessary only when we are rejecting the incoming block proposal and it has
 // been built by us.
 func (s *Service) evictLocalPayloadIfNecessary(
-	preState *statedb.StateDB,
```
abi87, line 375: As showed in https://github.com/berachain/beacon-kit/pull/2958, preState is affected by the changes carried out by the block, even if copied before the state is modified.
Also we don't know at this point whether the block is valid or not, and in the latter case we cannot trust any data in the state.
I believe the only safe way to do this (similarly to `s.handleOptimisticPayloadBuild`) is to load all relevant at the beginning (making sure to make a copy  to avoid polluting the block state), and avoid touching the state right after

calbera (commented):
My only issue is we are precomputing a lot of data that in most cases will not even be used. Also with the defer approach, `preState` is actually computed immediately, as arguments are evaluated when the defer statement is actually run. Later on when the deferred function executes the values of function args are already set will not have any effects.

beacon/blockchain/process_proposal.go
```diff
@@ -372,15 +409,21 @@ func (s *Service) shouldBuildNextPayload(isNextBlockProposer bool) bool {
...
+	blkSlot math.Slot,
+	parentBlockRoot common.Root,
+	prepareErr *error,
 	processProposalErr *error,
 	thisNodeAddress []byte,
 	proposerAddress []byte,
 ) {
+	if *prepareErr != nil {
```
calbera, line 419: would want to keep the prepareErr == nil check as to avoid nil ptr panics in critical path code, although its technically guaranteed as long as we always call it this way

abi87:
> > My only issue is we are precomputing a lot of data that in most cases will not even be used. Also with the defer approach, `preState` is actually computed immediately, as arguments are evaluated when the defer statement is actually run. Later on when the deferred function executes the values of function args are already set will not have any effects.
> 
> @calbera a few points:
> 
> * I totally agree about the amount of precompute done here. I technically disagree about the rest.
>   
>   * `s.storageBackend.StateFromContext(ctx).Copy(ctx)` is enough to guarantee correctness if the block is successfully validated. In this case the state would correct and `Copy` would ensure that the computation does in `defer` does not modify it. All good.
>   * however that is **not** enough in case the block fails. In that case we cannot assume anything is correct in the state and we must not use it to do any computation.
>   * Now it's true that we read from state only is `!errors.Is(*processProposalErr, ErrRebuildForFailedStateTransition)`, so this is technically enough that we don't read from an invalid state.
>   * Still I feel this is brittle and i would prefer, if we must keep `evictLocalPayloadIfNecessary` to precompute
> * More generally, I am doubtful about the overall approach:
>   
>   * I am very much in favour of splitting `GetAndEvict` in `Get` and `Delete` and of better use of `Delete`
>   * I am not sure we need `evictLocalPayloadIfNecessary` at all: I don't know of any unit test or scenario that passes if and only if `evictLocalPayloadIfNecessary` is kept, so it's not clear to me yet what we stand to gain by merging this in.
> 
> Definitely happy to chat about this whole

I should note that all the preprocessing goes away if we cache payloadIDs by slot only and not parent root

PR #2960, opened 2025-10-20
fix(block-building): `ProcessFork` over optimistic block building

abi87:
`ProcessFork` should be guaranteed to be called in every `PrepareProposal` code flow; specifically whether block is being built optimistically or not/

### DIFF
--- a/beacon/validator/block_builder.go
+++ b/beacon/validator/block_builder.go
@@ -28,6 +28,7 @@ import (
 	payloadtime "github.com/berachain/beacon-kit/beacon/payload-time"
 	ctypes "github.com/berachain/beacon-kit/consensus-types/types"
 	"github.com/berachain/beacon-kit/consensus/types"
+	engineprimitives "github.com/berachain/beacon-kit/engine-primitives/engine-primitives"
 	"github.com/berachain/beacon-kit/errors"
 	"github.com/berachain/beacon-kit/payload/builder"
 	"github.com/berachain/beacon-kit/primitives/bytes"
@@ -279,13 +280,16 @@ func (s *Service) retrieveExecutionPayload(
 	}
 
 	r := &builder.RequestPayloadData{
-		Slot:                 slot,
-		Timestamp:            nextPayloadTimestamp,
-		PayloadWithdrawals:   payloadWithdrawals,
-		PrevRandao:           prevRandao,
-		ParentBlockRoot:      parentBlockRoot,
-		HeadEth1BlockHash:    lph.GetBlockHash(),
-		FinalEth1BlockHash:   lph.GetParentHash(),
+		Slot:               slot,
+		Timestamp:          nextPayloadTimestamp,
+		PayloadWithdrawals: payloadWithdrawals,
+		PrevRandao:         prevRandao,
+		ParentBlockRoot:    parentBlockRoot,
+		FCState: engineprimitives.ForkchoiceStateV1{
+			HeadBlockHash:      lph.GetBlockHash(),
+			SafeBlockHash:      lph.GetParentHash(),
+			FinalizedBlockHash: lph.GetParentHash(),
+		},
 		ParentProposerPubkey: parentProposerPubkey,
 	}
 	return s.localPayloadBuilder.RequestPayloadSync(ctx, r)
--- a/beacon/blockchain/execution_engine.go
+++ b/beacon/blockchain/execution_engine.go
@@ -41,12 +41,26 @@ func (s *Service) sendPostBlockFCU(
 	}
 
 	// Send a forkchoice update without payload attributes to notify EL of the new head.
+	// Note that we are being conservative here as we don't mark the block we just finalized
+	// (which is irreversible due to CometBFT SSF) as final. If we keep doing this, we can
+	// spare the FCU update in case we have optimistic block building on, as we may have
+	// already sent the very same FCU request after we verified the block.
+	fcuData := &engineprimitives.ForkchoiceStateV1{
+		HeadBlockHash:      lph.GetBlockHash(),
+		SafeBlockHash:      lph.GetParentHash(),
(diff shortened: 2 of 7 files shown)


beacon/validator/block_builder.go
```diff
@@ -220,17 +220,11 @@ func (s *Service) retrieveExecutionPayload(
...
 	}
 
 	// If we failed to retrieve the payload, request a synchronous payload.
-	//
-	// NOTE: The state here is properly configured by the
-	// prepareStateForBuilding
-	//
-	// call that needs to be called before requesting the Payload.
-	// TODO: We should decouple the PayloadBuilder from BeaconState to make
-	// this less confusing.
```
abi87, line 240: outdated, meaningless comment

beacon/validator/block_builder.go
```diff
@@ -220,17 +220,11 @@ func (s *Service) retrieveExecutionPayload(
 	slot := slotData.GetSlot()
 	envelope, err := s.localPayloadBuilder.RetrievePayload(ctx, slot, parentBlockRoot)
 	if err == nil {
-		return envelope, nil
+		nextPayloadTimestamp := envelope.GetExecutionPayload().GetTimestamp()
+		return envelope, s.stateProcessor.ProcessFork(st, nextPayloadTimestamp, false)
```
abi87, line 231: this makes sure `ProcessFork` is duly called on both code flows (optimistic and just in time)

state-transition/core/state_processor_forks.go
```diff
@@ -48,6 +48,7 @@ import (
 func (sp *StateProcessor) ProcessFork(
 	st *statedb.StateDB, timestamp math.U64, logUpgrade bool,
 ) error {
+	sp.logger.Debug("invoked ProcessFork") // a log needed for testing purposes
```
abi87, line 51: current forks do not perform changes that are easily assertable from `simulated` tests. This little log simplifies the tests

beacon/validator/block_builder.go
abi87, line 253: `ProcessFork` is correctly called here, so it should be called even if the block is being built optimistically.

testing/simulated/optimistic_test.go
```diff
@@ -0,0 +1,131 @@
...
+		// 4- Build post-fork block. Make sure that the fork transition happens
+		prepareRequest := &types.PrepareProposalRequest{
+			Height:          nextBlockHeight,
+			Time:            time.Unix(int64(secondBlkConsensusTime), 0),
+			ProposerAddress: pubkey.Address(),
+		}
+		proposal, prepareErr := client.SimComet.Comet.PrepareProposal(client.CtxComet, prepareRequest)
+		s.Require().NoError(prepareErr)
+		s.Require().Len(proposal.Txs, 2)
+		s.Require().Contains(client.LogBuffer.String(), "invoked ProcessFork")
```
abi87, line 106: This is the simplest way I have found to add a test for this fix.
Almost moot honestly, as if there is not a a visible behaviour that is broken, do we really have a bug?
My reasoning is that in some future forks, we may have some behaviour that is visible and I want this fixed before we get there.

state-transition/core/state_processor_forks.go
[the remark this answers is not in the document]
abi87: You're right and I expect similar observations by other reviewers. See https://github.com/berachain/beacon-kit/pull/2960#discussion_r2444177502

calbera: a little annoying but im fine with leaving it in

testing/simulated/optimistic_test.go
```diff
@@ -0,0 +1,131 @@
...
+	payloadtime "github.com/berachain/beacon-kit/beacon/payload-time"
+	"github.com/berachain/beacon-kit/primitives/math"
+	"github.com/berachain/beacon-kit/testing/simulated"
+	"github.com/cometbft/cometbft/abci/types"
+)
+
+func (s *PectraForkSuite) Test_BuildOptimisticallyAtFork() {
+	// Initialize the chain state.
+	client := s.Geth
+	client.InitializeChain(s.T()) // 1 reth validator
```
calbera:
```suggestion
	client.InitializeChain(s.T()) // 1 geth validator
```

beacon/validator/block_builder.go
calbera: I'm not quite sure we need this? Not understanding adding this here because in the calling function `BuildBlockAndSidecars`, after calling `retrieveExecutionPayload` we call `computeStateRoot` which eventually calls `stateProcessor.Transition()` which does call `ProcessFork`. Is there a need to call ProcessFork now / earlier than when it gets called?

calbera: I see the new sim test proves that ProcessFork is called by the log assertion. But I'm confused why this isn't already passing since downstream ProcessFork should be called regardless?

abi87: @calbera  `ProcessFork` is called below if block is not built optimistically (i.e. if `getPayload` with cached payloadID fails). However it is not called when block is built optimistically.
Now the state a block sees should be the same no matter whether the block it is built optimistically or not right? So by symmetry we should either call `ProcessFork` in both cases or never. I lean the former

testing/simulated/optimistic_test.go
abi87: Replaced by the more generic `client.InitializeChain(s.T()) // init the validator`.

beacon/validator/block_builder.go
abi87: Also @calbera, you're right, thanks! The test indeed passes even if I comment out the change I propose, so the UT is really not proving anything and should be removed/fixed (not sure how).
Stil I maintain that we should call `ProcessFork` in the same order whether we build the block optimistically or not. Now IDK what to do with the UT though!

PR #2963, opened 2025-10-21
[2/n] test(payload-building): Simulated tests to uncover optimistic building edge cases

calbera:
It can be seen here which tests are broken. Subsequent PRs in this chain (3/n, 4/n, ...) should fix tests incrementally

Failing tests list:

- [ ] TestValidProposer_ProposesPostForkBlockIsNotFinalized_IsSuccessful (passing in [1/n PR](https://github.com/berachain/beacon-kit/pull/2944), now passing in 3/n)
- [ ] TestGeth_RebuildPayload_IsSuccessful (passing in 3/n)
- [ ] TestReth_RebuildPayload_IsSuccessful (passing in 3/n)
- [ ] TestReth_MustRebuildPreForkPayload_IsSuccessful (passing in 4/n)
- [ ] TestReth_MustRebuildPostForkPayload_IsSuccessful (passing in 4/n)
- [ ] TestReth_MustRebuildForFailedStateTransition_IsSuccessful (passing in 4/n)


testing/simulated/malicious_proposer_test.go
```diff
@@ -157,26 +157,27 @@ func (s *SimulatedSuite) TestProcessProposal_InvalidTimestamps_Errors() {
...
-
-	// Retrieve the BLS signer and proposer address.
-	blsSigner := simulated.GetBlsSigner(s.HomeDir)
-	pubkey, err := blsSigner.GetPubKey()
+	nodeAddress, err := s.SimComet.GetNodeAddress()
 	s.Require().NoError(err)
+	// This effectively enforces that optimistic builds do NOT happen in this test (since
+	// MoveChainToHeight does not set the NextProposerAddress field in ProcessProposalRequests).
+	// This allows us to skip the extra PrepareProposal call on deleted line 176.
+	s.SimComet.Comet.SetNodeAddress(nodeAddress)
```
calbera, line 163: This effectively enforces that optimistic builds do NOT happen in this test (since `MoveChainToHeight` does not set the `NextProposerAddress` field in ProcessProposalRequests). This allows us to skip the extra PrepareProposal call on deleted line 176.

testing/simulated/pectra_fork_test.go
calbera, line 1: changes in this file exist to simplify the `TestValidProposer_ProposesPostForkBlockIsNotFinalized_IsSuccessful` and still preserve the crux of the test. 1 geth node is sufficient (as is used in the rest of `PectraForkSuite`)

testing/simulated/payload_cache_test.go
```diff
@@ -0,0 +1,1066 @@
...
+		s.Require().NoError(processErr)
+		s.Require().Equal(types.PROCESS_PROPOSAL_STATUS_ACCEPT, processResp.Status)
+		s.Require().Contains(s.Reth.LogBuffer.String(), "Processing execution requests")
+	}
+}
+
+// Test a scenario where reth must rebuild a payload for a failed state transition.
+//
+// NOTE: this test requires reth with the --engine.always-process-payload-attributes-on-canonical-head flag.
+func (s *PayloadCacheSuite) TestReth_MustRebuildForFailedStateTransition_IsSuccessful() {
```
calbera: Note that this test is the same as `TestReth_MultiplePayloadRebuilds` from [here](https://github.com/berachain/beacon-kit/pull/2956/files#diff-c9033c103fe3561778a91428bd7d498f3cbd9a19a3bb88caf339feb13a45d011R210), just renamed and refactored for a 2-node chain

testing/simulated/pectra_fork_test.go
```diff
@@ -579,6 +548,107 @@ func (s *PectraForkSuite) TestValidProposer_ProposesPreForkBlockWithPostForkCons
 	}
 }
 
+// This test will show that an optimistically building a payload across the fork boundary
+// correctly invokes `ProcessFork` on the state processor.
+func (s *PectraForkSuite) Test_OptimisticBuildAtFork_IsSuccessful() {
```
calbera, line 590: renamed `Test_BuildOptimisticallyAtFork` from [here](https://github.com/berachain/beacon-kit/pull/2960/files#diff-c9033c103fe3561778a91428bd7d498f3cbd9a19a3bb88caf339feb13a45d011R34)

testing/simulated/malicious_proposer_test.go
abi87: I understand this but don't like.
It is very implicit and easy to forget. Like in 6 months I would have forget this and I would go nuts if I have to debug/maintain tests.
Is there any way where we properly set the nodeAddress in the nodes when we setup the suite and revisit the tests to that we make each call have the relevant attributes set?

LIke it probably was not so important before optimistic block building, maybe not it is and we should fix it properly?

testing/simulated/pectra_fork_test.go
calbera, line 470: Not sure about why we changed this test in [this commit](https://github.com/berachain/beacon-kit/pull/2963/commits/36c5b4e729353101f9c74bc58c5a835d5065dbd2) and now its failing? Since it was passing in the [1/n PR](https://github.com/berachain/beacon-kit/pull/2944)

testing/simulated/execution/reth.go
```diff
@@ -41,7 +41,9 @@ func ValidRethImage() docker.PullImageOptions {
 	}
 }
 
-// defaultRethCmdStrBuilder returns a command string tailored for running a Geth node.
+// defaultRethCmdStrBuilder returns a command string tailored for running a Reth node.
+//
+// TODO: consider adding flag --engine.always-process-payload-attributes-on-canonical-head.
```
abi87: I am against introducing this flag:

- a BeaconKit fix down the line (https://github.com/berachain/beacon-kit/pull/2965) makes it unncessary
- It's a deviation from specs which I'd like to avoid

calbera: will remove this comment from this PR

testing/simulated/malicious_proposer_test.go
calbera: fixed now, since `MoveChainToHeight` (and all other tests) set the node address and use it as `NextProposerAddress` to trigger the optimistic builds

testing/simulated/pectra_fork_test.go
abi87: I think this is fine. The reason it failed is that the second block (with pre-fork consensus time and post-fork payload time) used to be generated relaying on cached payload, while in the test as I modified it it is built without resorting to any node (just modifying the block bytes themselves).
I think this is fine. What is not fine is this test keeps on failing on 3/n. Looking at it now

PR #2964, opened 2025-10-23
[3/n] fix payloadID cache clearing

abi87:
- Split `GetAndEvict` into `Get` and `Delete`. `Get` won't consume payloadID anymore, so the same payload can be retrieved multiple times (given it's cached by the EL)
- `Delete` payloadIDs only if they are proved to be stale, otherwise keep them as they are
- Fix payload retrieval to check if payload built optimistically has the right fork. If not, drop it and and rebuild a payload with the right fork

Failing tests list, marked as done if they pass in this PR:

- [X] TestValidProposer_ProposesPostForkBlockIsNotFinalized_IsSuccessful
- [X] TestGeth_RebuildPayload_IsSuccessful
- [X] TestReth_RebuildPayload_IsSuccessful
- [ ] TestReth_MustRebuildPreForkPayload_IsSuccessful
- [ ] TestReth_MustRebuildPostForkPayload_IsSuccessful
- [ ] TestReth_MustRebuildForFailedStateTransition_IsSuccessful

No previously passing tests are broken by this PR

### DIFF
--- a/payload/builder/errors.go
+++ b/payload/builder/errors.go
@@ -34,10 +34,6 @@ var (
 	// cache.
 	ErrPayloadIDNotFound = errors.New("unable to find payload ID in cache")
 
-	// ErrNilPayloadEnvelope is returned when a nil payload envelope is
-	// received.
-	ErrNilPayloadEnvelope = errors.New("received nil payload envelope")
-
 	// ErrNilWithdrawals is returned when nil withdrawals list is received.
 	ErrNilWithdrawals = errors.New("nil withdrawals received from execution client")
 )


payload/builder/payload.go
```diff
@@ -53,7 +56,7 @@ func (pb *PayloadBuilder) RequestPayloadAsync(
 		return nil, common.Version{}, ErrPayloadBuilderDisabled
 	}
 
-	if payloadID, found := pb.pc.GetAndEvict(r.Slot, r.ParentBlockRoot); found {
+	if payloadID, found := pb.pc.Get(r.Slot, r.ParentBlockRoot); found {
```
fridrik01, line 59: should we have a metric counter like `payload_cache_hit`/`payload_cache_miss`  for this?

beacon/validator/interfaces.go
```diff
@@ -57,6 +57,7 @@ type PayloadBuilder interface {
 		ctx context.Context,
 		slot math.Slot,
 		parentBlockRoot common.Root,
+		expectedForkVersion common.Version,
```
fridrik01, line 61: update interface comment for the new param

node-core/components/interfaces.go
```diff
@@ -86,6 +86,7 @@ type (
 			ctx context.Context,
 			slot math.Slot,
 			parentBlockRoot common.Root,
+			expectedForkVersion common.Version,
```
fridrik01, line 90: update interface comment for the new param

payload/cache/payload_id.go
fridrik01: Can't we do here: 
```suggestion
p.mu.RLock()
defer p.mu.RUnlock()
```

payload/builder/payload.go
```diff
@@ -144,21 +157,33 @@ func (pb *PayloadBuilder) RetrievePayload(
...
-	envelope, err := pb.getPayload(ctx, payloadID.PayloadID, payloadID.ForkVersion)
+	envelope, err := pb.getPayload(ctx, payloadRes.PayloadID, payloadRes.ForkVersion)
 	if err != nil {
+		if errors.Is(err, engineerrors.ErrUnknownPayload) ||
+			errors.Is(err, engineerrors.ErrNilExecutionPayloadEnvelope) {
+			// We may have cached the payloadID, but the payload have become stale
+			// in the EVM, or there could have been other issues. Block builder will
+			// try and build again the payload just in time
+			pb.pc.Delete(slot, parentBlockRoot)
+		}
```
fridrik01: minor nit: consider DRYing this up

payload/cache/payload_id.go
abi87: Yes! Should not matter much, but Get used to be a GetAndEvict, which modified the map. Now the Get is purely a get, hence we can use RLock

payload/builder/payload.go
abi87: We could but would do in a different PR

beacon/validator/interfaces.go
abi87: Done, please double check @fridrik01

node-core/components/interfaces.go
abi87: Done, please double check @fridrik01

payload/builder/payload.go
abi87: Improved comment, hopefully. Please double check @fridrik01

beacon/validator/block_builder.go
```diff
@@ -218,7 +218,8 @@ func (s *Service) retrieveExecutionPayload(
 	//
 	// Get the payload for the block.
 	slot := slotData.GetSlot()
-	envelope, err := s.localPayloadBuilder.RetrievePayload(ctx, slot, parentBlockRoot)
+	expectedPayloadFork := s.chainSpec.ActiveForkVersionForTimestamp(slotData.GetConsensusTime())
```
calbera, line 226: can we add a comment similar to [this](https://github.com/berachain/beacon-kit/blob/18c1baf55919e939f388ae5381df0bd14262e70f/beacon/blockchain/process_proposal.go#L72-L84) (although much shorter)? It's a subtlety as to why we're using consensusTime here and not payloadTime

payload/builder/payload.go
```diff
@@ -132,8 +135,19 @@ func (pb *PayloadBuilder) RequestPayloadSync(
 		return nil, ctx.Err()
 	}
 
-	// Get the payload from the execution client.
-	return pb.getPayload(ctx, *payloadID, forkVersion)
+	payload, err := pb.getPayload(ctx, *payloadID, forkVersion)
+	if err != nil {
+		if errors.Is(err, engineerrors.ErrUnknownPayload) ||
+			errors.Is(err, engineerrors.ErrNilExecutionPayloadEnvelope) {
```
calbera: would strongly recommend cleaning these errors before actually "using" them for logical changes. See [here](https://github.com/berachain/beacon-kit/pull/2949/files#r2433189050) for details

payload/cache/payload_id.go
```diff
@@ -78,22 +78,20 @@ func (p *PayloadIDCache) Has(
...
+// done via Delete and via pruning in Set.
+func (p *PayloadIDCache) Get(
 	slot math.Slot,
 	blockRoot common.Root,
 ) (PayloadIDCacheResult, bool) {
-	p.mu.Lock()
-	defer p.mu.Unlock()
+	p.mu.RLock()
+	defer p.mu.RUnlock()
 	key := payloadIDCacheKey{slot, blockRoot}
```
calbera, line 90: we can simplify and get rid of the `if !ok`

```
pid, ok := p.slotToBlockRootToPayloadID[payloadIDCacheKey{slot, blockRoot}]
return pid, ok

File: payload/builder/payload.go [hunk 8]
```diff
@@ -205,7 +229,7 @@ func (pb *PayloadBuilder) getPayload(
 		return nil, err
 	}
 	if envelope == nil {
-		return nil, ErrNilPayloadEnvelope
+		return nil, ErrNilPayloadEnvelope // appease linter. This is checked already
```
calbera: this entire nil check (and the `ErrNilPayloadEnvelope` error which is unused elsewhere) can be removed because envelope is already guaranteed to be non-nil from [here](https://github.com/berachain/beacon-kit/blob/clear-payloadID-cache-2/execution/client/engine.go#L147-L150).

abi87: Agree, nilaway seems fine with dropping this. Dropped

payload/cache/payload_id.go
abi87: I would keep it as is, it's natural as querying the map, unless you feel this strongly.

beacon/validator/block_builder.go
abi87: Done above

PR #2967, opened 2025-10-23
Fix high cardinality metrics by removing unbounded labels

fridrik01:
This PR addresses a metrics issue by removing unbounded labels from Prometheus metrics that can cause cardinality explosion and degrade system performance in our monitoring systems.

I believe it should be fine to remove these labels as we should have information about them in our logs/loki.

Note: if any Grafana dashboards relied on these labels, they may need to be updated to reflect the new metric structure.

### DIFF
--- a/beacon/blockchain/deposit.go
+++ b/beacon/blockchain/deposit.go
@@ -24,7 +24,6 @@ import (
 	"context"
 	"maps"
 	"slices"
-	"strconv"
 	"time"
 
 	"github.com/berachain/beacon-kit/primitives/math"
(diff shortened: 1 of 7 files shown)


beacon/blockchain/deposit.go
```diff
@@ -55,11 +54,10 @@ func (s *Service) fetchAndStoreDeposits(
 	ctx context.Context,
 	blockNum math.U64,
 ) {
-	blockNumStr := strconv.FormatUint(blockNum.Unwrap(), 10)
 	deposits, err := s.depositContract.ReadDeposits(ctx, blockNum, blockNum)
 	if err != nil {
 		s.logger.Error("Failed to read deposits", "error", err)
-		s.metrics.FailedToGetBlockLogs.With("block_num", blockNumStr).Add(1)
+		s.metrics.FailedToGetBlockLogs.Add(1)
```
abi87, line 60: q: with this change, we should be able to get the time when the failure happens and retrieve the corresponding logs correct?
If so, I am all for this change

beacon/blockchain/payload.go
```diff
@@ -223,7 +222,7 @@ func (s *Service) handleRebuildPayloadForRejectedBlock(
 
 	s.latestFcuReq.Store(&buildData.FCState)
 
-	s.metrics.markRebuildPayloadForRejectedBlockSuccess(nextBlkSlot)
+	s.metrics.markRebuildPayloadForRejectedBlockSuccess()
```
abi87, line 225: just make sure we duly log where we are making the metrics less expressive

abi87 (commented):
would just double check we have all the loggings we need, where we are dropping metrics.
This is a very important PR to save costs, would encourage @calbera and @rezbera to double check it and merge it asap

PR #2988, opened 2025-11-24
Remove orphaned blobs on started after incomplete block finalization

fridrik01:
This PR fixes potential orphaned blob sidecars on disk from incomplete block finalization.

If `FinalizeBlock` fails after `FinalizeSidecars` has persisted blob sidecars to disk, we may leave behind orphaned blobs at slot `lastBlockHeight + 1`. These orphaned blobs:
  - Consume disk space unnecessarily
  - Represent incomplete/invalid state since no corresponding block was finalized
  - Persist across node restarts

This PR addresses this so that on node startup (during CometBFT service initialization), check for and remove any orphaned blob sidecars at slot `lastBlockHeight + 1`. 

**Test plan**

I changed finalizeBlock so it a block has blobs it returns an error after running `FinalizeSidecars`. I then started a local node and ran `spamoor` to send blob tx. I observed beacon-kit panicked with "CONSENSUS FAILURE!!!". When starting the node again the blob saved in `FinalizeSidecars` was removed successfully

```
# start local node
make start
make start-geth

# send blobs
./bin/spamoor blobs --privkey=0xfffdbb37105441e14b0ee6330d855d8504ff39e705c3afa8f859ac9865f99306 --rpchost=http://localhost:8545 --count=10 --sidecars=3 --throughput=1

# observed panic in beacon-kit as described above:
2025-11-27T13:55:39Z INFO Successfully stored all blob sidecars 🚗 service=da-store slot=7 num_sidecars=3
2025-11-27T13:55:39Z WARN We are finalizing block with sidecars, lets return error!
2025-11-27T13:55:39Z ERRR Error in proxyAppConn.FinalizeBlock module=state err=some random error
2025-11-27T13:55:39Z ERRR CONSENSUS FAILURE!!! module=consensus err=failed to apply block; error some random error

# confirm the node saved blobs for slot 7:
~/workspace/beacon-kit % ll .tmp/beacond/data/blobs/7
total 792
-rw-r--r--@ 1 fridrikasmundsson  staff  131928 Nov 27 13:55 0x807cb903e1f1f391c2a53e47d4b90772b88ba6a7c43639476c3c78302b20a0de5a6c004ccf6d4b676b111e054dcab339.ssz
-rw-r--r--@ 1 fridrikasmundsson  staff  131928 Nov 27 13:55 0x9516a34fc043963028a30afc14786cb57f276c05c4a4a20a6e65e599d851677648e710aa3cdb00d996ead59678063079.ssz
-rw-r--r--@ 1 fridrikasmundsson  staff  131928 Nov 27 13:55 0xa00a29bb78412ae38de1008979caa1169bdf5ea37b36400600340314a78190840716d970231707bc4496b80f868e0c41.ssz

#  when starting node again I saw
2025-11-27T13:56:14Z WARN enable state pruning to reduce memory footprint considerably strategy=nothing recommended strategies=everything or default
2025-11-27T13:56:14Z WARN Found orphaned blob sidecars from incomplete block finalization, removing service=blockchain slot=7 num_sidecars=3
2025-11-27T13:56:14Z INFO Successfully removed orphaned blob sidecars service=blockchain slot=7

# and blobs were cleaned up:
ll .tmp/beacond/data/blobs/7
ls: .tmp/beacond/data/blobs/7: No such file or directory
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -56,7 +56,7 @@ require (
 
 // build/test dependencies
 require (
-	github.com/attestantio/go-eth2-client v0.27.1
+	github.com/attestantio/go-eth2-client v0.27.2
 	github.com/cosmos/ics23/go v0.11.0
 	github.com/ethereum/go-ethereum v1.15.5
 	github.com/ferranbt/fastssz v0.1.5-0.20240903094032-455b54c08c81
--- a/go.sum
+++ b/go.sum
@@ -67,8 +67,8 @@ github.com/allegro/bigcache v1.2.1-0.20190218064605-e24eb225f156 h1:eMwmnE/GDgah
 github.com/allegro/bigcache v1.2.1-0.20190218064605-e24eb225f156/go.mod h1:Cb/ax3seSYIx7SuZdm2G2xzfwmv3TPSk2ucNfQESPXM=
 github.com/antihax/optional v1.0.0/go.mod h1:uupD/76wgC+ih3iEmQUL+0Ugr19nfwCT1kdvxnR2qWY=
 github.com/armon/go-metrics v0.4.1/go.mod h1:E6amYzXo6aW1tqzoZGT755KkbgrJsSdpwZ+3JqfkOG4=
-github.com/attestantio/go-eth2-client v0.27.1 h1:g7bm+gG/p+gfzYdEuxuAepVWYb8EO+2KojV5/Lo2BxM=
-github.com/attestantio/go-eth2-client v0.27.1/go.mod h1:fvULSL9WtNskkOB4i+Yyr6BKpNHXvmpGZj9969fCrfY=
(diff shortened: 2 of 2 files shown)


consensus/cometbft/service/service.go
```diff
@@ -182,6 +182,11 @@ func NewService(
 		}
 	}
 
+	// Clean up any orphaned blob sidecars from incomplete block finalization.
+	if err = s.Blockchain.PruneOrphanedBlobs(lastBlockHeight); err != nil {
```
calbera, line 186: PR lgtm. 

Just a question @fridrik01, is it true that the blobs for the last block height on disk will never be needed again? even if the finalization for that height was done correctly. Just want to make sure that the removal of blobs is safe.

fridrik01: @calbera  Yes, the removal is safe, the `PruneOrphanedBlobs` only removes blobs at slot `lastBlockHeight + 1`, which is one height ahead of the last finalized block.

calbera: thanks!

PR #2991, opened 2025-11-27
Bound the cachedStates cache

fridrik01:
This PR sets a bound the `candidateStates` cache to prevent unbounded memory growth during prolonged consensus failures. 

The cache  was added in https://github.com/berachain/beacon-kit/pull/2812 to make `FinalizeBlock` a no-op. It stores executed block states between `ProcessProposal` and `FinalizeBlock` and was previously an unbounded map that could grow if consensus rounds repeatedly failed before reaching +2/3 prevotes. 

This PR addresses this by capping this cache (currently set at 10 entries) by using a LRU (which was already a dependency)

### DIFF
--- a/.mockery.yaml
+++ b/.mockery.yaml
@@ -90,3 +90,8 @@ packages:
       recursive: False
       with-expecter: true
       include-regex: GenesisStateProcessor
+  github.com/berachain/beacon-kit/node-core/types:
+    config:
+      recursive: False
+      with-expecter: true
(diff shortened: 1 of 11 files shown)


calbera (commented):
The logic here all LGTM. I think we should take this opportunity to add metrics here, i.e. we only really need to track how big the cache is when Reset is called. So i would just add a counter for cachedStates size in the Reset function right before we call `cs.states.Purge()`.

PR #3068, opened 2026-03-27
feat(preconf): rebuild sequencer payload on consensus round change to keep flashblocks flowing

fridrik01:
When consensus goes to multiple rounds, the sequencer only builds a payload and emits flashblocks for round 0. If round 0 times out and a new proposer is selected, no fresh flashblocks are produced for any subsequent rounds. RPC nodes continue serving stale round-0 state, and the new proposer receives the same cached payload with no new transactions.

Detecting the round change in `ProcessProposal` is too late. By that point the new proposer has already fetched and proposed. There is also no round information in the ABCI `PrepareProposalRequest` or `ProcessProposalRequest`.

This PR aims to circumvent this by subscribing to CometBFT's local `EventNewRound` message through its message bus on the sequencer. This event should fires before the new proposer enters `PrepareProposal`, giving the sequencer enough time to invalidate the stale payload cache and trigger a fresh optimistic build with up-to-date transactions.

Note that no `bera-reth` changes should be needed because when a new index-0 flashblock arrives with a different payload ID, the RPC node's flashblock service discards the stale round-0 sequence and starts tracking the new one.

**Test plan**

The following e2e test stops a single validator to trigger a multi round and verifies that cometbft on sequencer sent a new round message and that flashblocks were emitted for round1:

```
go test -timeout 200s -tags e2e,bls12381,test ./testing/e2e/preconf/. -v -run TestPreconfE2ESuite/TestRoundChangeRebuild
```


bar-bera (commented):
lgtm, just maybe we should rename the existing `round_rotation_test.go` to something more appropriate since now `round_change_test.go` tests the detection of the round ratation, while the former tests the behavior of the network under the round rotation condition. or maybe they could be merged

consensus/cometbft/service/round_change.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-sub.Canceled():
+				s.logger.Warn("NewRound subscription cancelled", "error", sub.Err())
+				return
```
calbera: when does cometbft cancel subscriptions? and should we retry in this case if node is still continuing?

beacon/blockchain/round_change.go
```diff
@@ -0,0 +1,92 @@
...
+//
+// The proposerAddress is the CometBFT address of the new round's proposer.
+// The ctx should be an SDK context backed by a fresh committed multistore.
+func (s *Service) HandleRoundChange(
+	ctx context.Context,
+	height int64,
+	round int32,
+	proposerAddress []byte,
+) {
+	if !s.preconfCfg.IsSequencer() || !s.localBuilder.Enabled() {
```
calbera, line 40: Why keep the localBuilder enabled check? This should only run if its the sequencer correct?

consensus/cometbft/service/service.go
```diff
@@ -249,7 +249,15 @@ func (s *Service) Start(
...
 
-	return err
+	if err != nil {
+		return err
+	}
+
+	// Subscribe to round-change events so the sequencer can detect when
+	// a consensus round times out and rebuild the payload for the new proposer.
+	s.subscribeToRoundChanges(ctx)
+
```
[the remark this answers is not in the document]
calbera: Does feel wasteful that every node will run this subscription and listener goroutine but then not act on it unless its sequencer.. can we move the config check of ispreconfSqr to cometbft service?

beacon/blockchain/round_change.go
calbera: maybe somewhere upstream in node startup / config validation, we should enforce that if isSequencer, localBuilder must be enabled

consensus/cometbft/service/round_change.go
```diff
@@ -0,0 +1,76 @@
...
+				if !ok || event.Round < 1 {
+					continue
+				}
+				s.logger.Info("CometBFT round change detected",
+					"height", event.Height,
+					"round", event.Round,
+					"proposer_address", event.Proposer.Address,
+				)
+
+				freshState := s.resetState(ctx)
```
calbera: just wondering, why do we need to call resetState here?

consensus/cometbft/service/round_change.go
```diff
@@ -0,0 +1,76 @@
...
+	cmttypes "github.com/cometbft/cometbft/types"
+)
+
+const roundChangeSubscriber = "beacon-kit-round-change"
+
+// subscribeToRoundChanges subscribes to CometBFT's EventNewRound on the
+// local event bus and calls Blockchain.HandleRoundChange for rounds > 0.
+// This allows the sequencer to detect consensus round timeouts and trigger
+// a fresh payload build for the new proposer.
+func (s *Service) subscribeToRoundChanges(ctx context.Context) {
```
calbera: technically this function subscribes _and_ listens. So maybe change the name or split into 2 functions. itll be a little clearer where the listen goroutine is running.

calbera (approved):
mostly makes sense, utACK. 

EDIT: I see the e2e test now, but would it be possible to also send txs during this round change and observe that the newly built payload includes these new txs?

consensus/cometbft/service/service.go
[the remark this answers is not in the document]
fridrik01: done!

consensus/cometbft/service/round_change.go
fridrik01: It should rarely if ever happen, unless we stop draining the channel. I added a simple retry mechanism that logs if this ever happens

beacon/blockchain/round_change.go
fridrik01: done!

consensus/cometbft/service/round_change.go
fridrik01: I split the subscription/dispatch code with the handling of the message

consensus/cometbft/service/round_change.go
fridrik01: Because beacondb reads need an sdk.Context with a multistore embedded in the ctx (otherwise the unwrap returns nothing), its the same thing we do in prepare and process proposal. Thougoh this is always complex to read and reason about

fridrik01:
>EDIT: I see the e2e test now, but would it be possible to also send txs during this round change and observe that the newly built payload includes these new txs?

Updated the e2e test so that we start a background routine that sends txs after first payload and then verifies that at least 1 gets included in the next payload

calbera (approved):

PR #3069, opened 2026-03-31
feat(preconf): verify requesting validator is the expected proposer before serving payloads

fridrik01:
This PR adds proposer verification to the preconf server on the sequencer so that only the current expected proposer can retrieve payloads. It extends the `EventBus` round-change mechanism from #3068 by introducing a simple `ProposerTracker` that records which validator should be proposing for a given slot, updated on every `ProcessProposal` and `HandleRoundChange`. 

The sequencer then checks the `ProposerTracker` on each `GetPayload` request and rejects stale proposers making sure that only the proposing validator for that round is allowed to be served that payload.

Implicitly covered by the e2e `TestRoundChangeRebuild` test from #3068, which forces consensus round timeouts

### DIFF
--- a/beacon/blockchain/process_proposal.go
+++ b/beacon/blockchain/process_proposal.go
@@ -329,6 +329,9 @@ func (s *Service) VerifyIncomingBlock(
 				s.logger.Error("Failed to get next proposer pubkey", "error", pubkeyErr)
 			} else {
 				shouldBuildNextPayload = s.preconfWhitelist.IsWhitelisted(expectedProposerPubkey)
+				// Record expected proposer so the preconf server can verify
+				// that payload requests come from the correct validator.
+				s.preconfProposerTracker.SetExpectedProposer(blkSlot+1, expectedProposerPubkey)
 			}
 			s.logger.Info("Sequencer mode: determined next proposer",
 				"current_block_slot", blkSlot.Base10(),
--- a/beacon/blockchain/service.go
+++ b/beacon/blockchain/service.go
@@ -84,8 +84,7 @@ type Service struct {
 	// preconfProposerTracker tracks the expected proposer pubkey for upcoming slots.
 	// The sequencer writes entries here so the preconf server can verify that
 	// payload requests come from the correct proposer.
-	// Can be nil if preconf is disabled.
-	preconfProposerTracker *preconf.ProposerTracker
+	preconfProposerTracker preconf.ProposerTracker
 }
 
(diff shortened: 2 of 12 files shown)


beacon/blockchain/process_proposal.go
```diff
@@ -329,6 +329,11 @@ func (s *Service) VerifyIncomingBlock(
 				s.logger.Error("Failed to get next proposer pubkey", "error", pubkeyErr)
 			} else {
 				shouldBuildNextPayload = s.preconfWhitelist.IsWhitelisted(expectedProposerPubkey)
+				// Record expected proposer so the preconf server can verify
+				// that payload requests come from the correct validator.
+				if s.preconfProposerTracker != nil {
```
calbera: consider making this nil check inside the preconfProposerTracker as to avoid this check in the calling flow.

beacon/preconf/server.go
```diff
@@ -193,6 +196,16 @@ func (s *Server) handleGetPayload(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Verify the requesting validator is the expected proposer for this slot.
```
calbera, line 199: I'm not quite understanding why this is necessary? is it just some DOS protection? like we think validators could spam the sequencer with getPayloads?

fridrik01: Two main reasons: making sure we reject a stale proposer after round changes  and prevent non-proposing validators from setting details about the upcoming block.

beacon/blockchain/service.go
```diff
@@ -80,6 +80,12 @@ type Service struct {
 	// - Validator: checks if self is whitelisted to fetch payload from sequencer
 	// Can be nil if preconf is disabled.
 	preconfWhitelist preconf.Whitelist
+
+	// preconfProposerTracker tracks the expected proposer pubkey for upcoming slots.
+	// The sequencer writes entries here so the preconf server can verify that
+	// payload requests come from the correct proposer.
+	// Can be nil if preconf is disabled.
+	preconfProposerTracker *preconf.ProposerTracker
```
calbera: if this is nil, wouldnt this panic on all the calls?

fridrik01: It should never be nil since DI always returns a non-nil value. However this comment is wrong, and just changed this to be an interface instead (like the whitelist)

bar-bera (approved):
lgtm

PR #3072, opened 2026-04-01
chore: upgrade go version and add deps badge to toplevel readme

fridrik01:
We removed the dependabot in https://github.com/berachain/beacon-kit/pull/3032 and replaced it with daily vulncheck workflow. That workflow does 3 things:
 1. run `make vulncheck` to for daily symbol-level vulnerability scan
 2. runs unit tests against locked deps 
 3. runs unit test against latest deps (it bumps all version deps) to catch breakage early 

For the past week the 3) has been failing since the updated dependencies require go 1.26. This PR addressess that by:
- upgrading the go versino 
- filtering out some direct deps we know don't work (cosmossdk, kurtosis, etc) and only updating all other deps

I also added a rep badge to top level readme which lets us know if this build is failing (we can always check it in https://github.com/berachain/beacon-kit/actions/workflows/vuln-and-dep-check.yml, also I want to avoid spamming us on slack unless necessary) so this is a nice comprimise


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/berachain/beacon-kit
 
-go 1.25.8
+go 1.26.1
```
calbera, line 3: should update the README too

calbera (approved):

PR #3073, opened 2026-04-02
fix(payload): enforce the fork version of cached verified payload is consistent

calbera:
Ensures using the cached latest verified payload when building a block is consistent even across fork boundary. If we go to multiple consensus rounds and that happens to cross over the new fork, we should not use this cached payload and instead build just in time/


fridrik01 (approved):
LGTM

payload/builder/payload_test.go
bar-bera: maybe this todo could be addressed in this PR since already touching this file

bar-bera (approved):
LGTM, just left a comment on minor possible improvement in the test file

PR #3079, opened 2026-04-14
fix(ci): use go `run` instead of `install` for vulncheck

calbera:
The vulncheck CI job started failing after new Go stdlib vulnerabilities (GO-2026-4864, GO-2026-4865, etc.) were published on 2026-04-07. These affect go1.26.1 and are fixed in go1.26.2. Bumping the Go version resolves all reported stdlib vulnerabilities.

Also switch the vulncheck Make target from `go install` + bare command to `go run` with GOTOOLCHAIN pinning so it works locally without requiring $GOPATH/bin in PATH.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -29,6 +29,7 @@ on:
   push:
     branches:
       - main
+      - preconf-dev
     tags:
       - "v*"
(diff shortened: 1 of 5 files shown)


bar-bera (approved):
lgtm

PR #3081, opened 2026-04-16
feat(chain): implement BRIP-0008 hysteresis update for Fulu fork

calbera:
https://github.com/berachain/BRIPs/blob/main/meta/BRIP-0010.md

Add the Fulu fork (CL component of the Fusaka hardfork) with:

- BRIP-0008: Fork-gated hysteresis parameter updates that improve validator capital efficiency by reducing the upward buffer from 125% to ~110% of the effective balance increment (HysteresisQuotient: 4->100, UpwardMultiplier: 5->10, DownwardMultiplier: 1->1 unchanged).

- PoL vNext: New EVMInflationAddressFulu and EVMInflationPerBlockFulu fields following the same pattern as the Genesis-to-Deneb1 inflation migration. Values are placeholders pending PoL vNext design finalization.

- Fulu fork plumbing: version constant (0x06000000), fork time in chain spec, fork ordering validation, state upgrade handler, Engine API routing, consensus type support, and osakaTime in EL genesis configs.

All devnet/testnet/mainnet chain specs and TOML configs updated with placeholder values (devnet: active from genesis, testnet/mainnet: far future).

### DIFF
--- a/chain/spec.go
+++ b/chain/spec.go
@@ -48,15 +48,18 @@ type HysteresisSpec interface {
 	// HysteresisQuotient returns the quotient used in effective balance
 	// calculations to create hysteresis. This provides resistance to small
 	// balance changes triggering effective balance updates.
-	HysteresisQuotient() math.U64
+	// The value is fork-gated by timestamp (updated in Fulu per BRIP-0008).
+	HysteresisQuotient(timestamp math.U64) math.U64
 
 	// HysteresisDownwardMultiplier returns the multiplier used when checking
 	// if the effective balance should be decreased.
-	HysteresisDownwardMultiplier() math.U64
+	// The value is fork-gated by timestamp (updated in Fulu per BRIP-0008).
+	HysteresisDownwardMultiplier(timestamp math.U64) math.U64
 
 	// HysteresisUpwardMultiplier returns the multiplier used when checking
 	// if the effective balance should be increased.
-	HysteresisUpwardMultiplier() math.U64
+	// The value is fork-gated by timestamp (updated in Fulu per BRIP-0008).
+	HysteresisUpwardMultiplier(timestamp math.U64) math.U64
 }
 
 type DepositSpec interface {
@@ -109,6 +112,9 @@ type ForkSpec interface {
 
 	// Electra1ForkTime returns the time at which the Electra1 fork takes effect.
 	Electra1ForkTime() uint64
+
+	// FuluForkTime returns the time at which the Fulu fork takes effect.
+	FuluForkTime() uint64
 }
 
 type BlobSpec interface {
@@ -269,6 +275,7 @@ func (s spec) validate() error {
 		s.Data.Deneb1ForkTime,
 		s.Data.ElectraForkTime,
 		s.Data.Electra1ForkTime,
+		s.Data.FuluForkTime,
 	}
 	for i := 1; i < len(orderedForkTimes); i++ {
 		prev, cur := orderedForkTimes[i-1], orderedForkTimes[i]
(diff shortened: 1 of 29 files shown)


state-transition/core/state_processor_forks.go
```diff
@@ -268,3 +279,50 @@ func (sp *StateProcessor) logElectra1Fork(
...
+	fork.Epoch = sp.cs.SlotToEpoch(slot)
+	if err := st.SetFork(fork); err != nil {
+		return err
+	}
+
+	// Initialize the pending partial withdrawals to an empty array if not already initialized.
+	// This handles the case where the chain starts directly on Fulu (e.g., devnet).
+	if _, err := st.GetPendingPartialWithdrawals(); errors.Is(err, collections.ErrNotFound) {
+		sp.metrics.gaugePartialWithdrawalsEnqueued(0)
+		return st.SetPendingPartialWithdrawals([]*types.PendingPartialWithdrawal{})
```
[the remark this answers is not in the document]
calbera: this is how we check if the item exists in cosmos-sdk's collections store, logic didn't change in this PR

state-transition/core/state_processor_forks.go
```diff
@@ -268,3 +279,50 @@ func (sp *StateProcessor) logElectra1Fork(
...
+	}
```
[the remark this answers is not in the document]
calbera: this is how we check if the item exists in cosmos-sdk's collections store, logic didn't change in this PR

state-transition/core/state_processor.go
```diff
@@ -390,11 +390,19 @@ func (sp *StateProcessor) processEffectiveBalanceUpdates(st *state.StateDB) erro
 		return err
 	}
 
+	// Get the timestamp from the latest execution payload header to determine the active fork
+	// version for fork-gated hysteresis parameters (BRIP-0008).
+	lph, err := st.GetLatestExecutionPayloadHeader()
+	if err != nil {
+		return err
+	}
+	timestamp := lph.GetTimestamp()
```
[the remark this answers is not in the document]
calbera: Commented about the one epoch delay [here](https://github.com/berachain/beacon-kit/pull/3081/commits/a29b833f9fd4eebce319e8079c05556e67cf5c18)

chain/spec.go
```diff
@@ -518,10 +537,12 @@ func (s spec) ValidatorSetCap() uint64 {
 // inflation amount of native EVM balance through a withdrawal every block.
 func (s spec) EVMInflationAddress(timestamp math.U64) common.ExecutionAddress {
 	fv := s.ActiveForkVersionForTimestamp(timestamp)
-	switch fv {
-	case version.Deneb1(), version.Electra(), version.Electra1():
+	switch {
+	case version.EqualsOrIsAfter(fv, version.Fulu()):
```
fridrik01: Should we not check for exact version here? If we add Fulu1 (or future fork) then we would silently return here instead of panicking which is how the switch was written explicitly before.

```suggestion
	case version.Equals(fv, version.Fulu()):
```

chain/spec.go
```diff
@@ -48,15 +48,17 @@ type HysteresisSpec interface {
 	// HysteresisQuotient returns the quotient used in effective balance
 	// calculations to create hysteresis. This provides resistance to small
 	// balance changes triggering effective balance updates.
-	HysteresisQuotient() math.U64
+	// The value is fork-gated by timestamp (updated in Fulu per BRIP-0008).
+	HysteresisQuotient(timestamp math.U64) math.U64
 
 	// HysteresisDownwardMultiplier returns the multiplier used when checking
 	// if the effective balance should be decreased.
 	HysteresisDownwardMultiplier() math.U64
```
fridrik01, line 57: Looking at the code its unclear why HysteresisUpwardMultiplier takes an argument but HysteresisDownwardMultiplier does not. Since HysteresisDownwardMultiplier does not change in the fork there is no need to, but maybe add an inline comment (or pass a unused timetsamp)

fridrik01 (approved):
mostsly nits

chain/spec.go
calbera: Added comment in https://github.com/berachain/beacon-kit/pull/3081/commits/3cdc1385ef932032b25a9bd6171ab090aede69c9

chain/spec.go
calbera: Well I'd argue it's more correct now. For example the value we set in Deneb1 `EVMInflationAddressDeneb1` is actually valid from fork Deneb1 and onwards (including Electra and Electra1) until its overriden with a new value. Which is now `EVMInflationAddressFulu`; this value is valid until overriden again. So for Fulu1 it would also be used by default.

Also adjusted other functions to work like this in https://github.com/berachain/beacon-kit/pull/3081/commits/4d88d47bdb168fd0dfeefd401ebf95052d0cbcd2

fridrik01: I understand that, but I am concerned there may be a slight chance that we miss updating all call sites when adding a non compatible fork. Before this PR, we would see a panic immedietly and then have to evaluate if its valid to current handling (adding the fork to the case) , or whether it needs a separate handling (like here for fulu).

calbera: When a chain spec value needs to be updated via hard fork, there **needs** to be an explicit change via a PR. Hence why we don't create a new variable for every value of the chain spec in every new fork.

If a chain spec value is being called on an **unsupported** fork version, then we should panic. 

IMO this convention means we actually always implicitly "evaluate" the necessary conditions.

all that being said, its pretty nitpicky. We can make this change in another PR throughout the repo to maintain the same standard everywhere. Will revert back to old way on this PR for now.

PR #3084, opened 2026-04-17
feat(preconf): add observability metrics for sequencer and validator flows

fridrik01:
This PR adds observability metrics for preconf sequencer/validator flows:
- Sequencer-side: request outcomes, proposer-check results
- Validator-side: fetch outcomes, fallback reasons, sequencer-available gauge
- Blockchain: round-change counter
- Internal errors now return `500` (was `404`) so HTTP status matches the metric label

### DIFF
--- a/beacon/preconf/server_metrics.go
+++ b/beacon/preconf/server_metrics.go
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: BUSL-1.1
+//
+// Copyright (C) 2025, Berachain Foundation. All rights reserved.
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file of this repository and at www.mariadb.com/bsl11.
+//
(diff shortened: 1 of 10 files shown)


beacon/preconf/server_metrics.go
```diff
@@ -0,0 +1,62 @@
...
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, NON-INFRINGEMENT, AND
+// TITLE.
+
+package preconf
+
+import "strconv"
+
+// Preconf server payload request outcomes used as the `result` label on
+// beacon_kit.preconf.server.payload_request_total.
+const (
```
calbera, line 27: nit: could make this a typed string `type ServerResult string` and just have the value constants be `OK`, `Unauthorized`, etc..

`fetchAndWritePayload` would then return `ServerResult`

calbera (approved):

PR #3086, opened 2026-04-20
chore(preconf): enrich sequencer server health check impl.

bar-bera:
Uses existent services to add checks on the `/eth/v1/preconf/health` endpoint of the sequencer server: 
* node has a state to work with
* node is synced
* EL is connected


beacon/preconf/server.go
```diff
@@ -157,13 +177,49 @@ func (s *Server) Stop() error {
...
+		resp.SyncDistance = syncToHeight - latestHeight
+		if resp.SyncDistance < 0 {
+			resp.SyncDistance = 0
+		}
+		resp.IsSyncing = resp.SyncDistance > 0
+	}
+
+	if s.elChecker != nil {
+		resp.ELConnected = s.elChecker.IsConnected()
+	}
```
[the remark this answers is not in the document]
bar-bera: fair b4889299

beacon/preconf/server.go
```diff
@@ -157,13 +177,42 @@ func (s *Server) Stop() error {
...
+}
+
+// buildHealthResponse inspects the node's sync state and EL connectivity
+// and produces a HealthResponse.
+func (s *Server) buildHealthResponse() *HealthResponse {
+	resp := new(HealthResponse)
+
+	if s.syncChecker != nil {
+		resp.IsReady = s.syncChecker.IsAppReady() == nil
```
calbera, line 213: IsAppReady should return a boolean (in general any func thats named `Is<something>` should return bool) since the error value is never used.

calbera (commented):
one nit but mostly looks good

beacon/preconf/server.go
bar-bera: agree it's a bit weird but this is just the `ConsensusService` relaying the comet interface. I would avoid to wrap it with an extra layer

beacon/preconf/server.go
```diff
@@ -61,13 +61,29 @@ type PayloadProvider interface {
 	GetPayloadBySlot(ctx context.Context, slot math.Slot, parentBlockRoot common.Root) (ctypes.BuiltExecutionPayloadEnv, error)
 }
 
+// SyncChecker exposes the node's sync status for health checks.
+type SyncChecker interface {
+	// IsAppReady returns nil if the node has committed at least one block.
+	IsAppReady() error
```
calbera: nit: to make it clear from reading this code, we can add a quick comment here explaining:
- why/how we use the error 
- why its not overriden (bc its part of the comet service already)

calbera (approved):
lgtm, once e2e passes

calbera (approved):
fix merge conflicts then lgtm

PR #3090, opened 2026-04-22
fix(engine): retry on EL connection errors to survive EL restarts

### LINKED ISSUE #3083: Cycling EL wedges Beacon-Kit
Observed with Bera-Reth 1.3.3 and BK 1.3.8

Steps to reproduce:
1. Start Beacon-Kit and bera-reth as normal.  Allow to begin syncing.
2. Restart bera-reth.
3. Overall system does not resume syncing

fridrik01:
Fixes #3083. Restarting the EL (e.g. `bera-reth`) left `beacond` unable to recover on its own and the beacon node had to be restarted too.

The root cause was that `ErrBadConnection` was classified as fatal and wrapped in `backoff.Permanent`, so the retry loop (configured for infinite retries) bailed on the first attempt. `FinalizeBlock` returned the error to CometBFT and consensus could not progress.

This PR addresses this by reclassifying `ErrBadConnection` as non-fatal. A dropped TCP connection is exactly the retryable case the existing backoff was built for. The retry loop now waits out the outage and resumes once the EL is back. Truly fatal errors (JSON-RPC predefined errors, invalid payload status) stay fatal.

**Test Plan**
- Unit test pinning the classification of `ErrBadConnection`.
- Simulated test that drops the TCP connection mid-flight and asserts `FinalizeBlock` recovers once the EL returns.
- Also removed `TestProcessProposal_CrashedExecutionClient_Errors`. It pinned the old fast-fail behavior and is now redundant with the unit test and the new connection-drop recovery test.
- Tested this on local kurtosis devnet and obvserved that beacond successfully resumed when bera-reth was restarted

### DIFF
--- a/execution/client/errors.go
+++ b/execution/client/errors.go
@@ -123,10 +123,7 @@ func (s *EngineClient) handleRPCError(
 // IsFatalError defines errors that indicate a bad request or an otherwise
 // unusable client.
 func IsFatalError(err error) bool {
-	return jsonrpc.IsPreDefinedError(err) || errors.IsAny(
-		err,
-		ErrBadConnection,
-	)
+	return jsonrpc.IsPreDefinedError(err)
 }
 
 // IsNonFatalError defines errors that should be ephemeral and can be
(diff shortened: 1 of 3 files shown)


calbera (approved):
utACK. if it solves the EL restart issue, no problem in treating it non-fatal

PR #3094, opened 2026-04-28
chore(fusaka): correct eth genesis configs

calbera:


node-core/services/version/version.go
```diff
@@ -153,6 +153,7 @@ func (rs *ReportingService) printToConsole(
 	+ 🍴 Deneb1 Fork Time: %-52d+
 	+ 🍴 Electra Fork Time: %-51d+
 	+ 🍴 Electra1 Fork Time: %-50d+
+	+ 🍴 Fulu Fork Time: %-49d+
```
fridrik01: Confirm that `%-49d+` is correct here

calbera: you're right it should be 54

calbera: fixed in https://github.com/berachain/beacon-kit/pull/3094/commits/238b8e550ed0476a7ee285eac4c29d7218211bea

image

fridrik01 (approved):

PR #3096, opened 2026-04-30
fix(da): guard against nil inner BeaconBlockHeader in BlobSidecar decode

calbera:
Karalabe SSZ permits the inner Header pointer of a SignedBeaconBlockHeader to decode to nil when the wire payload encodes that field as zero-length. The existing nil-fill in ValidateAfterDecodingSSZ only initialized the outer SignedBeaconBlockHeader, leaving the inner Header pointer nil and allowing a crafted gossiped sidecar to panic any node that reaches da/store.Persist (GetBeaconBlockHeader().GetSlot()).

Extend the same zero-fill mitigation one level deeper so downstream accessors are nil-safe. Adds a regression test covering both the outer and inner nil shapes.


fridrik01 (approved):
LGTM

PR #3098, opened 2026-05-04
chore(fusaka): Set fulu fork time for bepolia

calbera:
Bepolia HF to Fusaka will occur at unix `1779897600` May 27th 16:00:00 UTC. 

Also bump's comet dep after https://github.com/berachain/cometbft/pull/41


bar-bera (approved):
lgtm

PR #3102, opened 2026-05-11
fix(node-api): unbounded filter input

bar-bera:
Unbounded input in API request
  * `POST /eth/v1/beacon/states/{state_id}/validators`
  * `POST /eth/v1/beacon/states/{state_id}/validator_balances`

causes O(N*M) algorithmic complexity DoS with
  * N validator registry (not set) size
  * M attacker slice size (indices or pubkeys)

With Berachain registry size < 100 impact is minimal.

  ### Fix
  * Cap on inputs
  * Filter implementation by mapping instead of arrays.


node-api/handlers/beacon/types/limits.go
```diff
@@ -0,0 +1,31 @@
...
+
+// MaxValidatorIDsPerRequest caps the IDs[] filter on /validators and
+// /validator_balances to bound per-request work.
+// Mirrored as a literal in request.go struct tags (max=1024).
+const MaxValidatorIDsPerRequest = 1024
+
+// MaxValidatorStatusesPerRequest is a sanity bound; 9 valid statuses exist
+// (see ValidateValidatorStatus).
+// Mirrored as a literal in request.go struct tags (max=16).
+const MaxValidatorStatusesPerRequest = 16
```
fridrik01: Can we remove these constants? Changing them does not change what is hardcoded in the validation inside the struct, only in the test. We should just move these inside the test itself

bar-bera: [ea051d2](https://github.com/berachain/beacon-kit/pull/3102/commits/ea051d2613c53768a1b3acb5da8ec5814e9aae02)

fridrik01 (approved):
LGTM

PR #3104, opened 2026-05-11
chore: do not ignore error on deposits ingestion iterator

bar-bera:
Fixes silent truncation and processing of `partialDeposits` when iteration aborts mid-stream.

Likelyhood for this to happen is basically none, still the fix is trivial and improving the code correctness.


fridrik01 (approved):
LGTM

PR #3109, opened 2026-05-20
fix(engine): classify EL HTTP 4xx as fatal via typed HTTPStatusError

fridrik01:
This PR Introduces a typed `rpc.HTTPStatusError` so the EL's HTTP status code propagates through to `handleRPCError`. Previously, `callRaw` collapsed every non-200 response into `fmt.Errorf("unexpected status code %d: ...")`, which didn't satisfy the `jsonrpc.Error` interface and so fell into the catch-all `ErrBadConnection` (retryable) bucket.

With the typed error in place, `handleRPCError` now classifies HTTP 4xx as fatal (`ErrHTTPClientError`) and leaves 5xx and other transport failures retryable (`ErrBadConnection`), so  the engine no longer retries forever on rejections the EL won't accept on a second try.

Will likely made a follow-up PR which will split the engine retry policy between `ProcessProposal` (bounded retry where a failing proposal should be rejected so consensus can move on) and `FinalizeBlock` (unbounded retry where the block is already agreed, must be applied).

### DIFF
--- a/execution/client/client.go
+++ b/execution/client/client.go
@@ -24,6 +24,7 @@ import (
 	"context"
 	"fmt"
 	"math/big"
+	nethttp "net/http"
 	"sync"
 	"time"
 
@@ -32,7 +33,6 @@ import (
 	ethclientrpc "github.com/berachain/beacon-kit/execution/client/ethclient/rpc"
 	"github.com/berachain/beacon-kit/log"
 	"github.com/berachain/beacon-kit/primitives/math"
-	"github.com/berachain/beacon-kit/primitives/net/http"
 	"github.com/berachain/beacon-kit/primitives/net/jwt"
 )
 
@@ -193,8 +193,8 @@ func (s *EngineClient) verifyChainIDAndConnection(
 	// After the initial dial, check to make sure the chain ID is correct.
 	chainID, err = s.Client.ChainID(ctx)
 	if err != nil {
-		if errors.Is(err, http.ErrUnauthorized) {
-			// We always log this error as it is a critical error.
+		var httpErr *ethclientrpc.HTTPStatusError
+		if errors.As(err, &httpErr) && httpErr != nil && httpErr.StatusCode == nethttp.StatusUnauthorized {
 			s.logger.Error(UnauthenticatedConnectionErrorStr)
 		}
 		return err
--- a/execution/client/errors.go
+++ b/execution/client/errors.go
@@ -23,6 +23,7 @@ package client
 import (
 	engineerrors "github.com/berachain/beacon-kit/engine-primitives/errors"
 	"github.com/berachain/beacon-kit/errors"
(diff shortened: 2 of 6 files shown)


execution/client/errors.go
```diff
@@ -62,14 +72,25 @@ func (s *EngineClient) handleRPCError(
...
 	ok := errors.As(err, &e)
 	if !ok || e == nil {
+		// Not a JSON-RPC error, classify by transport-level signal.
+		var httpErr *rpc.HTTPStatusError
+		if errors.As(err, &httpErr) && httpErr != nil {
+			switch {
+			case httpErr.StatusCode == nethttp.StatusUnauthorized:
+				// Preserve the typed sentinel so callers (notably the EL startup ChainID check at
+				// execution/client/client.go) can log JWT-specific remediation guidance.
+				return errors.Join(http.ErrUnauthorized, err)
```
calbera: why do we still keep our custom `Unauthorized` error? Since you removed the check for this custom error [here](https://github.com/berachain/beacon-kit/pull/3109/changes#diff-59382f0151b180226ff395e7ce8b0d8ebb79daf1e3415f3bed001f1719337f9dR197)? I think its okay to remove `ErrUnauthorized` from the codebase entirely as it adds unnecessary duplication of an already existing HTTP paradigm

execution/client/client.go
```diff
@@ -193,8 +193,8 @@ func (s *EngineClient) verifyChainIDAndConnection(
 	// After the initial dial, check to make sure the chain ID is correct.
 	chainID, err = s.Client.ChainID(ctx)
 	if err != nil {
-		if errors.Is(err, http.ErrUnauthorized) {
-			// We always log this error as it is a critical error.
+		var httpErr *ethclientrpc.HTTPStatusError
+		if errors.As(err, &httpErr) && httpErr != nil && httpErr.StatusCode == nethttp.StatusUnauthorized {
```
calbera, line 197: confused by `errors.As(err, &httpErr)` since `&httpErr` is a double pointer and `httpErr` is nil. Does go magically resolve it?

Also we don't log anything specific with the error log below, which is what it seems you intended to have?

> // Preserve the typed sentinel so callers (notably the EL startup ChainID check at
   // execution/client/client.go) can log JWT-specific remediation guidance.

fridrik01: there was a nilaway error on checking for nil httpErr so I added the check. Having a double pointer and checking if its nil is fine in go. For the error log below `UnauthenticatedConnectionErrorStr` is just teh generic error string `could not verify execution chain ID ...`

execution/client/errors.go
fridrik01: yes thanks correct!

bar-bera (approved):
looks good

calbera (approved):

PR #3113, opened 2026-05-26
perf(beacondb): add validator set cache to  avoid full scan in GetValidators

fridrik01:
I noticed `devnet-core` blocktimes degraded from 2s to ~14s after running for ~6 days. CPU profiling (`go tool pprof`) showed almost 50% of CPU was spent in `GetValidators()`. This method  iterates all validators from PebbleDB on every call, and is called multiple times per block.

 This PR makes three fixes:
 - `GetTotalValidators()` now reads the sequence counter instead of loading all validators just to count them. This is called every block from `processWithdrawals()`.
- `GetValidators()` results are cached in `KVStore`, invalidated on writes (`AddValidator`, `UpdateValidatorAtIndex`) and on `WithContext()` (since epoch validator updates commit to disk but don't update the parent's cache). Each block pays the DB cost at most once instead of 5-10 times.
- `GetValidators()` now also properly propagates `iter.Close()` errors via named return values.

If further optimization is needed, the cache could be made to persist across blocks by propagating the copy's updated cache back to the parent singleton after block commit (reducing the full Pebble scan from once-per-block to once-per-epoch or every 32 blocks).

### DIFF
--- a/storage/beacondb/registry.go
+++ b/storage/beacondb/registry.go
@@ -41,6 +41,7 @@ func (kv *KVStore) AddValidator(val *ctypes.Validator) error {
 		return err
 	}
 
+	kv.validatorsCache = nil
 	return kv.balances.Set(kv.ctx, idx, 0)
 }
 
@@ -49,6 +50,7 @@ func (kv *KVStore) UpdateValidatorAtIndex(
 	index math.ValidatorIndex,
 	val *ctypes.Validator,
 ) error {
+	kv.validatorsCache = nil
 	return kv.validators.Set(kv.ctx, index.Unwrap(), val)
 }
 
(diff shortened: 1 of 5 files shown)


calbera (approved):
utACK. everything here looks safe but would want to make sure a syncing node, through blocks where validators are added/updated, has no issues

storage/beacondb/registry.go
```diff
@@ -93,18 +95,18 @@ func (kv *KVStore) ValidatorByIndex(
 }
 
 // GetValidators retrieves all validators from the beacon state.
-func (kv *KVStore) GetValidators() (
-	ctypes.Validators, error,
-) {
+func (kv *KVStore) GetValidators() (_ ctypes.Validators, err error) {
+	if kv.validatorsCache != nil {
```
bar-bera, line 99: maybe consider inverting the pattern to gain 1 line code dup lol
```
if cache == nil {
   // fill
}

return kv.validatorsCache.Copy(), nil
```

storage/beacondb/registry.go
```diff
@@ -93,18 +95,18 @@ func (kv *KVStore) ValidatorByIndex(
...
+		// Return a deep copy so callers that mutate validators dont corrupt the cache.
```
bar-bera:
```suggestion
		// Return a deep copy so callers that mutate validators don't corrupt the cache.
```

bar-bera (approved):
micro nit, lgtm

storage/beacondb/registry.go
fridrik01: I actually prefer keeping as is as it makes the code less indented.

PR #3115, opened 2026-05-28
feat(preconf): add TLS support to sequencer server and validator client

fridrik01:
The preconf sequencer and validators exchange JWT auth tokens and execution payloads over plain HTTP. For a sequencer reachable over the public internet, this leaks credentials and MEV-sensitive transaction ordering in cleartext. This PR adds opt-in TLS to the sequencer server and validator client along with the tooling to run and rotate it. 

More specifically, the main changes are:

**Server TLS** (`beacon/preconf/server.go`)
- `tls-cert-path` + `tls-key-path` switch the API to HTTPS, served via `ServeTLS` with a `GetCertificate` hook.
- The TCP listener binds synchronously in `Start()`, so a bad cert/key or an in-use port fails node startup loudly instead of being logged in a goroutine behind a dead endpoint.
- No dual-mode listener: when TLS is configured the server only speaks HTTPS.

**Client pinning** (`beacon/preconf/client.go`)
- Optional `sequencer-ca-cert-path` pins the sequencer to a single CA and rejects certificates from any other CA, guarding against CA mis-issuance and BGP-hijack attacks. When unset, it falls back to the system trust store.
- A TLS 1.2 floor is enforced on every connection.

**Config and operations**
- `Config.Validate()` rejects half-configured TLS (cert without key, or pinning without an `https://` URL).
- The node logs a startup warning when a validator points at a plaintext sequencer on a non-loopback host.
- The sequencer hot-reloads its certificate on `SIGHUP` (next to the existing whitelist reload), so rotation needs no restart and a bad cert file leaves the old one serving.
- New CLI flags and `config.toml` template entries cover all three paths.

**Devnet** (`kurtosis/`)
- The kurtosis preconf devnet now runs with TLS enabled by default (`preconf.tls: true`). At plan time it generates an internal CA and a CA-signed server cert (SAN matching the sequencer service name), serves the leaf on the sequencer, and pins the CA on the validators. This mirrors the production trust topology rather than a self-signed shortcut. TLS stays a toggle, so the devnet can still run plaintext with `tls: false`.

**Documentation** (`beacon/preconf/README.md`)
- Documents the preconf system and the full TLS workflow: cert generation for dev and production, pinning, and rotation.

**Test plan**
- Unit tests cover certificate pinning (correct CA connects, wrong CA rejected), plaintext rejection, `SIGHUP` cert reload, bind-failure handling, and config validation.
- Verified end to end on the kurtosis preconf devnet (via `make start-devnet-preconf`). Confirmed validators fetch payloads from the sequencer over HTTPS with CA pinning and no certificate verification failures.


beacon/preconf/config.go
```diff
@@ -94,3 +111,24 @@ func (c *Config) IsSequencer() bool {
...
+// Validate checks the config for structural consistency.
+func (c *Config) Validate() error {
+	if (c.TLSCertPath != "") != (c.TLSKeyPath != "") {
+		return errors.New("tls-cert-path and tls-key-path must both be set or both be empty")
+	}
+	if c.SequencerCACertPath != "" {
+		if c.SequencerURL == "" {
+			return errors.New("sequencer-ca-cert-path requires sequencer-url to be set")
+		}
+		if !strings.HasPrefix(c.SequencerURL, "https://") {
```
calbera: nit: can use url.Parse instead of strings.HasPrefix as is done [here](https://github.com/berachain/beacon-kit/pull/3115/changes#diff-776e8acda3ff0a7ab894618db04ed959d066ee168ca1aa52fdae47ab3aa7e74fR81)

node-core/components/preconf_client.go
```diff
@@ -59,6 +63,32 @@ func ProvidePreconfClient(in PreconfClientInput) (*preconf.Client, error) {
 		return nil, errors.Wrapf(err, "failed to load sequencer JWT from: %s", cfg.SequencerJWTPath)
 	}
 
+	if err = cfg.Validate(); err != nil {
```
calbera: nit: could call validate on line 55 where the checks on config are done first

beacon/preconf/README.md
```diff
@@ -0,0 +1,238 @@
...
+
+# 2. Create the server key + CSR
+openssl req -newkey ec -pkeyopt ec_paramgen_curve:P-256 \
+  -keyout server-key.pem -out server.csr -nodes \
+  -subj "/CN=sequencer" \
+  -addext "subjectAltName=DNS:sequencer.example.com,IP:1.2.3.4"
+
+# 3. Sign the server cert with the CA
+openssl x509 -req -in server.csr -CA ca-cert.pem -CAkey ca-key.pem \
+  -CAcreateserial -out server-cert.pem -days 365 -copy_extensions copyall
```
calbera: `copyall` blindly copies CSR extensions. since this guide is for prod it would be safer to use `-extfile`

calbera (commented):
generally lgtm, left some nits + will review again.

Some potential followups from claude:

- **No mTLS / client certificates.** The client is authenticated at the app layer via JWT; TLS only authenticates the server. For this threat model that's adequate; mTLS would be defense-in-depth, a fine follow-up.
- **No rate-limiting / the JWT check is O(n) HMAC over all validator secrets before the whitelist/proposer checks.** Pre-existing, but this PR's whole premise is internet exposure, which makes an unauthenticated-request amplification slightly more relevant. Follow-up, not a blocker.

fridrik01:
> generally lgtm, left some nits + will review again.
> 
> Some potential followups from claude:
> 
> * **No mTLS / client certificates.** The client is authenticated at the app layer via JWT; TLS only authenticates the server. For this threat model that's adequate; mTLS would be defense-in-depth, a fine follow-up.
> * **No rate-limiting / the JWT check is O(n) HMAC over all validator secrets before the whitelist/proposer checks.** Pre-existing, but this PR's whole premise is internet exposure, which makes an unauthenticated-request amplification slightly more relevant. Follow-up, not a blocker.

mTLS is not required imo, but the others are part of https://berachain.clickup.com/t/9014124274/86b9exmww and https://berachain.clickup.com/t/9014124274/86b9exmz9

bar-bera (approved):
looks good

calbera (approved):

PR #3117, opened 2026-05-28
fix(test): fix rpcErrorProxy race causing flaky ConnectionDrop recovery test

fridrik01:
This PR fixes a flaky `TestFinalizeBlock_ConnectionDrop_Recovery` test by snapshotting `active`/`dropConn` once in `intercept()` instead of reading them separately.

`deactivate()` clears the two flags in separate stores, so a request could race in and serve a spurious empty JSON-RPC error instead of dropping the connection. The CL treats that error as non-retryable, so `FinalizeBlock` fails instead of recovering.


calbera (approved):
utack

PR #3118, opened 2026-05-29
core(cometbft): bump version to include fixes

bar-bera:
Includes cometbft fixes:
- signed header validation in light client attack evidence to avoid nil deref
- blocksync response validation to avoid memory alloc.


fridrik01:
@bar-bera reminder to put me/cal as reviewers on PR when its ready to be reviewed so we notice it quicker

fridrik01 (approved):

PR #3119, opened 2026-05-29
feat(preconf): harden HTTP server limits

fridrik01:
This PR hardens the preconf HTTP server and validator client against DoS and OOM, The sequencer is internet-facing, so it previously accepted unbounded request bodies, had no read/write/idle timeouts, and no cap on concurrent connections, leaving it open to OOM, Slowloris, and connection-exhaustion attacks.

More specifically, the main changes are:

- Request body cap (server): `MaxBytesReader` (2 KB) -> `413` on oversized bodies.
- Read/Write/Idle timeouts (server): 15s / 30s / 60s, closes Slowloris and idle keep-alive vectors.
- Max concurrent connections (server): `netutil.LimitListener(256)`.
- Response body cap (client): default 16 MiB, configurable via `max-response-size`, to prevent a malicious sequencer OOMing a validator. Picked 16 MiB which should cover the worst-case valid block, full hex-encoded transactions (~6 MiB) + 6 blobs (~1.5 MiB) with ~2× headroom.

### DIFF
--- a/beacon/preconf/client.go
+++ b/beacon/preconf/client.go
@@ -55,11 +55,12 @@ var (
 
 // Client is an HTTP client for fetching payloads from the sequencer.
 type Client struct {
-	logger       log.Logger
-	httpClient   *http.Client
-	sequencerURL string
-	jwtSecret    *jwt.Secret
-	timeout      time.Duration
+	logger          log.Logger
+	httpClient      *http.Client
+	sequencerURL    string
+	jwtSecret       *jwt.Secret
+	timeout         time.Duration
+	maxResponseSize int64
 
 	// mu protects the JWT token cache
 	mu          sync.RWMutex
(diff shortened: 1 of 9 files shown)


calbera (commented):
lgtm but can we add e2e tests for the scenarios you mentioned were previously possible (but shouldn't be anymore)?

> previously accepted unbounded request bodies, had no read/write/idle timeouts, and no cap on concurrent connections

bar-bera (commented):
agree on testing. Also `MaxHeaderBytes` (`DefaultMaxHeaderBytes = 1 << 20 // 1 MB`) and `MaxResponseHeaderBytes` (default at 10 MB) are not set so the  the effective per-request memory may be underestimated, not game changing since connections are already limited but maybe you want to consider setting them.

fridrik01:
> lgtm but can we add e2e tests for the scenarios you mentioned were previously possible (but shouldn't be anymore)?
> 
> > previously accepted unbounded request bodies, had no read/write/idle timeouts, and no cap on concurrent connections

Both size caps are covered by unit tests in this PR (oversized request body on the server, oversized response on the client). Want to avoid e2e tests for the other scenarios as its basically just socket-timing tests that are just testing go stdlib code.

PR #3122, opened 2026-06-10
chore(exec-client): add check for nil payload

bar-bera:


execution/client/engine.go
```diff
@@ -148,6 +148,10 @@ func (s *EngineClient) GetPayload(
 		// Engine API returns the Unknown Payload (-38001) error if a nil result is returned.
 		return result, engineerrors.ErrUnknownPayload
 	}
+	if result.GetExecutionPayload() == nil {
+		// Reject JSON "null" payload with a clean error.
+		return result, engineerrors.ErrNilExecutionPayload
+	}
```
fridrik01, line 154: This looks good, we are now checking `GetExecutionPayload()` and `GetBlobsBundle()` for `nil`, but in the [BuiltExecutionPayloadEnv](https://github.com/berachain/beacon-kit/blob/main/consensus-types/types/payload_env.go#L32) interface is a third method `GetBlockValue()` which can return nil, should we not add that also just for consistency? I don't think we ever call GetBlockValue() right now but that may change with preconfs so adding it would catch potential panics later

So adding this here as well: 

```
if result.GetBlockValue() == nil {
    return result, engineerrors.ErrNilBlockValue
}
```

fridrik01 (approved):
LGTM, just one note

execution/client/engine.go
bar-bera: [ad459d7](https://github.com/berachain/beacon-kit/pull/3122/commits/ad459d7f477b9b7cac25aea81a001bc1ac744be1)

calbera (approved):

PR #3124, opened 2026-06-11
feat(deposits): enforce deposit source per fork

calbera:
Add a sanity check in processOperations that rejects blocks sourcing deposits from the wrong place for their fork. Before Fulu, deposits must come from the deposit-contract events carried on the beacon block body and EIP-6110 deposit requests must be absent; from Fulu onwards (after the single first-Fulu catchup block) deposits must come from EIP-6110 execution requests and the block body must carry none.

This mirrors the execution layer, which only emits EIP-6110 deposit requests from the Osaka fork onwards, and prevents deposits from being silently double counted or dropped if the execution layer gates EIP-6110 incorrectly across the fork boundary.

Covered by unit tests for both rejection directions in state-transition/core and a simulated test exercising the post-Fulu body-deposit rejection through ProcessProposal.


fridrik01 (approved):
LGTM, 

One thing to confirm before merge. This adds a retroactive validity rule, so replay breaks if any finalized mainnet or bepolia block carries pre-Fulu deposit requests. Given the prague gate issue I would want to test this on mainnet sync nodes, or alternatively skip the pre-fulu error condition.

bar-bera (approved):
> LGTM,
> 
> One thing to confirm before merge. This adds a retroactive validity rule, so replay breaks if any finalized mainnet or bepolia block carries pre-Fulu deposit requests. Given the prague gate issue I would want to test this on mainnet sync nodes, or alternatively skip the pre-fulu error condition.

Confirmed by checking in a block range (from `21796167` 4 Jun 2026 21:04:15 +0, to `22091805` 11 Jun 2026 17:18:46 +0) that all blocks with deposits have empty `.result.requestsHash` from `eth_getBlockByNumber`.

PR #3129, opened 2026-06-19
fix(node-api): enforce DA period check in GetBlobSidecars

### LINKED ISSUE #3127: GetBlobSidecars DA period check compares slot against itself
## Problem

`GetBlobSidecars` is supposed to reject requests for slots outside `MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS`. It calls `WithinDAPeriod` with the requested slot as **both** arguments:

https://github.com/berachain/beacon-kit/blob/main/node-api/handlers/beacon/blobs.go#L73-L79

    if !h.cs.WithinDAPeriod(slot, slot) {

`WithinDAPeriod(block, current)` is defined as:

https://github.com/berachain/beacon-kit/blob/main/chain/helpers.go#L57-L61

    func (s spec) WithinDAPeriod(block, current math.Slot) bool {
    	return s.SlotToEpoch(block)+s.MinEpochsForBlobsSidecarsRequest() >= s.SlotToEpoch(current)
    }

When `block == current`, that's just `epoch(slot) + N >= epoch(slot)`, which is always true. The check can never fail.

We already use this correctly on the consensus side — `FinalizeSidecars` passes the block slot and chain head:

https://github.com/berachain/beacon-kit/blob/main/beacon/blockchain/finalize_block.go#L97

    s.chainSpec.WithinDAPeriod(blk.GetSlot(), math.Slot(syncingToHeight))

Looks like the API handler just wired up the wrong second arg. `GetSyncData()` is already available in the same function (used a few lines above for `head` block id).

## Expected

If someone asks for blob sidecars at a slot that's fallen outside the DA window, we should fail the request up front — before we touch the availability store.

The logic for this already exists in `WithinDAPeriod`. There's a test for it in [chain/helpers_test.go](https://github.com/berachain/beacon-kit/blob/main/chain/helpers_test.go#L102-L135) with `MinEpochsForBlobsSidecarsRequest = 5`:

- slot `0` vs head at `160` → still in window
- slot `0` vs head at `192` → outside window, should reject

So for a concrete example: chain head at slot 192, client requests sidecars for slot 0 → API should return an error. Right now it doesn't.

## Actual

The guard at line 74 never trips because both args are the same slot. Every request gets through to `GetBlobSidecarsAtSlot`.

What the caller gets back is basically luck — if pruning hasn't cleaned up old blobs yet, they'll get data for a slot we shouldn't be serving. If pruning already ran, they get empty. Either way the API isn't enforcing the retention policy itself.

## Suggested fix

Use chain head as the second arg. `GetSyncData()` is already called in this handler for the `head` block id case, so it's just wiring:

    headHeight, _ := h.backend.GetSyncData()
    if headHeight < 0 {
        return nil, errors.New("invalid negative block height")
    }
    if !h.cs.WithinDAPeriod(slot, math.Slot(headHeight)) {
        return nil, fmt.Errorf(
            "requested slot (%d) is not within Data Availability Period (previous %d epochs)",
            slotID, h.cs.MinEpochsForBlobsSidecarsRequest(),
        )
    }

Would also add a test in [blob_test.go](https://github.com/berachain/beacon-kit/blob/main/node-api/handlers/beacon/blob_test.go) — mock head at something like slot 192, request slot 0, assert we get an error. Currently only the happy path with `head` is covered.

NikhilSharmaWe:
Fixes #3127 

## Summary

`GetBlobSidecars` was calling `WithinDAPeriod(slot, slot)`, which always evaluates to true, so the DA retention guard never ran.

This change compares the requested slot against **local head** from `GetSyncData()`, using the same retention window as availability pruning. The error message now prints the resolved slot instead of `slotID` (which is `-1` for `head`).

**Before:** `WithinDAPeriod(slot, slot)` → always passes  
**After:** `WithinDAPeriod(slot, headHeight)` → rejects slots outside `MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS`

Local head is used (not sync target) so `/blob_sidecars/head` still works while the node is syncing.

## Test plan

- [x] `go test -tags=test ./node-api/handlers/beacon/... -run TestGetBlobSidecars`
- [x] `go test -tags=test ./node-api/handlers/beacon/...`
- [x] Head request succeeds (`head` block id)
- [x] Head request succeeds while syncing (`headHeight=1000`, `syncToHeight=800_000`)
- [x] Explicit in-window slot accepted (999000 with head at 1_000_000)
- [x] Out-of-window slot rejected (212568 with head at 1_000_000)
- [x] Store not queried when DA check fails (`AssertNotCalled` on `GetBlobSidecarsAtSlot`)

### DIFF
--- a/node-api/handlers/beacon/blob_test.go
+++ b/node-api/handlers/beacon/blob_test.go
@@ -168,7 +168,7 @@ func TestGetBlobSidecars(t *testing.T) {
 				require.Nil(t, res)
 				require.ErrorContains(t, err, "not within Data Availability Period")
 				require.ErrorContains(t, err, "212568")
-				backend.AssertNotCalled(t, "GetBlobSidecarsAtSlot")
+				backend.AssertNumberOfCalls(t, "GetBlobSidecarsAtSlot", 0)
 			},
 		},
 	}
--- a/node-api/handlers/beacon/blob_test.go
+++ b/node-api/handlers/beacon/blob_test.go
@@ -70,10 +70,10 @@ func TestGetBlobSidecars(t *testing.T) {
 		name                string
 		inputs              func() beacontypes.GetBlobSidecarsRequest
 		setMockExpectations func(*testing.T, *mocks.Backend)
-		check               func(t *testing.T, res any, err error)
+		check               func(t *testing.T, backend *mocks.Backend, res any, err error)
 	}{
 		{
-			name: "success",
+			name: "success - head",
 			inputs: func() beacontypes.GetBlobSidecarsRequest {
 				return beacontypes.GetBlobSidecarsRequest{
 					BlockIDRequest: handlertypes.BlockIDRequest{
@@ -88,7 +88,7 @@ func TestGetBlobSidecars(t *testing.T) {
 				b.EXPECT().GetSyncData().Return(int64(1234), int64(1234))
 				b.EXPECT().GetBlobSidecarsAtSlot(mock.Anything).Return(testSidecars, nil)
 			},
-			check: func(t *testing.T, res any, err error) {
+			check: func(t *testing.T, _ *mocks.Backend, res any, err error) {
 				t.Helper()
 
 				require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


NikhilSharmaWe:
@fridrik01 PTAL, thanks

PR #3136, opened 2026-06-24
fix(vulncheck): build govulncheck against patched x/tools (golang/go#80055)

fridrik01:
Currently, running `make vulncheck` crashes due to a known bug in the upstream tooling that's already fixed but not yet released. In the meantime this results in errors on our CI pipeline and  our daily security scan.

This PR addresses this by running `govulncheck` with the patched upstream fix until it lands in a future release.

Once a `govulncheck` release includes the fix, delete `scripts/build/govulncheck.sh` and restore the original one-line command.


bar-bera (approved):
looks good

PR #3140, opened 2026-06-25
chore(deps): bump go-ethereum to v1.17.4

fridrik01:
The nightly `test-latest-deps` job fails (see https://github.com/berachain/beacon-kit/actions/runs/28165028424/job/83414743879) because the newest go-ethereum (v1.17.4) renamed its Verkle genesis APIs to UBT. That broke the `gethlib` build and everything importing it.

This PR addresses this:
- Bump go-ethereum to v1.17.4
- Migrates to the new API 

Both empty hashes are the zero value, so behavior is unchanged.

After merging this PR our  nightly `test-latest-deps` job should be green as this was AFAICT the only error generated after bumping all dependencies to newest minor/patch version


bar-bera (approved):
looks good. should we rename all of this stuff too? `EnableVerkleAtGenesis`, `IsVerkleGenesis()` `VerkleTime`. or do you prefer to not touch them?

calbera (approved):

bar-bera (approved):

PR #3143, opened 2026-07-04
fix(ci): pre-build the Go coverage tool to fix flaky coverage test failures

fridrik01:
Since we now have a nightly CI pipeline that reports error on slack I noticed a flaky failure this morning on CI when running `make test-unit-cover`, see https://github.com/berachain/beacon-kit/actions/runs/28702600175/job/85122973342

The issue seems to be that when collecting coverage, `go test` builds the Go coverage tool (`covdata`) into the build cache on demand, then runs it. In our parallel coverage runs it can get executed while still being written, failing with `text file busy` as we see in that run above.

This PR addresses this by building covdata once before the tests start, so the parallel runs reuse the cached binary instead of racing to build-and-run it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ ROOT_DIR := $(shell pwd)
 .PHONY: clean format lint \
 	buf-install proto-clean \
 	test-unit test-unit-cover test-simulated test-forge-cover test-forge-fuzz \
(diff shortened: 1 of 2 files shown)


bar-bera (approved):
lgtm

PR #3145, opened 2026-07-10
chore(test): fix flaky run on proofs tests

bar-bera:
Flaky run: https://github.com/berachain/beacon-kit/actions/runs/29086626434/job/86341770240

Simply refactors the requests into a single function, adding check on http response status (which was not checked).
Catches the http error and retries request.
Fix extended to all tests even though just `TestValidatorPubkeyProof` was recorded to fail.

### DIFF
--- a/testing/e2e/standard/proofs_test.go
+++ b/testing/e2e/standard/proofs_test.go
@@ -25,15 +25,31 @@ package standard_test
 import (
 	"math/big"
 	"strconv"
+	"strings"
+	"time"
 
 	"github.com/berachain/beacon-kit/config/spec"
 	"github.com/berachain/beacon-kit/gethlib/ssztest"
 	"github.com/berachain/beacon-kit/node-api/handlers/proof/merkle"
+	ptypes "github.com/berachain/beacon-kit/node-api/handlers/proof/types"
 	"github.com/berachain/beacon-kit/primitives/common"
 	"github.com/berachain/beacon-kit/primitives/math"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	coretypes "github.com/ethereum/go-ethereum/core/types"
 )
+
+// errSlotNotIndexed is the transient server error returned while a queried node is still indexing the head.
+const errSlotNotIndexed = "slot not found at timestamp"
+
+// retryTimestampProof re-runs a t<timestamp> proof fetch while the queried node returns the errSlotNotIndexed.
+func retryTimestampProof[T any](fetch func() (*T, error)) (*T, error) {
+	resp, err := fetch()
+	for i := 0; i < 15 && err != nil && strings.Contains(err.Error(), errSlotNotIndexed); i++ {
+		time.Sleep(200 * time.Millisecond)
+		resp, err = fetch()
+	}
+	return resp, err
+}
 
 // TestBlockProposerProof tests the block proposer proof endpoint by fetching and verifying
 // the block proposer proofs against the SSZTest contract. Refer to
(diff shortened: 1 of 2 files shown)


testing/e2e/standard/proofs_test.go
```diff
@@ -25,16 +25,32 @@ package standard_test
...
 	"github.com/berachain/beacon-kit/primitives/math"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	coretypes "github.com/ethereum/go-ethereum/core/types"
 )
 
+// errSlotNotIndexed is the transient server error returned while a queried node is still indexing the head.
+const errSlotNotIndexed = "slot not found at timestamp"
+
+// retryTimestampProof re-runs a t<timestamp> proof fetch while the queried node returns the errSlotNotIndexed.
+func retryTimestampProof[T any](fetch func() (*T, error)) (*T, error) {
```
fridrik01: Suggest better name like

```suggestion
func retryUntilIndexed[T any](fetch func() (*T, error)) (*T, error) {
```

testing/e2e/standard/proofs_test.go
```diff
@@ -25,16 +25,32 @@ package standard_test
...
 	"github.com/berachain/beacon-kit/primitives/common"
 	"github.com/berachain/beacon-kit/primitives/math"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	coretypes "github.com/ethereum/go-ethereum/core/types"
 )
 
+// errSlotNotIndexed is the transient server error returned while a queried node is still indexing the head.
+const errSlotNotIndexed = "slot not found at timestamp"
+
+// retryTimestampProof re-runs a t<timestamp> proof fetch while the queried node returns the errSlotNotIndexed.
```
fridrik01:
```suggestion
// retryUntilIndexed re-runs fetch while the node reports it has not yet indexed the queried timestamp 
```

testing/e2e/standard/proofs_test.go
```diff
@@ -25,16 +25,32 @@ package standard_test
...
 	"github.com/berachain/beacon-kit/node-api/handlers/proof/merkle"
+	ptypes "github.com/berachain/beacon-kit/node-api/handlers/proof/types"
 	"github.com/berachain/beacon-kit/primitives/common"
 	"github.com/berachain/beacon-kit/primitives/math"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	coretypes "github.com/ethereum/go-ethereum/core/types"
 )
 
+// errSlotNotIndexed is the transient server error returned while a queried node is still indexing the head.
+const errSlotNotIndexed = "slot not found at timestamp"
```
fridrik01: Should comment that this relies on error string in `storage/block/store.go:128` not changing

testing/e2e/suite/types/beacon_client.go
```diff
@@ -233,13 +234,8 @@ func (c *customBeaconClient) Validators(
...
 
-// BlockProposerProof calls `bkit/v1/proof/block_proposer/:timestamp_id` endpoint to get
-// the merkle proofs that can be used to verify the block proposer for a given timestamp id.
-func (c *customBeaconClient) BlockProposerProof(
-	ctx context.Context, timestampID string,
-) (*ptypes.BlockProposerResponse, error) {
-	url := fmt.Sprintf("%s/bkit/v1/proof/block_proposer/%s", c.address, timestampID)
-
+// doProofRequest issues a GET to a proof endpoint and decodes the JSON body into *T.
+func doProofRequest[T any](ctx context.Context, c *customBeaconClient, url string) (*T, error) {
```
fridrik01, line 260: question: could we move the retry into doProofRequest instead? Then the tests don't change at all and we avoid passing function variables

fridrik01 (commented):
quick pass

testing/e2e/suite/types/beacon_client.go
bar-bera: [4e4dfd4](https://github.com/berachain/beacon-kit/pull/3145/commits/4e4dfd4580d37f78db4bc5c032cca531d2d85dac) - wrapped the doProofRequest within the retry loop. also safe for calls that do not need the loop (blockNumber-1)

fridrik01 (approved):
LGTM

PR #3147, opened 2026-07-16
feat(consensus): Implement and wire halt-height and halt-time

fridrik01:
This PR implements and wires the  `--halt-height` and `--halt-time` flags so a network can perform a coordinated halt for binary-swap upgrades. I noticed these were missing while working on the cometbft re-fork (berachain/cometbft#51).

This is the upgrade path for releases where old and new binaries cannot safely coexist in one validator set.

With every validator configured to the same halt point, the network commits the halt block and shuts itself down. Operators then swap the beacond binary and restart without the halt flags, and the chain resumes at the next height from the same data directories, with no replay or rollback.

The implementation makes this safe in three ways:
- The halt runs only after the block is fully committed, and waits a 5s grace period before exiting so slower peers can finish committing the halt block before gossip goes quiet.
- A node can never commit a block past the halt point. A node restarted with the halt flags still set refuses to start, and the ABCI handlers neither propose nor finalize past the halt point, so the shutdown stays graceful instead of escalating into a CometBFT consensus-failure panic.
- `Commit` now records the real block height and time in the commit header, which also fixes `min-retain-blocks` pruning, silently a no-op until now (see notes below).

### Testing

- Unit tests for the halt predicate and restart refusal (`commit_internal_test.go`).
- New e2e harness (`testing/upgrade/halt-swap-resume-test.sh`): a local multi-validator devnet halts on the old binary, one node must refuse to restart with the flag set, then all nodes resume from the same data dirs on the new binary with tx load asserted on both sides of the swap. Runs in two flavors:
  - `make test-halt-swap-resume` (halt at a fixed height, --halt-height)
  - `make test-halt-swap-resume-time` (halt at a wall-clock time, --halt-time)
- This is a bash harness rather than a kurtosis e2e test because the subject under test is process lifecycle, halting nodes, swapping the binary underneath them, and restarting from the same data directories, which containers built from a fixed image make awkward. Bare processes also keep the run Docker-free and fast enough for nightly CI.
- Both run in nightly CI, with logs uploaded on failure.

### Notes for release

- The `min-retain-blocks` setting now takes effect. It has been silently ignored until now, so operators who have it configured will see CometBFT begin pruning old blocks after upgrading. Anyone relying on it having no effect should unset it before upgrading.
- Forgotten `halt-height` or `halt-time` values in existing configs become live. A non-zero value the old binary ignored will now halt the node, so configs are worth checking before rolling this out.
- The e2e test uncovered an existing cometbft bug that this PR does not fix, though I tried to mitigate it:
  1. When validators exit right after committing the halt block, a slower validator can be left still waiting for the final precommits it needs to commit that block itself, stuck one block short with no peers left to send them. 
  2. After the swap it stays stuck as long as it lags by exactly one block, because restarted peers only hold the aggregated commit and vote gossip needs the individual signatures that aggregation discards. 
  3. If the stuck validators hold less than 1/3 of the power the rest of the network resumes without them and they recover as soon as the next block is committed, but past 1/3 the network cannot produce that block and the restart deadlocks. 
  4. The 5s grace period exists to prevent the stranding in the first place. 
  5. The correct solution to this is fixing this in our cometbft forl (tracked in https://berachain.clickup.com/t/9014124274/86bayuecy)


consensus/cometbft/service/commit.go
```diff
@@ -23,43 +23,123 @@ package cometbft
...
+	s.logger.Info("halting node per configuration",
+		"halt_height", s.haltHeight, "halt_time", s.haltTime, "committed_height", s.finalizedHeight, "grace_period", haltGracePeriod)
+
+	// Sleeping here blocks the consensus state machine inside Commit, so no halting node can advance to the
+	// next height while its peer gossip routines keep serving the halt-block precommits from the live vote set.
+	time.Sleep(haltGracePeriod)
+
+	// Signal our own process so the node's regular shutdown path runs, the same mechanism a cosmos-sdk baseapp
+	// uses for halt-height.
+	p, err := os.FindProcess(os.Getpid())
```
bar-bera: From docs:

On Unix systems, FindProcess always succeeds and returns a Process for the given pid, regardless of whether the process exists. To test whether the process actually exists, see whether p.Signal(syscall.Signal(0)) reports an error.

So could nit out the err check.

testing/upgrade/halt-swap-resume-test.sh
```diff
@@ -0,0 +1,522 @@
...
+        --beacon-kit.logger.log-level info "$@" >"$LOG_DIR/$logfile" 2>&1 &
+    echo $!
+}
+
+# Validators that exited after logging the halt line, respectively ones still running.
+halted_count() {
+    local n=0 k
+    for ((k = 0; k < NUM_VALS; k++)); do
+        if ! kill -0 "${CL_PIDS[k]}" 2>/dev/null &&
+            grep -q "halting node per configuration" "$LOG_DIR/cl$k.old.log"; then
```
bar-bera, line 141: strictly coupled to log strings, but ok i don't see a way to do it better without worsening the whole structure

bar-bera (approved):
super good

PR #3150, opened 2026-07-23
fix(log): cap oversized log lines with a truncation marker

fridrik01:
This PR caps any single log line at `64KB` in the `phuslu` logger and replaces the dropped bytes with a `...[truncated, total size <N> bytes]` marker so one oversized value can't expand an entry into megabytes. It applies to both JSON and console styles and leaves normal logs untouched.

Note: truncation a JSON line will not be well-formed JSON. To address that properly we need to rely on go generics and bind each field value before serialization, way more complex but doable. Just not sure we need it as consumers of json would be able to handle a single ill-formed json line.

### DIFF
--- a/log/phuslu/bound.go
+++ b/log/phuslu/bound.go
@@ -1,0 +1,61 @@
+// SPDX-License-Identifier: BUSL-1.1
+//
+// Copyright (C) 2025, Berachain Foundation. All rights reserved.
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file of this repository and at www.mariadb.com/bsl11.
+//
+// ANY USE OF THE LICENSED WORK IN VIOLATION OF THIS LICENSE WILL AUTOMATICALLY
+// TERMINATE YOUR RIGHTS UNDER THIS LICENSE FOR THE CURRENT AND ALL OTHER
+// VERSIONS OF THE LICENSED WORK.
+//
+// THIS LICENSE DOES NOT GRANT YOU ANY RIGHT IN ANY TRADEMARK OR LOGO OF
+// LICENSOR OR ITS AFFILIATES (PROVIDED THAT YOU MAY USE A TRADEMARK OR LOGO OF
+// LICENSOR AS EXPRESSLY REQUIRED BY THIS LICENSE).
+//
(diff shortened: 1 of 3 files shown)


log/phuslu/bound.go
```diff
@@ -0,0 +1,61 @@
...
+		keep = 0
+	}
+	line := make([]byte, 0, keep+len(marker))
+	line = append(line, p[:keep]...)
+	line = append(line, marker...)
+	n, err := w.out.Write(line)
+	if err != nil {
+		return 0, err
+	}
+	if n < len(line) {
```
calbera, line 55: question: why do we need to this explicit error handling? this error can be handled up by caller no? as it would be if normally calling `.Write()` on an arbitrary `writer`

log/phuslu/logger.go
```diff
@@ -172,14 +172,14 @@ func (l *Logger) withLogLevel(level string) {
...
-		Writer:    l.out,
+		Writer:    truncatingWriter{out: l.out},
 		Formatter: l.formatter.Format,
 	})
 }
 
 // useJSONWriter sets the logger to use a IOWriter wrapper.
 func (l *Logger) useJSONWriter() {
-	l.setWriter(log.IOWriter{Writer: l.out})
+	l.setWriter(log.IOWriter{Writer: truncatingWriter{out: l.out}})
```
calbera, line 182: this won't be pretty for JSON as you mentioned, but this case is so rare (and probably a form of DOS) that it's probably fine to be defensive here

calbera (approved):
LGTM

log/phuslu/bound.go
fridrik01: Its to retain the callers ability to see a short write, which it otherwise cant here since we truncated the line and return `len(p)` instead of the real `n`. Basically preventive code in case the underlying `io.Writer` ever returns `nil` error when it returns `n < len(line)`

log/phuslu/logger.go
fridrik01: I really thought about implementing the generics part to keep the json well formed, but its quite a lot of code and complex, and as you say if it ever happens its most likely DoS / amplification attack, etc so aimed for simplicity. However, we can definetly implement it later if needed.

PR #3152, opened 2026-07-29
chore(forks): set osaka1 mainnet activation time

fridrik01:
This PR wires the osaka1 timestamp into the mainnet genesis file

Unix Timestamp: `1785945600`
GMT: `Wednesday, August 5, 2026 at 16:00:00 UTC`

Osaka1 is EL-only, so no beacon-kit behavior changes. This matters for operators who start `bera-reth` with an explicit genesis file rather than the built-in `--chain mainnet` preset (or defaults), since a stale file leaves the fork unscheduled with no error and the node halts at activation.


chuck-bear (approved):
LGTM

PR #3155, opened 2026-08-14
chore(deps): move onto cometbft bera-v0.39.x re-fork

fridrik01:
This PR updates beacon-kit to move off the `bera-v1.x` cometbft line onto the `bera-v0.39.x` re-fork (berachain/cometbft#51). The v0.39 line drops the v1 `api` module and ABCI surface, so cosmos-sdk moves with it to v0.54 and `store/v2` (the old `cosmossdk.io/store` relocated upstream, same IAVL disk format). 

Most of the diff is mechanical renames and import moves. The substantive bits are stubs for the new mempool ABCI methods and a local replacement for the sdk validator-file helpers, which lost BLS key-type selection.

One behavior change, v0.39 has no `SyncingToHeight`, so the DA-window check and node-api sync status use the finalized height instead. We may want to restore that field in the fork.


fridrik01:
Replaced with https://github.com/berachain/beacon-kit/pull/3158 swhich targets the 0.40 dev line

PR #3164, opened 2026-09-01
chore(testing/bepolia): update el-peers and el-bootnodes

chuck-bear:
## Summary

Add operational seed nodes and replace unreachable bootnode IPs with publicly dialable nodes for Bepolia EL peering.


fridrik01 (approved):
LGTM

PR #3165, opened 2026-09-02
chore(testing/mainnet): refresh el-peers and el-bootnodes

chuck-bear:
## Summary

- Keep Asia (Korea/Singapore) seed nodes
- Refresh Germany and Canada seed enodes from live seed pods against their TCP LoadBalancer IPs


fridrik01 (approved):
LGTM

PR #3167, opened 2026-09-17
fix(storage): prune deposits in [start, end) instead of [start, start+end)

### LINKED ISSUE #3153: Deposit KVStore.Prune deletes `[start, start+end)` instead of `[start, end)`
### Description

`Prune` is documented as removing deposits in `[start, end)`, but the loop runs `for i := range end` and deletes `start+i`, so it removes `end` keys from `start` instead of the keys in `[start, end)`.
https://github.com/berachain/beacon-kit/blob/main/storage/deposit/v1/store.go#L147-L159
```go
// Prune removes the [start, end) deposits from the store.
for i := range end {
	if err := kv.store.Remove(ctx, start+i); err != nil {
		return errors.Wrapf(err, "failed to prune deposit %d", start+i)
	}
}
```
Example: `Prune(5, 10)` should delete `5..9`. It deletes `5..14`.
Same `[start, end)` contract is handled correctly in `RangeDB`:
https://github.com/berachain/beacon-kit/blob/main/storage/filedb/range_db.go#L123-L132
### Impact
Not hit in production today. `depositPruneRangeFn` always returns `(0, 0)` because the full deposit list is kept for consensus:
https://github.com/berachain/beacon-kit/blob/main/beacon/blockchain/pruning.go#L47-L51
Still wrong vs the API comment. If pruning is ever re-enabled with `start > 0`, this over-deletes. (`Prune(0, N)` happens to work because `start+i == i`.)
### Fix
Use `for i := start; i < end; i++` with `Remove(ctx, i)`, plus a unit test for `Prune(5, 10)`.
Happy to open a PR if this seems worth fixing.

Tyagiquamar:
Closes #3153. Prune is documented as removing deposits in [start, end), but the loop ran for i in [0, end) and deleted start+i, removing end keys from start. Prune(5, 10) deleted 5..14 instead of 5..9. Fixed the loop to iterate i in [start, end), matching the RangeDB Prune contract. Added TestPruneRemovesStartEndRange: fails unpatched, passes patched. Verified in Docker golang:1.26-bookworm: gofmt clean, go vet exit 0, go build ./storage/... exit 0, go test -tags bls12381,test ./storage/deposit/... exit 0.


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

